PR #75, opened 2015-09-10
Undisposed Graphics object

stephbu:
Noticed it while I was cruising through the source, probably no biggie but it was in there anyway.
g/Graphics is a disposable object backed by native GDI objects.


JuliusSweetland:
Thanks - I'll merge this when I get a chance.

PR #90, opened 2015-09-11
Fixed crash on startup

dabbertorres:
Cause was from EnableLUA registry value not existing.

I'm not sure how common of a configuration this is, but OptiKey crashed on startup.

Turns out my machine does not have an EnableLUA registry value.

Simple solution was a try..catch for a NullReferenceException. If the value does not exist, UAC must not be enabled, so I just returned false.

Runs on my machine just fine now.

### DIFF
--- a/src/Static/DiagnosticInfo.cs
+++ b/src/Static/DiagnosticInfo.cs
@@ -228,8 +228,16 @@ namespace JuliusSweetland.OptiKey.Static
         {
             get
             {
-                var uacKey = Registry.LocalMachine.OpenSubKey(uacRegistryKey, false);
-                return uacKey.GetValue(uacRegistryValue).Equals(1);
+                try
+                {
+                    var uacKey = Registry.LocalMachine.OpenSubKey(uacRegistryKey, false);
+                    return uacKey.GetValue(uacRegistryValue).Equals(1);
+                }
+                catch(NullReferenceException)
+                {
+                    // if there is no key, then it must not be enabled
+                    return false;
+                }
             }
         }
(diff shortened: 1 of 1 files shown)


src/Static/DiagnosticInfo.cs
```diff
@@ -228,8 +228,16 @@ private static bool IsUacEnabled
         {
             get
             {
-                var uacKey = Registry.LocalMachine.OpenSubKey(uacRegistryKey, false);
-                return uacKey.GetValue(uacRegistryValue).Equals(1);
+                try
+                {
+                    var uacKey = Registry.LocalMachine.OpenSubKey(uacRegistryKey, false);
```
SamPlusPlus: Wouldn't it be better to actually check for null and not rely on a try catch?

JuliusSweetland:
Yes it would actually - I presume uacKey is potentially null. Would you make this change and submit it?

SamPlusPlus:
@JuliusSweetland I am interested in helping out with Optikey, not sure when I will have the time in the near future. This should be a quick change, I will see if I can sneak it in tonight.

JuliusSweetland:
Thanks Samuel. No rush. Otherwise I can make the change later this evening
as I have a couple of other pull requests on the localisation branch to
check and merge.

Thanks,
Julius

SamPlusPlus:
Looks like you beat me to it!

JuliusSweetland:
I've got so much going on it was easier to just do it than look out for the
pull request! Thank you for pointing it out.

PR #95, opened 2015-09-12
Fixed window movement from top or bottom docked position

suvjunmd:
The window can get offscreen after the following steps:


JuliusSweetland:
Fantastic! Thank you. Excellent video demo of the issue also - really impressed with that! Releasing this fix in v.2.0.7.

PR #97, opened 2015-09-14
Add french dictionary

Gadal:
Currently ~330k lines.  May need to check a word frequency database to pare this down to ~60k, in line with other three dictionaries' lengths.


JuliusSweetland:
Looks good. We can test to find out if it is too large. Before I merge could you please remove any one letter entries, e.g. "a". Every dictionary entry needs to be at least 2 characters long to work effectively. Thank you.

Gadal:
Removed "a", "à", and "y".

PR #100, opened 2015-09-16
Strings

Ar2rZ:
Add Strings spreadsheet


JuliusSweetland:
I have manually added the strings spreadsheet (as "Localisation_Strings.xls") to avoid the "test" commits. I have also made a couple of slight changes in the "NOTIFICATIONS" section and 1 addition in that section. From a quick look it seems to be an excellent addition - thank you.

PR #112, opened 2015-09-21
Fully localized UI and French translation

feugy:
To start with OptiKey, I wanted to perform a simple localization task.
It turns out that it was a pretty huge work :)

I moved to Resources every hard-coded string I found in Xaml files, as well as strings displayed in UI (like confirmation or enums) declared in Cs files.

Then I made a French translation, added a dictionary, and a tiny walkthrough (in markdown, I can translate it to txt if needed) for further localizations.

Even If I checked every screen (portrait and landscape) I may have missed some values (notably in Cs files). So don't hesitate to tell me so.

EDIT: I just discovered this morning the "localisa" branch... I'm sorry having donc this work in a slight different manner that you planned to.


JuliusSweetland:
@feugy wow that is excellent work. How do you think we can integrate this with the work being done by Arthur (and everyone helping to translate) in the "localisa" branch and this issue https://github.com/JuliusSweetland/OptiKey/issues/73?

I really don't want to waste either set of work. I'll add a comment to the issue to ask people's thoughts there also.

Really great work - I very much appreciate it.

JuliusSweetland:
@feugy FYI you can see the translated set of strings into French from the transifex site here; https://www.transifex.com/optikey/dashboard/all_resources/fr/?project=optikey it might be useful to compare the strings you have found.

JuliusSweetland:
Closing as we need this on the "localisa" branch. Check my latest comments on the #73 issue - we are going to keep your work, but we need to also keep the resource string keys from the Transifex site.

PR #113, opened 2015-09-25
Updated full keyboard screenshots

pmurley:
I haven't quite finished all of them, but I want to make sure this is what you are looking for with these.


pmurley:
Added more screenshots. Did not do individual keys or management console. Tried to get everything that had an old keyboard in it.

PR #117, opened 2015-09-27
Upper case french strings.

feugy:
Use uppercased string in french translation
Allow Key UI component to apply a case transformation on its Text property
Add a new settings to control case transformation on all displayed keys

### DIFF
--- a/src/App.config
+++ b/src/App.config
@@ -53,7 +53,8 @@
       <setting name="GazeTrackerUdpRegex" serializeAs="String">
         <value>^STREAM_DATA\s(?&lt;instanceTime&gt;\d+)\s(?&lt;x&gt;-?\d+(\.[0-9]+)?)\s(?&lt;y&gt;-?\d+(\.[0-9]+)?)</value>
       </setting>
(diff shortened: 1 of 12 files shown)


JuliusSweetland:
Very nice indeed. When I change the case setting in the management console what is triggering the key's Text dependency property change handler to be triggered and re-evaluate the text again? I can't see it from a quick glance.

Is there anything else you can think of that should be done for this? I think French is ready to be released.

feugy:
The key case is updated because the whole keyboard is refreshed when settings are changed.

See on [Key.cs](https://github.com/JuliusSweetland/OptiKey/blob/localisa/src/UI/Controls/Key.cs#L214) file: the TextChange is called again.

It may not be the best way to refresh the text, but it seems not too intrusive to me.

I think we're done with French, and frankly, I'm glad it's over :)

JuliusSweetland:
Understood.

You've been an absolute star. I'll release this when I get home tomorrow,
initially as a pre release and then a general release when a few people
have tried it.

PR #118, opened 2015-09-29
dictionary check added

ninehundred1:
Tried to add the requested update for the old file name. Please check

### DIFF
--- a/src/Services/DictionaryService.cs
+++ b/src/Services/DictionaryService.cs
@@ -64,7 +64,33 @@ namespace JuliusSweetland.OptiKey.Services
 
                 if (File.Exists(userDictionaryPath))
                 {
-                    LoadUserDictionaryFromFile(userDictionaryPath);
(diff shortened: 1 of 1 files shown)


JuliusSweetland:
@ninehundred1 Thank you - I've built on this and released this fix to the "localisa" branch, but I can't merge this change as it's not completely correct. The problem with git is that you can't merge in a change, make alterations, and then accept it, so the history on the project gets a bit messed up if you accept something that isn't quite finished and then correct it with another checkin. I hope you don't mind that I've done it this way.

FYI my commit is here if you would like to see it: https://github.com/JuliusSweetland/OptiKey/commit/89df864098f249aea1f730643f2b846a05579390

ninehundred1:
Thanks for being nice but what you wrote has very little left of what I
wrote, so don't worry. These are good ways to learn, and seeing how you did
that teaches me stuff.

PR #127, opened 2015-10-03
Found an untranslated string

feugy:
I'vd added a resource key for "calibration complete" message (Tobii EyeX).
Sorry for that, but my Visual Studio automatically reorder keys in the english resx file (ENGLISH_IK, ENGLISH_CA, ENGLISH_US, KEY_CASE_LABEL, TITLE_CASE, LOWER_CASE, UPPER_CASE)

Just pay attention to a proposal in FRANCE_FRENCH: I think we should provide default locale names in the locale language, to allow user detect their own langage, even if they're not confident with english.

Sorry for not having seen this missing key before. To me the 2.1.0 version is ready for french. I've accepted the french values on trensiflex because no one avec made any change since a week.

### DIFF
--- a/src/Properties/Resources.Designer.cs
+++ b/src/Properties/Resources.Designer.cs
@@ -143,15 +143,6 @@ namespace JuliusSweetland.OptiKey.Properties {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to American English.
-        /// </summary>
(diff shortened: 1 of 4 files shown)


JuliusSweetland:
Very good catch - I've updated the resource file on Transifex to add the missing string and merged your changes into master.

Issue #69, opened 2015-06-06
Next word prediction - Presage


JuliusSweetland:
After each letter when spelling a word - create word completions that are tolerant of misspellings AND whole word corrections (assuming word is complete but incorrectly spelled)

  Multi key capture

```
Netspell

Nhunspell
```

JuliusSweetland:
Could be solved as part of the potential integration of Presage for next word prediction.

JuliusSweetland:
From Arnaud (https://github.com/amartel/OptiKey)

Install and setup PRESAGE is very easy, especially for you as english is the default language database (English language is generated from single novel "The picture of Dorian Gray" by Oscar Wilde)
1) download the installer for Windows 32 bits (and not the 64bits version) at http://sourceforge.net/projects/presage/files/presage/0.9.1/win32/presage-0.9.1-32bit-setup.exe/download, run it and install all the options (wcf service is required and will start automatically at user login).
2) in Optikey project, add a reference to "C:\Program Files (x86)\presage\bin\presage.net.dll ». 
3) Sample code to integrate PRESAGE

using presage;

//context to be used to predict the next suggestion. For Optikey, this can be all the text entered in the scratchpad area, including in progress word (using private variable Text in KeyboardOutputService)
string Text;

//callback used to return the context for the predictions
private string callback_get_past_stream() {
	return Text;
}

//I don’t really understand the use of this callback function but, when looking in the presage samples, it always return an empty string ;-)
private string callback_get_future_stream() {
	return "";
}

Presage prsg = new Presage(this.callback_get_past_stream, this.callback_get_future_stream);
string[] suggestions = prsg.predict();

Note: I had crashes when using a single instance of Presage object. The workaround was to instantiate a new Presage object each time I wanted suggestions.

4) To build a new language model for presage:
a) export all the text you want to use in a text file with encoding ISO-8859-1 (for example, export.txt). Note: For my friend, Cédric, I exported all the mails in my OUTBOX folder
b) generate the model with the following commands:
 "C:\Program Files (x86)\presage\bin\text2ngram.exe"  -n 1 -f sqlite -o my_model.db export.txt
 "C:\Program Files (x86)\presage\bin\text2ngram.exe"  -n 2 -f sqlite -o my_model.db export.txt
 "C:\Program Files (x86)\presage\bin\text2ngram.exe"  -n 3 -f sqlite -o my_model.db export.txt
[...]

PR #128, opened 2015-10-03
Add tolerance for incorrect spelling #69

chadkulesa:
Parts of the DictionaryService class were refactored into OrginalAutoComplete, and NGramAutoComplete was added as an alternative implementation.

N-gram works fairly well across multiple languages for fuzzy searching, and is pretty fast to compute. The exact parameters for some languages might need some adjustments for best results. We should look at how best to allow languages to set these parameters.

The settings hard coded works well for the English language, and most of the [common misspellings](http://www.oxforddictionaries.com/words/common-misspellings) are returned in the top 10 results. 

We might want to look at adjusting the scoring in NGramAutoComplete to better use word frequence.

MapCaptureToEntries hasn't been modified yet, but it shouldn't be too hard to do. N-gram should perform a lot faster than the current method.

The IAutoComplete and IManageAutoComoplete were added as interfaces that the OriginalAutoComplete and NGramAutoComplete classes use. When we start implementing east Asian languages, we'll want to use the built in IME support for those languages in place of the current auto complete and dictionaries.

### DIFF
--- a/src/JuliusSweetland.OptiKey.csproj
+++ b/src/JuliusSweetland.OptiKey.csproj
@@ -255,6 +255,7 @@
     <Compile Include="Services\AutoComplete\IManageAutoComplete.cs" />
     <Compile Include="Services\AutoComplete\OriginalAutoComplete.cs" />
     <Compile Include="Services\AutoComplete\IAutoComplete.cs" />
+    <Compile Include="Services\AutoComplete\NGramAutoComplete.cs" />
     <Compile Include="Services\ICalibrationService.cs" />
     <Compile Include="Services\IMouseOutputService.cs" />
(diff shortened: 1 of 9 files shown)


JuliusSweetland:
I will have a proper look at this asap. Should be this week. Thank you - it looks like an excellent piece of work.

JuliusSweetland:
I'll be looking at your pull request soon. I've got some reading to do on n-grams before I can really understand your implementation.

EddPorter:
I rebased the first two commits as #202

EddPorter:
The commits in this pull request have all been merged into the ngram branch now. This request can be closed.

JuliusSweetland:
All commits merged into new "ngram" branch to be merged into "master" branch soon.

PR #136, opened 2015-10-06
Shortened some German key strings

ninehundred1:


JuliusSweetland:
@ninehundred1 In terms of the symbols on some keys being made smaller due to long text; this is only for strings like SHIFT and SLEEP. The other text that you've shortened is fine and doesn't need to be shortened as it is displayed on a popup where the text size isn't limited in the same way.

I can close this pull request and wait for another where you only impact the strings used on strings, or merge this in if you are still happy with the changes?

ninehundred1:
I actually prefer the shorter strings for the pop up, they are more to the
point and still have the same message in german.
So you can go ahead and merge the current.

PR #138, opened 2015-10-09
updated strings and dictionary

ninehundred1:
This should be pretty much all the changes I would make


ninehundred1:
ok, let me know when another build is ready, but won't be able to run it
until monday.

JuliusSweetland:
@ninehundred1 Ready here: https://github.com/JuliusSweetland/OptiKey/releases/tag/v2.1.1_German_alpha_1

I made a design change - I customise the diacritic keyboards so that the first section of the first page of special characters is customised to the current language; e.g. for German the first section starts äöüß, and then a space, and then the rest of the special characters. French is similarly customised. This means you don't have to go hunting to find the few characters you actually care about.

I plan to speed up access to special characters even more with another change later which allows you to move directly to the diacritics keyboard from any other keyboard, rather than having to go through the menu each time.

ninehundred1:
ok, i will try it out. I was thinking that too, if you need a special
character, it takes a few switches to get there.
Also the multikey is swype, where you can just move from key to key
continuously?

JuliusSweetland:
@ninehundred1 Cool and yes, "multi-key selection" = swype.

ninehundred1:
Ok, fixed a few more strings, now it looks fine.

JuliusSweetland:
Do you want to submit a new pull request and I can prepare another test
version?

ninehundred1:
I think I did submit one.

JuliusSweetland:
I'll take a look after breakfast.

PR #139, opened 2015-10-09
fixed 4 more strings

ninehundred1:
I think that should be all. I don't know about the error messages and all that, but as those are whole sentences, they should be fine and the context should be clear.

### DIFF
--- a/src/Properties/Resources.de-DE.resx
+++ b/src/Properties/Resources.de-DE.resx
@@ -171,7 +171,7 @@ HINZUFÜGEN</value>
 ZURÜCK</value>
   </data>
   <data name="BACK_WORD_SPLIT_WITH_NEWLINE" xml:space="preserve">
-    <value>ZEICHEN
+    <value>WORT
 ZURÜCK</value>
   </data>
   <data name="BASS_UNABLE_TO_CREATE_STREAM_FROM_FILE" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


JuliusSweetland:
Looks good. I have identified 2 more strings which are causing some icons to be displayed smaller than they need to be and have just submitted the following change (breaking "Add to dictionary" and "Re-calibrate" one more time: c2b920006abb006140d3442652d25db5e41a7678

ninehundred1:
So there is a change I need to do with  those? I can just search for those in the German source file?

JuliusSweetland:
You don't need to do anything. If you take a look at that commit that I
linked in my comment you'll see what I changed. I've also been advised by
one of my German users (who I gave an early copy to) to do the following.
Let me know if you disagree, otherwise I'll try to make these changes
tomorrow:

1. The enter key must be labelled " Eingabe" ( spelling error)

2. The back word key should be labelled "Wort Zurück"

3. In the main menue the labels for "Transparenz erhöhen" and " Transparenz
verringern" ( meaning reducing transparency) should be swapped.

4. Please sort the German diacritics like this: ä ö ü ß

ninehundred1:
Did you give your user the commit from last night? I fixed the Wort  
zurück, and will look at the others, which look they need changing. I  
can do that tomorrow.

JuliusSweetland:
He may be commenting on the previous version. I'll take a look tomorrow. At
the rugby

ninehundred1:
I checked the mistakes the german user has reported and made a pull
request. see below (especially 3).

1. The enter key must be labelled " Eingabe" ( spelling error)

-fixed



2. The back word key should be labelled "Wort Zurück"

-fixed that last round



3. In the main menue the labels for "Transparenz erhöhen" and " Transparenz
verringern" ( meaning reducing transparency) should be swapped.

-this is wrong and correct. The strings are correctly translated from the
english original, but he is right, they are swapped. Decrease transparency
in the current case makes the window be more see through, which is actually
an increase in the transparency. So you might have to look at the original
to fix that and probably just swap the GUI button handle around. I didn't
notice because decrease made it less present, which I assumed was correct.



4. Please sort the German diacritics like this: ä ö ü ß

-I assume he means the keyboard, which is not in that order. How can I do
that, is that something Arthur needs to do, it is part of the diacritic
keyboard

JuliusSweetland:
Great stuff. Don't worry about number 4 - that's one for me. I move the
current language diacritics before all the rest and he's just asking me to
change the order a bit. I'll fix that tonight or tomorrow night and prepare
a new pre release.

ninehundred1:
I think after that is should be fine.

JuliusSweetland:
Excellent. I'll try to find time tonight then.

PR #159, opened 2015-10-24
adjust text size on "back" button on conversational keyboard

Ar2rZ:


JuliusSweetland:
Testing this and all strings are now even - good commit Arthur!

PR #162, opened 2015-10-26
Unit tests

padmo:
This completes the tests for MainViewModel.cs I could do, the other part of the partial class in MainViewModel.ServiceEventHandlers.cs should follow.


JuliusSweetland:
Looking good. Merged into master.

I am going to scrap the idea of a separate UnitTests branch. Tests are easy to pull down locally, run and inspect and then commit straight onto master, so shall we do that going forward? Saves a bit of messing about.

padmo:
Perfect! Works for me.

PR #166, opened 2015-10-28
Capture part of behavior AttachServiceEventHandlers on MainViewModel and of ManagementViewModel

padmo:
MainViewModel.HandleFunctionKeySelectionResult is not tested, I don't expect i'll have the courage anytime soon to cover that by tests :)

If you think this important however, you could perhaps point out the most important FunctionKeys cases, which would have the biggest impact when broken.

ManagementViewModel instantiates other view models, so I cannot properly unit test it in isolation.
I think you could benefit from working with an IoC container, but for now I'm not suggesting to do any big refactors until more of the application is covered by tests.


JuliusSweetland:
Merged - thank you very much.

Issue #154, opened 2015-10-23
Splash screen at start up is incorrectly sized


JuliusSweetland:
When the splash screen dimensions are calculated at startup they are not calculated from the current window size, but instead seem to be based on the last size of the window (which can be when it closes).

Steps to reproduce:
1.Configure OptiKey to display splash screens at startup (from the management console)
2.Navigate to the MOUSE keyboard
3.If it is expanded then collapse the keyboard
4.Close OptiKey in the collapsed MOUSE keyboard (click on OptiKey to make sure it is selected and press ALT+F4 to exit).
5.Start OptiKey again and observe the splash screen sized to the collapsed MOUSE keyboard.

EddPorter:
ApplyAndPersistSizeAndPosition is dispatched on the UI thread (from SetAppBarSizeAndPosition) to set the keyboard rectangle, but ShowSplashScreen is already in progress, blocking the resize operation until it completes.

JuliusSweetland:
Sounds reasonable. Is there a straightforward fix? Also, I knew you'd contribute eventually. It was only a matter of time!

EddPorter:
I lied - nothing to do with threading. What is actually happening is multiple calls to `ApplyAndPersistSizeAndPosition`. The first originates from `CoerceSavedStateAndApply` within the `WindowManipulationService`. This uses the previously saved state, for example the `Collapsed` Mouse view, to set the size. Because the priority is `ApplicationIdle`, I think this event sits on some stack waiting to be executed even though the handler for this event hasn't been attached at this point! When the application becomes idle, it triggers the `SizeAndPositionInitialised` event, which creates the Splash Screen.

The second call is from the `MainViewModel` constructor, which initialises the Keyboard based on the `StartupKeyboard` configuration. The default `Alpha` keyboard forces the window back to `Full` mode and calls `ApplyAndPersistSizeAndPosition`. By this time, it's too late.

I'll keep at it.

JuliusSweetland:
Good catch. You're probably right as the dispatch priority call puts the
call on a queue.

EddPorter:
I've created two pull requests - a "global solution" using the screen size to determine the toast size (edf89ad) and another that suppresses the initial size event and ensures a second is always fired (615eef1).

PR #176, opened 2015-11-24
Use screen size to determine toast notification size #154

EddPorter:
I gave up trying to fix the start up order and moved to setting the toast notification popup from the screen size. Will require a change to the user's local settings as the percentage height has changed. This has a more global impact then perhaps you were envisaging.


JuliusSweetland:
Your other fix sorted out the issue, so I'm going to leave this for now. Feel free to give me grief about this decision.

Issue #171, opened 2015-11-06
Clean up old log4net log files


JuliusSweetland:
At the moment OptiKey keeps an infinite number of log files. This is controlled by the log4net.config file in the root of the project. Logs are output to (on my computer): C:\Users\Julius\AppData\Roaming\JuliusSweetland\OptiKey\Logs

SamPlusPlus:
@JuliusSweetland I could probably take look at this, but can you elaborate on the desired behavior. 

For example I tend to do something like 3 rolling files up to 5MB each. Do you have a preference?

JuliusSweetland:
Hi Samuel. Great, thanks. I think the desired behaviour would be one file
per run (which is what I was trying to achieve with the filename mask) and
a maximum of x files to correspond to the last a runs. For arguments sake
I'd say 10 log files should be stored. I don't think file size is
particularly important, but please advise if you have other ideas.

SamPlusPlus:
Just curious, why a log per a run?  Doesn't that make is harder on the user to send you a log file. For example. "Hi Julius, I keep getting an exception randomly but have no idea why it happens" if you ask the user for the log we won't know in which run it happened, or how many times they have run the application since their last crash"

Would it not be better to just have 3 rolling logs of  say 5MB's each, at any given time you could ask the user for 15mb's of text and log information, that should be plenty, and then you could search for the exception. You could also have the application write to the log that it started up.

I think this use case is supported by Log4Net by default, without any special logic. 

In other words whats the reason per a log file per an instance run?

JuliusSweetland:
If OptiKey crashes the latest log is automatically emailed over. It needs
to be as small as possible really.

SamPlusPlus:
@JuliusSweetland I figured you had a reason. Thanks for explaining, I will look into this.

SamPlusPlus:
Looks like I need to update my machine to support this. You using Visual Studio 2015 with .net 4.6?

JuliusSweetland:
.net 4.6, and either visual studio 2012, 2013, or 2015

NikitasKotsolakos:
Hello. First of all, just wanted you to know that you are running a great project and it is the one that finally inspired me to start contributing to open source.

The reason the logfiles are not limited to the number specified in maxSizeRollBackups  is because the date pattern in the filename makes the maxSizeRollBackups property to not work correctly. I removed the date file name and it works fine, see my pull request. (It is still creating one log per application run)

[...]

PR #177, opened 2015-11-27
#171 Clean up old log4net log files 

NikitasKotsolakos:
The date pattern on the logfile name (on log4net.config) caused the maxSizeRollBackups to not work correctly, since it is not supported for dates. Removing it makes things work correctly.

This will still create a logfile per application run as you wanted. However, if for some reason you really need the date to also be on the filename, then let me know to search for another solution or something.

Since I left the staticLogFileName value="true" (as before), now the most recent logfile will be called Optikey.log ,  the previous Optikey.log.1, before that Optikey.log.2 etc.

However, since I do not have experience with Nbug and how it sends the email with the logfile, I could not find whether you have to change something there too so that it sends the correct file (if it decides which log file to send by its name).

Also, while on #171 comments you mentioned "For arguments sake I'd say 10 log files should be stored" , you had set the maxSizeRollBackups to 25, so I decided to keep it as is and let it be 25 (you can change this to 10 if needed).

(For your tests, note it will keep current log + 25, so a total of 26).

### DIFF
--- a/src/JuliusSweetland.OptiKey/log4net.config
+++ b/src/JuliusSweetland.OptiKey/log4net.config
@@ -5,7 +5,7 @@
   </configSections>
   <log4net>
     <appender name="RollingFileAppender" type="log4net.Appender.RollingFileAppender">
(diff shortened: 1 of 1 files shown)


JuliusSweetland:
Merged - thank you!

Issue #181, opened 2015-12-01
Localisation: Dutch


JuliusSweetland:
**1.Transifex translation of UI strings:** 100% - Complete

**2.1.Netherlands Dutch Keyboard layout:**
![image](https://cloud.githubusercontent.com/assets/4722561/11516425/fbe794ce-987a-11e5-995d-345f32fe6f62.png)
**2.2.Belgian Dutch Keyboard Layout:**
![image](https://cloud.githubusercontent.com/assets/4722561/11516455/35feec20-987b-11e5-8449-cdea1a971747.png)

Special characters that need to be easy to get to: ?

**3.Word list:**

**4.Anything else:** There are at least 2 versions of Dutch that we need to consider; Netherlands Dutch and Belgian Dutch. The word lists for the two should be very similar, or the same, but the keyboard layouts differ (Netherlands use QWERTY, while Belgians use AZERTY). The layouts above are taken from google images and need confirmation.

**BRANCH:** 'Dutch' branch - awaiting pull request from @petermorlion

petermorlion:
Great! I'll get on this.

I can not confirm the Netherlands QWERTY keyboard, but the Belgian is more or less correct. Mind you, it is an Apple keyboard which has some subtle differences:

 - It doesn't seem to include the Alt Gr key, which is necessary for certain characters. Although these are mostly keys not often used (maybe @ is an exception).
 - The top left key is usually ² and ³ (respectively non-shift and shift), whereas Apple puts in @ en #

JuliusSweetland:
Hi Peter. I've merged your pull request into the new 'Dutch' OptiKey branch. Looks good. I believe the following are still 'in progress':

1.There is no Netherlands Dutch resource file so selecting this language for the UI results in OptiKey falling back to the English strings.
2.It looks like the use of the French layout for Dutch Belgium meets the Belgium layout above. I have extended this to also use the French Diacritic keyboard layout, which puts the àçéèù characters first.
3.The word lists are a bit short at 11k words.

I think we've touched on these issues before, but it's good to list them out so we can tick them off one at a time.

petermorlion:
Cool! I'll get to work again this week.

Sent from my Windows Phone
________________________________

JuliusSweetland:
I forgot:
[...]

PR #190, opened 2015-12-12
Changes for Dutch, third try :)

petermorlion:
This should address points 1 and 4 from issue #181:

 - Selecting Dutch (Netherlands) now also changes the UI to Dutch. I've used one resource file as written Dutch doesn't differ that much between the Netherlands and Belgium.
 - Some resources were split over multiple lines so the hovertext for keys can increase.

(I hope this pull request works now. I must have done something wrong in GitHub desktop, I'm using the web interface now).


petermorlion:
This should be mergeable now I believe. This time, I get a message telling me I don't have write access, which is what I expected all along. Maybe GitHub desktop doesn't support creating pull requests across forks?

Issue #14, opened 2014-07-15
Write unit tests!


JuliusSweetland:

DanielHutt:
Are there any key places that you would like help with unit testing?

JuliusSweetland:
Hi Dan. Yes! The observables. They are complicated as anything and have
zero coverage (like the rest of this project!). Thank you. Help would be
amazing

DanielHutt:
Awesome. I'll pull down the repo when I get home so that I can hopefully start plucking away at it.

DanielHutt:
Hey Julius, do you think it would be wise for me to take the first step and create a Test project that has all the right files and folders ready to unit test stuff? That way it would be easy for people to just unit test one class and move on.

JuliusSweetland:
Hi Dan,

Yes that sounds like an excellent plan. Which test runner are you thinking
of using?

Regards,
Julius

DanielHutt:
I'm most familiar with NCrunch and NSubstitute.  Any objections or suggestions?

JuliusSweetland:
Both fine, but I (and possibly most others) do not have licenses for those
tools. As long as the tests are thanks via ms test or nunit I think most
people will be alright.

Thanks!

DanielHutt:
Shoot, I didn't even think about that. NSubstitute is free and open source and I've found fairly easy to introduce to new people.  I'll leave any references to NCrunch out of my branch.

JuliusSweetland:
Sounds good, thanks Dan. Can NSubstitue be referenced via NuGet in your new
project? Saves people having to try and find it.

JuliusSweetland:
Actually I'm not even sure if that's possible. Ignore that.

DanielHutt:
I would expect that to be possible.  I'll definitely make it as easy as possible for people to jump in and unit test a very small portion of the code. The lower the barrier of entry, the better!

JuliusSweetland:
Absolutely. Thanks Dan. Really appreciate you getting involved.

chadkulesa:
I've created a test project with nunit, and wrote a couple unit tests. 

The main project is directly under src, Do you want to keep the test project outside of src, or move the main project into a sub folder of src and have the test project be under src as well?

JuliusSweetland:
Looking forward to seeing it. I think sub folders of src would be good.

JuliusSweetland:
[...]

PR #192, opened 2015-12-23
KeyboardKeyDownUpSource tests

LeeCampbell:
Added Unit tests for `KeyboardKeyDownUpSource`.
Introduces `Rx-Testing` and `BDDfy` dependencies to _JuliusSweetland.OptiKey.UnitTests_ project.
Minor refactoring of the `KeyboardKeyDownUpSource` class to delegate some responsibilities out. This enables mocking of the underlying point source and key Up/Down sequences.


LeeCampbell:
Targets #14

LeeCampbell:
Interested in feedback on if the new `IKeyboardHookListener` and implementation belong where they are, or should these be renamed and moved to `services`?

JuliusSweetland:
Hi @LeeCampbell - I agree that the new `IKeyboardHookListener` should be moved to a service - it fits better there.

I merged this change in, but had a few issues when testing. I had to revert the change for the moment. The revert PR is here: https://github.com/OptiKey/OptiKey/pull/195 - where I have also added detail of the problems I encountered.

PR #193, opened 2015-12-24
Added some unit test for some Extension classes.

SamPlusPlus:
I added some unit test for the extension classes. This is hopefully just the start and I can get around to adding more test and upping our code coverage.

The unit test did expose a small bug, and some obvious code duplication which I removed, but I tried not to go to crazy on refactoring just yet, until we have more test.

### DIFF
--- a/src/JuliusSweetland.OptiKey.UnitTests/Extensions/DoubleExtensionsTests.cs
+++ b/src/JuliusSweetland.OptiKey.UnitTests/Extensions/DoubleExtensionsTests.cs
@@ -1,0 +1,24 @@
+﻿using JuliusSweetland.OptiKey.Extensions;
+using NUnit.Framework;
+
+namespace JuliusSweetland.OptiKey.UnitTests.Extensions
+{
+    [TestFixture]
+    public class DoubleExtensionsTests
+    {
+        [Test]
+        public void TestCoerceToUpperLimit()
+        {
+            Assert.AreEqual(10, 100.5.CoerceToUpperLimit(10));
+            Assert.AreEqual(50.5, 50.5.CoerceToUpperLimit(100));
+        }
+
+        [Test]
+        public void TestCoerceToLowerLimit()
+        {
+            Assert.AreEqual(100.5, 100.5.CoerceToLowerLimit(10));
+            Assert.AreEqual(100, 50.5.CoerceToLowerLimit(100));
+        }
+
+    }
+}
--- a/src/JuliusSweetland.OptiKey.UnitTests/Extensions/PointExtensionsTests.cs
+++ b/src/JuliusSweetland.OptiKey.UnitTests/Extensions/PointExtensionsTests.cs
@@ -1,0 +1,27 @@
+﻿using System.Collections.Generic;
+using System.Windows;
+using JuliusSweetland.OptiKey.Extensions;
+using NUnit.Framework;
+
+namespace JuliusSweetland.OptiKey.UnitTests.Extensions
+{
+    [TestFixture]
+    public class PointExtensionsTests
+    {
+        [Test]
+        public void TestCalculateCentrePoint()
+        {
+            var points = new List<Point>
+            {
+                new Point(1.5, 1),
+                new Point(5, 5.1),
+                new Point(10, 10.9)
+            };
+
+            var centrePoint = points.CalculateCentrePoint();
+            Assert.AreEqual(6, centrePoint.X);
+            Assert.AreEqual(6, centrePoint.Y);
+        }
+
+    }
+}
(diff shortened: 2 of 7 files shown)


src/JuliusSweetland.OptiKey.UnitTests/Extensions/PointExtensionsTests.cs
```diff
@@ -0,0 +1,27 @@
...
+using JuliusSweetland.OptiKey.Extensions;
+using NUnit.Framework;
+
+namespace JuliusSweetland.OptiKey.UnitTests.Extensions
+{
+    [TestFixture]
+    public class PointExtensionsTests
+    {
+        [Test]
+        public void TestCalculateCentrePoint()
```
SamPlusPlus: When unit testing I noticed that PointExtensions.CalculateCentrePoint is using Convert.ToInt32  which rounds to an even number.

 According to https://msdn.microsoft.com/en-us/library/ffdk7eyz.aspx

"If value is halfway between two whole numbers, the even number is returned; that is, 4.5 is converted to 4, and 5.5 is converted to 6."

Not sure if this rounding is enough to throw anything off, but I figured I would mention it.

src/JuliusSweetland.OptiKey.UnitTests/Extensions/DoubleExtensionsTests.cs
```diff
@@ -0,0 +1,24 @@
+﻿using JuliusSweetland.OptiKey.Extensions;
+using NUnit.Framework;
+
+namespace JuliusSweetland.OptiKey.UnitTests.Extensions
+{
+    [TestFixture]
+    public class DoubleExtensionsTests
+    {
+        [Test]
+        public void TestCoerceToUpperLimit()
```
SamPlusPlus: It is minor but DoubleExtensions.CoerceToUpperLimit and DoubleExtensions.CoerceToLowerLimit could use Math.Min and Math.Max for readability.

src/JuliusSweetland.OptiKey/Extensions/CharExtensions.cs
```diff
@@ -13,10 +13,10 @@ public static char ToUpperAndRemoveDiacritics(this char c)
             var cAsString = c.ToString(CultureInfo.InvariantCulture);
```
SamPlusPlus: Bug Fix here: cAsString.RemoveDiacritics(); was never actually returning the value. I am not sure this is being using anywhere, but I fixed it anyway.

SamPlusPlus:
Targets #14

src/JuliusSweetland.OptiKey.UnitTests/Extensions/PointExtensionsTests.cs
JuliusSweetland: @SamPlusPlus That's interesting. Rounding to the nearest even number is pretty arbitrary, but I'm not particularly concerned about it as the decision to round in either direction would also be arbitrary. From a practical perspective we're dealing in pixels, so half a pixel is meaningless, and choosing one pixel to the left, right, up, or down shouldn't make much difference. My only concern with explicitly performing a Math.Ceiling would be that the result still needs to be converted to an Int.

If you would like to change this to a Math.Ceiling and then convert to an Int then I'll happily accept the pull request.

src/JuliusSweetland.OptiKey.UnitTests/Extensions/DoubleExtensionsTests.cs
JuliusSweetland: True - I'll accept a PR if you'd like to make this change?

src/JuliusSweetland.OptiKey/Extensions/CharExtensions.cs
JuliusSweetland: Thank you - good spot! The whole method is unused as you say so I'll remove it after merging in your changes to clean up the source a tiny bit.

src/JuliusSweetland.OptiKey.UnitTests/Extensions/DoubleExtensionsTests.cs
JuliusSweetland: Actually I'll add your suggested change now. Thanks.

src/JuliusSweetland.OptiKey.UnitTests/Extensions/PointExtensionsTests.cs
JuliusSweetland: Actually I'll add your suggested change now. Thanks.

PR #194, opened 2015-12-24
KeyValue struct to be immutable 

LeeCampbell:
Making the `KeyValue` _struct_ immutable as per standard guidance*.
This also meant that more static maps needed to be created between `FunctionKeys` and `KeyValue` in the `KeyValues` _class_.

Xaml was updated to use the `KeyValues` instead of `FunctionKeys` directly. _Strings_ use a TypeConverter to map into `KeyValue` types in Xaml.

*References : 
http://stackoverflow.com/questions/3751911/why-are-c-sharp-structs-immutable and https://msdn.microsoft.com/en-us/library/ms229031(v=vs.110).aspx


JuliusSweetland:
@LeeCampbell Really useful, thank you. I had to refactor the C#6 language features as many people (myself included) are still using VS2013 and C#5 - specifically the property initialisation stuff in the KeyValue class. We'll all catch up with you soon enough.

PR #201, opened 2016-01-13
Use QWERTY keyboard for Netherlands

petermorlion:


JuliusSweetland:
Should the Dutch Netherlands alpha keyboard (alpha and conversation alpha) not need the same additional accented characters as Dutch Belgium?

petermorlion:
Nope :) Belgium has three official languages: Dutch, French and German (the last one being a very small community). So we (Belgians) need the French accents more than they do in the Netherlands, where there is only one official language: Dutch. The Dutch-speaking part of Belgium will use the accents way less than the French-speaking part, but we use the same keyboard. Also, the Dutch-speaking part will speak/type French more than the citizens of the Netherlands would.

JuliusSweetland:
Thanks Peter. Would the Dutch Netherlands keyboard users use accented
characters in everyday typing, when typing/speaking Dutch? If they would,
even a little bit, then I think the accented should be included. If Dutch
doesn't use accented at all, and they are only included on physical
keyboards to allow French speakers to type accented characters, then we can
leave them out.

What do you think?

petermorlion:
I'm not sure how they do it in the Netherlands, but Dutch does have accented letters in words: café, à la carte, scène, gêne, hé, etc. There are also conjugations that use ¨ like geüpload.

We should find someone from the Netherlands that know more about how they do it there. I'll see if I can find someone.

JuliusSweetland:
Ok. I may be being too simplistic, but it sounds like the only difference
between Dutch in Belgium and the Netherlands is the keyboard layout. If we
ignore French and German (as the user can switch between them using the
Language selection from the OptiKey menu) then both Dutch Belgium and Dutch
Netherlands should make the same accented characters available (éëïóöü)?

petermorlion:
Well that is what I don't know. The accented characters aren't on the physical keyboards used in the Netherlands, according to [Wikipedia](https://en.wikipedia.org/wiki/QWERTY#Dutch_.28Netherlands.29). :confused:

Sidenote: maybe we should continue discussing this in issue #181 instead of in this pull request? I'll put my comment there too.

PR #202, opened 2016-01-14
Ngrams

EddPorter:
I started looking at the ngram commits from @chadkulesa. I rebased the first two which extract the auto completion into an interface. These were originally 0de54a0 and c596b1f.


JuliusSweetland:
Thanks Edd - could you create a new PR against the branch "ngrams" which I have just created, or are you happy that this should be merged straight into master?

EddPorter:
Done. See #203.

PR #203, opened 2016-01-14
Refactor prior to adding Ngram support

EddPorter:
New pull request - I started looking at the ngram commits from @chadkulesa. I rebased the first two which extract the auto completion into an interface. These were originally 0de54a0 and c596b1f.


JuliusSweetland:
Looks good, thanks. Merged. Enjoy Bath.

PR #204, opened 2016-01-17
Remove C# 6.0 syntax

SamPlusPlus:
I believe at this time we are avoiding c# 6.0 syntax so I removed this.  Also use this constructor to avoid having to declare RGB twice.


JuliusSweetland:
Agree - no C# 6.0 at this time. Good spot.

PR #220, opened 2016-02-03
Add user setting for auto complete mode choice

EddPorter:
Let's finish this branch off.
Three new user resource strings to be translated...


JuliusSweetland:
Is this branch functionally complete then Edd? Are you happy with how the new word completion (using trigrams) performs?

EddPorter:
Well... It makes suggestions and not all of them are crazy. Perhaps we should mark it as "experimental". Have you tried it?

PR #222, opened 2016-02-10
First push to the Greek translation branch

andylamp:
-- Added a new language in the Languages enumeration
-- Fully implented Standard, Conversation keyboards
-- Added required actions in KeyboardHost, KeyValues, FunctionKeys
-- Updated default .resx file to have the Greek (required when selecting a language)


andylamp:
@JuliusSweetland this is the first pull request -- take a look, I think I've done everything as it seems to work. Let me know if you find something missing so that I can take a look at the dictionary issue.

Also for the clean-up of the files located in `$(Appdata)\JuliusSweetland\` I made a custom target clean action of type `AfterClean` which removes the files stored by Optikey in `$(AppData)\JuliusSweetland`, so it's user independent -- additionally it *only* works when you manually clean the OptiKey project and not the whole Solution or UnitTest project.

Let me know what you think.

PR #225, opened 2016-02-13
Refinements in the Greek branch

andylamp:
-- Added required `functionKey` entry for `Greek` in  `Settings.settings` `<dictionary>` sub-tree.
-- Updated `Settings.settings` file to use normal brackets instead of xml coded ones
-- Changed the keyboard layout for landscape (3 x 4 grid) and portrait (4 x 2, maybe do a 4 x 3?) grid while also slotting in Greek in the correct spot.


andylamp:
@JuliusSweetland I've done the changes we talked about yesterday -- the replacement of html tags to normal ones doesn't seem to break anything which is good news. Let me know what you think when you check it out.

andylamp:
@JuliusSweetland Okay, while the application runs fine -- UnitTests fail while serializing that specific part of the XML file. Weird. I'll look into it.

JuliusSweetland:
Thanks. I looked briefly yesterday and I'm confused why the app.config file
on the Greek branch doesn't contain a userSettings section. It is meant to
be auto generated from the settings.settings file and exists on master.

I would think that the section is generated on settings.settings changes,
but I have no idea why it would have been deleted.

andylamp:
@JuliusSweetland Okay, I've fixed it. **BUT** the issue I think for the ugly XML format is how Visual Studio handles these characters; for example if you swap the HTML tags `&gt, &lt` with their "normal" counterparts `<, >` the parsed value of `KeySelectionTriggerFixationCompleteTimesByKeyValues` inside `Settings.settings` file changes... instead of listing the full XML tree-like format that you'd wish it lists only the values inside the brackets as a straight concatenated string... which is weird -- a possible visual studio bug maybe?

JuliusSweetland:
That's annoying. Is the userSettings section generated in your app.config
file? It's there on master, but not Greek, and it is definitely auto
generated by visual studio. I'm not sure if it's significant as the
settings.settings designer file includes the default values also, but I
find it odd that it's missing. Perhaps it's a visual studio version
difference between 2013 and 2015?

andylamp:
@JuliusSweetland I have absolutely no idea... `App.config` does not have that entry. I just manually entered it -- no clue as to why this happens.

JuliusSweetland:
I'll experiment today and see if it's important.

andylamp:
@JuliusSweetland okay let me know if there anything else to do on my end before I start tackling the Dictionary issue.

Let me know when you can ;).

JuliusSweetland:
Can you create a PR with your latest changes and I'll check. Thanks.

andylamp:
@JuliusSweetland I have done that already, it's the latest commit in this thread with hash `82ecefc`.

PR #247, opened 2016-04-15
Changed some special key names (German)

hiwye:
* Changed scroll lock to "Rollen" as it is shown on most german keyboards
* Changed pg up and pg down to "Bild hoch" and "Bild runter" as it is shown on most german keyboards.

### DIFF
--- a/src/JuliusSweetland.OptiKey/Properties/Resources.de-DE.resx
+++ b/src/JuliusSweetland.OptiKey/Properties/Resources.de-DE.resx
@@ -837,12 +837,12 @@ SYMBOLE</value>
     <value>Hauptfenster konnte nicht identifizert werden. Es kann nicht fortgefahren werden</value>
   </data>
   <data name="PG_DN_SPLIT_WITH_NEWLINE" xml:space="preserve">
-    <value>PG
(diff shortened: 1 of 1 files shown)


JuliusSweetland:
@hiwye Thanks Sebastian. The new PgUp and PgDown strings are longer than the previous strings - do they still look ok on the Physical Keys keyboard? I am concerned that they will make all strings across that keyboard smaller. If so the fix is easy - just replace the space in each string with a newline.

hiwye:
Could we just use arrows like these?
http://stackoverflow.com/a/2701226/4713557

JuliusSweetland:
@hiwye Certainly not impossible, but I shied away from using unicode symbols in favour of using SVG graphics (using Paths in WPF) as the latter gave me much more control over sizing. Do the new strings impact the Physical keyboard key labels then?

hiwye:
@JuliusSweetland This now really closely resembles the german keyboard buttons.
Sizing works great aswell.

JuliusSweetland:
@hiwye Ok great. I'll take a look over the weekend and look to include this change in the next release (v2.3.1).

PR #250, opened 2016-04-26
Translation how-to guide overhaul

andylamp:
Give it a first look @JuliusSweetland as it's quite a long and detailed guide -- I tried to be as concise as possible but if I missed the point in some areas do let me know so I can fix it. Syntax may be wrong in some places as it's late and I might have missed something, so check also for that ;).

Let me know what you think.


JuliusSweetland:
@andylamp It looks great from a quick glance, but I'll look over it in detail this week and merge it in. Thank you for the effort - it's very much appreciated.

Issue #149, opened 2015-10-19
Localisation: Spanish


JuliusSweetland:
**1.Transifex translation of UI strings:** 100%

**2.Keyboard layout:**
![image](https://cloud.githubusercontent.com/assets/4722561/13410154/1a9388b6-df2e-11e5-95fb-6e60996626d7.png)
http://ascii-table.com/keyboard.php/071-2

**3.Alphabet:**
a)Letters: 27 individual letters - https://en.wikipedia.org/wiki/Spanish_orthography#Alphabet_in_Spanish
a-z plus ñ (lower and upper case)

b)Diacritics that can be added to individual letters (these will be dead keys): 
acute accent - used on á, é, í, ó, ú (lower and upper cases)
diaeresis - used on ü (lower and upper case)

c)Other keys: No (Catalan uses the apostrophe, but it is rare in modern Spanish so not needed on the main Alpha keyboard)

Total: 29 keys

c)Complexity: *nothing so far...*

**4.Word list:** SpanishSpain.dic on master branch - 71k

**5.Anything else:** "¿" and "¡" characters: Input from Gaspar: "The usage of “ ¿ and ¡” characters has been reduced to formal documents where proper writing is required. For everyday use, we don’t tend to use them (Facebook, chats, whatsapp, informal email), so I wouldn’t be very worried about giving them a very important location."

**6.Other:** Spanish & Catalan will be developed separately - their differences are greater than originally though. New Catalan localisation issue here: https://github.com/OptiKey/OptiKey/issues/235

**BRANCH:** None yet

JuliusSweetland:
From an email discussion with Juan:

I find a very impressive Spanish word list made by a team of La Rioja University lead by Juan L. Varona. It is included at least in winedt: http://www.winedt.org/Dict/

247.049 words no less.

The problem is verb conjugation. I'm not sure if you speak Spanish. If it not the case, let me explain: we have 6 persons (as in English), each one with a suffix. Besides, there are 17 conjugations that make the life of a Spanish learner close to miserable, I guess.

[...]

PR #259, opened 2016-05-13
Spanish

A01221624:
Here is the Spanish addition (#149).
Relevant comments:

- The dictionary is not the one which was on the Candidates folder.
- The current dictionary is of 100k - 1 entries and is sorted by frequency, so the tail can be removed (source is [RAE](http://corpus.rae.es/lfrecuencias.html), the Spanish Royal Academy, so it is as official as it gets).
- An extra row/column was necessary in the language selection menu as it was already full, please let me know if I should arrange them in a different way.
- The diacritics in the keyboard layout are slightly different from the QWERTY es_ES keyboard, but this way the keys could remain as large as possible.

Please let me know if any changes or additions are necessary.

~NaN


JuliusSweetland:
@A01221624 I'll review asap, but the CI has failed the build with this error:
```
C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1820,5): warning MSB3270: There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "Tobii.EyeX.Client.Net20", "x86". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project. [C:\projects\optikey-53pu1\src\JuliusSweetland.OptiKey\JuliusSweetland.OptiKey.csproj]
UI\Controls\KeyboardHost.cs(24,65): error CS0234: The type or namespace name 'Spanish' does not exist in the namespace 'JuliusSweetland.OptiKey.UI.Views.Keyboards' (are you missing an assembly reference?) [C:\projects\optikey-53pu1\src\JuliusSweetland.OptiKey\3nlvw0ag.tmp_proj]
  Restoring NuGet packages...
  To prevent NuGet from downloading packages during b
```

A01221624:
Whoops, sorry about that. Silly me didn't realize the .csproj file was necessary.
Added it and it seems to be passing now.

PR #265, opened 2016-06-01
First merge for Croatian localization

vdjerek:
Everything more or less works, but needs more testing. Selecting the language from within the OptiKey doesn't work, only works from the Alt+M menu


JuliusSweetland:
I have merged and fixed the language keyboard problem.

There are a few key layout changes I will make tomorrow, but it looks good!

PR #290, opened 2016-12-05
Presage beta version for next word prediction

mostlyjason:
So I finished the presage integration as discussed in https://github.com/OptiKey/OptiKey/issues/69! I had to change some behavior for next word prediction. The auto space needs to come with the suggestion so that presage knows the word is finished and it should predict the next word. Also, I changed the back one behavior to literally go back one instead of deleting the whole word. This allows the user to remove the auto space if desired. They can still click back word to remove the entire word. 

I tested the existing functionality and didn't find any regressions, but let me know if you find any. I ran the auto complete prediction project and it works. I wasn't sure how to run the unit tests because when I ran test in visual studio it didn't find any test cases. Let me know how to run the tests if you'd like me to update them.

To run it with presage (which is optional and off by default), you need to install presage 0.9.2 32-bit from here https://sourceforge.net/projects/presage/files/presage/0.9.2~beta/win32/. If you installed as a non-admin you need to add a registry key with your install path at HKEY_CURRENT_USER\Software\Presage. Also change the SuggestionMethod in settings to Presage. I can write more detailed instructions for your wiki if desired. We can make the installation smoother in a future version. 

Let me know if you have any questions or suggestions!


JuliusSweetland:
@mostlyjason Hi Jason - excellent. I'll check this out asap. The build is failing at the moment:

"app.config : error MSB3249: Application Configuration file "app.config" is invalid. Could not find file 'C:\projects\optikey\src\JuliusSweetland.OptiKey\app.config'. [C:\projects\optikey\src\JuliusSweetland.OptiKey\JuliusSweetland.OptiKey.csproj]"

Visual studio will generate an app.config file when you edit the settings.settings file, but you can delete this. I would guess that you've committed the change to the proj file (adding a reference to the app.config file), but not committed the app.config file. To resolve this could you delete the app.config file locally and commit your proj file again?

mostlyjason:
Ahh you're using nunit. I'll have to install it and see if I can get these test cases updated.

mostlyjason:
@JuliusSweetland fixed!

JuliusSweetland:
@mostlyjason It looks really promising - great work! I've found an initial issue though...

From a factory start (so I deleted everything in "C:\Users\USER_NAME\AppData\Roaming\JuliusSweetland") I typed the sentence "Hello my name is Julius". I then spoke that sentence, cleared it, and then started typing "H" for hello again. Correctly it suggested "Hello", which I selected, which was followed by suggestions which included "my", "name", and "Julius", BUT when I selected any of them it did not output the selection and the suggestions remained the same. Can you replicate this?

Really impressed with this though. It's going to be a brilliant addition.

mostlyjason:
@JuliusSweetland I think that should address it. Let me know if you find any more.

JuliusSweetland:
@mostlyjason I've yet to test your last commit (that will most likely be tomorrow evening as I don't have any time before then), but I've reviewed the code. Everything looks good, but I have 2 questions;

1.In PresageSuggestions.cs what are the stubbed out methods such as AddToDictionary()? Are they going to be fleshed out later? Does the fact that they're stubs cause a problem with adding new words to the dictionary?

2.In KeyboardOutputService - I noticed that the autoAddSpace variable is set to false, where it was previously true. Does this impact how OptiKey works if you are not using presage?

mostlyjason:
@JuliusSweetland 

1. In this beta version we have not added support for adding words to the dictionary, but the stubs are there so that we can add that support in a future version.
2. As I mentioned at the top, I changed the behavior so that the auto space is added at the time that the user accepts a suggestion rather than waiting for the next word to add the space. Adding the space allows the next word to be predicted. I figured suggestions are usually whole words, so if they accept one they are ready for the next word. Do you agree with this approach?

aluizmv:
Hi Julius!  Great news! How can I download this beta version?Tks

Enviado do Yahoo Mail no Android 
 
  Em ter, 6 dez, 2016 às 2:48 PM, Jason Skowronski&It;notifications@github.com> escreveu:   
@JuliusSweetland
   
   - In this beta version we have not added support for adding words to the dictionary, but the stubs are there so that we can add that support in a future version.
   - As I mentioned at the top, I changed the behavior so that the auto space is added at the time that the user accepts a suggestion rather than waiting for the next word to add the space. Adding the space allows the next word to be predicted. I figured suggestions are usually whole words, so if they accept one they are ready for the next word. Do you agree with this approach?

mostlyjason:
@aluizmv are you familiar with git or are you hoping for a packaged download? In git, you can download the code from my fork: 
`git clone https://github.com/mostlyjason/OptiKey.git`
or you can fetch the pull request into a local branch

JuliusSweetland:
@mostlyjason I haven't had a chance to try your auto-space and 'back one' changes, but how does it behave if you capture a multi-key selection (a whole word or phrase), or select a suggestion. I presume the auto space is inserted immediately, 'back one' removes that space, and 'back many' removes the auto-space and the preceding word, or does 'back many' only remove the space, but then selecting it again removes the preceding word? I'm trying to think through how this change affects existing users.

aluizmv:
Hi, sorry but I am not familiar with Git. So I would like to download a version for tests. Is it possible? Tks

Enviado do Yahoo Mail no Android 
 
  Em qua, 7 dez, 2016 às 2:20 AM, Jason Skowronski&It;notifications@github.com> escreveu:   
@aluizmv are you familiar with git or are you hoping for a packaged download? In git, you can download the code from my fork:
git clone https://github.com/mostlyjason/OptiKey.git
or you can fetch the pull request into a local branch

mostlyjason:
@JuliusSweetland actually for multikey selections I kept the old behavior where the space is not added because the automatically inserted word may not be the one you`re looking for. This gives you the opportunity to pick a different word from the suggestion list. However, once you pick a word from the suggestion list, it assumes that you are done with that word and ready for the next one. Does that make sense?

@aluizmv I think in that case you will have to wait for Julius to release a new version

JuliusSweetland:
Hi Jason. I've been mulling it over and I feel like the existing behaviour
is more efficient as the space is only ever inserted when we know that the
user wants to continue typing. The problem with always inserting the space
is that the user has to delete the last one in a sentence when a space is
not required, e.g. when a full stop is entered. Would it be possible to
reinstate this behaviour for presage?

JuliusSweetland:
Hi again. There's a new issue in this area also:
https://github.com/OptiKey/OptiKey/issues/291

Have you encountered these behaviors?

Fyi I'm not going to be in front of a computer for a couple of weeks, but i
can look around the new year.

mostlyjason:
@JuliusSweetland I thought about this some more and it might be possible. We'd need to make the ISuggestions interface aware of whether we are asking for a current or next word prediction, and then force Presage to generate the one we want by adding or removing a space in a private local context.  However, it'd be nice if you could do some testing and let me know all required changes before I make any more.  I can help this week but then I'll be out of town until mid January.

JuliusSweetland:
Is it not possible to restore the auto space logic in the output service
and just use passage to populate suggestions? To the output service it
doesn't matter the source of the captured char/string/function key and the
auto space can be inserted using the previous logic, no?

I can test over the Christmas weekend and then we can work together mid
January to complete everything?

mostlyjason:
The space is how presage determines whether to give suggestions for the current word or the next word. If you don't include a space, presage will give suggestions for completions or corrections for the current word. If you do include a space, presage will give suggestions for the next word. If you don't want to add a space in your text editor, but you still want to get a next word suggestion, then you have to give presage a copy of the context text with a "virtual" space added. We could add the virtual space when we are requesting a next word suggestion. This would allow you to keep your existing auto space logic in the output service. Does that make sense?

Sure let me know how your tests go.

JuliusSweetland:
I understand now. I think it would be best if I play with it for a bit and
see how it feels. Shall we catch up in the new year when we're both free to
finish this together?

A01221624:
Sorry for butting in, but I've been following the discussion and have a comment: Data-wise, I believe it would be very useful to have the suggestions include their type (auto-complete, next-word, auto-correct) and use this information to determine necessary actions upon selection (append, delete word, add space).

JuliusSweetland:
Interesting idea. Before presage (using basic or ngram auto complete) all
suggestions were a combination of auto complete and auto correct. If a
suggestion is selected then the other suggestions remain and act to swap
out the previous selected suggestion. With presage the workflow is a bit
different, but there are still only about 5 spaces for suggestions. The
current pull request will first populate suggestions which are auto
completions (and also auto corrections I believe but I can't test this
now). Selecting a suggestion then populates the suggestions with great word
suggestions, rather than alternate auto complete/correct suggestions. If
I'm understanding your idea we would display a the best auto complete/auto
correct as the user is typing a word. Once they select a suggestion the
suggestions would then be a combination of the best alternate auto
complete/correct suggestions AND next word suggestions, with the difference
being obvious. If that's correct then the problem will be lack of space. We
could discuss adding another line of suggestions, with one line for current
word and the other four next word.

Thoughts?

A01221624:
That workflow makes a lot of sense, and if that's the design you're using then the suggestion type information is implicit and not as useful if explicitly included in the suggestion itself.

You're right in that I was thinking of a scenario in which a fixed workflow isn't assumed and different suggestion types might be mixed. To show the most relevant ones, I would consider using an algorithm which sorts them by rank based on heuristics of type and n-gram frequency and obviously only show the best ranked (which might only be auto-correction/auto-complete or a combination of suggestion types). With heuristic of type I mean something as simple as "auto-correction first, then auto-complete, then next-word", which combined with what you already have for ranking based on n-gram frequency, might just work for this.

Of course, saying things is easy and since I don't currently have an algorithm for this, I think that's best left for later, since the workflow you mention is natural and easy to use/understand.

As for separating suggestions by type in different lines, I think that might negatively affect the intuitiveness of your current UI, as well as its current lack of clutter, both of which are things that I really like about it as it is.

JuliusSweetland:
Ok great. I'm going to spend a bit of time testing the new presage pull
request in a couple of weeks time so I'll bear all of these ideas in mind
and then I'll report back here.

mostlyjason:
Yeah I agree with @A01221624 that multiple rows of boxes for suggestions would clutter the UI. I believe we can automatically determine whether the user wants an auto completion or a next word based on what they have typed already. This will be more intuitive because the suggestions will always be relevant. It will also make for a simpler UI. I think we just need to be careful thinking through the cases so that we always offer the right suggestion.

To address Julius's concern, we can make the auto space logic even smarter. If the last input from the user was accepting a suggestion then we should assume they are ready for the next word even if a space has not been added. This way, they don't have to press the back button to insert a period at the end of the sentence.  A space would only be added if start typing the next word.

@JuliusSweetland  sure I'll be happy to help with improvements in mid-January when I return.

JuliusSweetland:
Hi @mostlyjason - I've just had a chance to test the lastest pull request. Is it possible to suppress the auto-added space until the next word is begun (either with a single letter or whole word/phrase)? As discussed I think the optimal solution would be to fake the space with presage so it knows to predict the next word, but only output the space using the old logic, i.e. by keeping tracking of whether an auto-space should be added on the next capture and then injecting it then. What do you think?

mostlyjason:
@JuliusSweetland sure we can give that a try. It will require a few code changes to work and might take some time. @vijayrajanna perhaps the DLL for presage is not loading correctly? Make sure it's installed and the directory matches your registry key.

JuliusSweetland:
Thanks Jason. I look forward to trying it whenever you're ready.

JuliusSweetland:
@mostlyjason FYI I'm closing *this* pull request, but I'm looking forward to the next one with the modified space logic.

PR #293, opened 2016-12-24
Basic continuous integration via Appveyor

Razzeee:
Hey there,

this adds basic ci via appveyor.

1. Get's rid of the old way of using nugget - so please be aware that this is also happening
2. Does a basic Debug|Release setup for appveyor. Building and running tests.
3. Also creates very basic artifacts, just a zip files created from the build files for now.

What you need to do is enable [Appveyor](https://www.appveyor.com/) for your profile and activate it for the main repo. We could also add a build badge to the repo if you would like that.

After that every new merge request should automatically be build by appveyor and it will let you know if tests/build fails.

If you have any questions, problems or else, please let me know.


Razzeee:
Just noted your already using Appveyor, but with the server side config :D
Well this might add some value anyway

JuliusSweetland:
Hi @Razzeee thanks for this. I'm not familiar with the difference between this change and the existing server side config? Would this change the current behaviour?

Razzeee:
Well I obviously don't know what you've configured on your server side.
But it seems like my config does this more:
1. Build artifacts (only zip files for download right now)
2. Cache the packages folder

Other than that be big difference between having the config server side and repo side is, that every fork will be able to do build via appveyor, without the need of having to figure out how to set it up server side first.
So everybody that branches after this can just enable appveyor and will get builds/tests of their commits for free. Even before they PR.

JuliusSweetland:
Sounds good. If I pull it in to the master branch what do I need to do to
start using it? Enable on repo side and disable on server side?

Razzeee:
No, afaik the `appveyor.yml` will just overwrite the other config 
https://www.appveyor.com/docs/build-configuration/#configuring-build

JuliusSweetland:
Brilliant, I'll check this out asap.

JuliusSweetland:
@Razzeee Merged. Thank you for your contribution. Merry Christmas.

Razzeee:
@JuliusSweetland 
Glad I could help. Merry christmas and a happy new year to you.

PR #296, opened 2017-01-02
Adding presage next word prediction

mostlyjason:
This is a new pull request that incorporates updates discussed in #290. It addresses issue #69. This version has reverted to the original auto space behavior where a space is only added after a user has started typing the next word. In order to make this work with presage, we explicitly pass a variable indicating whether we want to predict the next word. If this is true, then we add a hidden virtual space to force the next word prediction. This variable defaults to false which maintains current behavior. 

I'm also adding a new setting called SuggestNextWord where if it's set to true it will show suggestions for the next word, and if it's false it will show alternative completions. It set to false by default which will maintain the current behavior, and should be set to true when using the presage suggestion method. Please test and let me know if you have feedback. 

I'm also cleaning up some duplicated keys found in the setting KeySelectionTriggerFixationCompleteTimesByKeyValues. This was causing a runtime exception.


JuliusSweetland:
@mostlyjason Just found a few minutes to test. Couple of issues/questions:

1.With SuggestNextWord set to false: if I start a new instance of OptiKey the suggested words do not output to the scratchpad, or to external applications, when selected. The suggested words change, but nothing else.
2.Using the CLEAR key changes the suggestions, but selecting CLEAR removes all suggestions. What is the expected behaviour here, as I would expect Presage to suggest the same sentence starting words each time?
3.With SuggestNextWord set to true: issue 1 is fixed. I guess the problem is that no suggestions should be created when OptiKey starts with SuggestNextWord set to false? Issue 2 remains with SuggestNextWord set to true.
4.With SuggestNextWord set to true: my sentence starting suggestions include "Hello". I select this and OptiKey outputs "Hello" correctly (to the scratchpad and external application). The next set of suggestions (for the next word) include "my". Selecting the "my" key does not output a space and then "my" (resulting in "Hello my", but instead replaces "Hello" with "my" both in the scratchpad and external application.

mostlyjason:
Hey @JuliusSweetland I'm back from vacation and can work on this now. 

1. Were you trying this with Presage turned on? I just added a check so that if you are using Presage SuggestNextWord must be true. You should test this with NGram or Basic instead, and it worked for me.
2. Pressing CLEAR twice cleared the suggestions so I changed it to generate suggestions each time it's pressed.
3. Fixed with the check I described in 1
4. Can you describe the exact steps to reproduce? I tried again and it's working for me.  See this video https://www.dropbox.com/s/c4lat7hy2qb9g3o/test%204.mp4?dl=0

JuliusSweetland:
@mostlyjason Apologies for the long absence. I will endeavour to give this the attention it deserves asap. My personal life has been a bit full recently and I've struggled to find the time.

JuliusSweetland:
@mostlyjason One question I have now is whether it is possible to reliably detect the presence of Presage and automatically default the word suggestion settings?

mostlyjason:
@JuliusSweetland no worries glad you have time to look at it.  Let me know if I can help move this forward.  Yes it should be possible to detect presage. However, there may be times they want to turn it off, such as if it doesn't support their desired language. Perhaps we can explore ways to make setup easier when we're ready to move past the beta stage? I was hoping a beta release would let more advanced users test and try it out.

JuliusSweetland:
@zelmon64 While I think of it - would you be interested in looking at the PR later? Almost all of the work is done, and it would allow us to add next word prediction, but it needs some attention. No problem if not, but I've been very impressed with your work so thought I'd ask.

zelmon64:
@JuliusSweetland Thanks, that's very nice of you to say. I'd be happy to have a go at it after MaryTTS. I've used ACAT before and the presage word prediction is a very nice feature. It would be great to have it in OptiKey!

zelmon64:
@JuliusSweetland I'm having some trouble getting this to work. I've built `mostlyjason:master` and installed presage/0.9.2~beta/win32/ (I previously had 0.9.1) but no next word predictions appear. The word completion is working and from what I've read that should enable next word prediction as well. Am I forgetting a setting I need to actively enable? I've tried both debug and release just in case. Thanks for your help.

JuliusSweetland:
I'll have to look tonight. From memory there may be a setting to turn on.
Check in the words tab of the manager studio, or see if you can see it in
the PR's modified files view in the settings file?

zelmon64:
Thanks for the tip. I got it working but by hacking it as I can't seem to find the proper way. I forced the [DictionaryService(Settings.Default.SuggestionMethod](https://github.com/mostlyjason/OptiKey/blob/7d2b37a8c401219ba10160f12633af6598c1068f/src/JuliusSweetland.OptiKey/App.xaml.cs#L137) to use Presage and had to set `Settings.Default.SuggestNextWords = true` nect to it for Presage to work. Now that I know the key settings I should be able to hunt down where they're supposed to be set.

If you do run it later, I think the update message is interfering with the mouse control so `Check for updates` has to be disabled.

JuliusSweetland:
Ah right. That sounds like a bug that I fixed a while ago, but possibly
after this PR. If you disable check for updates for now I think that
problem will go away when this change is merged into the updated master
branch.

zelmon64:
@JuliusSweetland I've found that the `AutoCompleteMethod` (aka `SuggestionMethod`) setting was removed from the management console in eff92858e5bbb5859f3035063473d614fb140445. The "correct" way to enable Presage seems to be to set `SuggestionMethod` to `Presage` in the user settings file and also add the the following lines.

```
            <setting name="SuggestNextWords" serializeAs="String">
                <value>True</value>
            </setting>
```

Is this the desired method or shall I re-add `SuggestionMethod` to the management console? Should I then have it set `SuggestNextWords` automatically according to the chosen `SuggestionMethod` or add it as an additional setting?

JuliusSweetland:
I think re-add SuggestionMethod to the management console. I believe the
reason next word suggestion was a setting was because when enabled it
prevents suggestions on the current word and instead only suggests the next
word. The setting allows you to control that behaviour to say whether you
want Presage to suggest alternate current words or suggest the next word.

zelmon64:
Looking at [DictionaryService.cs](https://github.com/mostlyjason/OptiKey/blob/master/src/JuliusSweetland.OptiKey/Services/DictionaryService.cs#L663-L681) each type of SuggestionMethod only works with a specific `SuggestNextWords` setting. I think I'll go with forcing `SuggestNextWords` to the correct value for the selected SuggestionMethod for now if that's okay?

JuliusSweetland:
Yeah that sounds fine. That area of code has been significantly refactored
so don't worry too much.

zelmon64:
@JuliusSweetland I've rebased against master and it seems to be working okay. I've pushed it onto a new branch here: https://github.com/zelmon64/OptiKey/tree/presage-rebased. Should I now submit a new PR or is there other work to be done first?

JuliusSweetland:
Awesome! The only other part was to bundle Presage with OptiKey. Does it
need to be installed or can it be bundled in a similar way to MaryTTS?

(By the way did you end up bundling MaryTTS, or does the end user have to
install it manually to use it?)

zelmon64:
I left MaryTTS as a separate install since it seemed more of an extra add on that might not be used (like if the user has paid for better voices). It can be packaged with OptiKey instead if you'd like. The question then would be which languages/voices to already include and how much larger it'd make the installation file.

With ACAT it's installer somehow triggers the installer for Presage. That's probably the best way but I have no idea how easy/difficult it is to do.

JuliusSweetland:
Understood. I agree that MaryTTS can be installed manually by the user.
I'll look at bundling the Presage installer. It may not be easy as the
setup tool I use is not particularly customisable and probably can't handle
situations where a different version of Presage is already installed, for
example. For now relying on the user to install Presage or MaryTTS seems
fair.

zelmon64:
[Here's the Presage installer in ACAT](https://github.com/01org/acat/tree/master/src/Setup/PresageInstaller). At face value I'd assume it could be directly spliced into OptiKey's installer.

JuliusSweetland:
That installer looks to just be a form that spans a process to run the
Presage installer. I'm not really sure what the benefit is over just
linking to the Presage installer and asking the user to install it
themselves. I will look at changing my installer so it either informs them
that they can install Presage and change a setting to get next word
prediction, or ideally ask them if they want to install a specific Presage
installer and launch a download and install as part of the same setup
process. I'll have a play when I review this PR.

JuliusSweetland:
Also GitHub is warning that this PR has merge conflicts. I'm happy to
resolve them, but you mentioned that you had updated from master already so
do you know why it's complaining?

zelmon64:
Sorry I hadn't submitted the new PR. It's up now. That sounds like a good idea for the installer.

JuliusSweetland:
Awesome. Thank you.

JuliusSweetland:
Work repacked into PR #337 @mostlyjason thank you for your work on this. It should be released soon!

mostlyjason:
@JuliusSweetland @zelmon64 Nice glad it still worked after all this time! Will be happy to see it published!

PR #299, opened 2017-01-02
Remove unused enums EyeXSensitivities and TaskbarPosition

adolfosilva:


JuliusSweetland:
@adolfosilva Looks good, merging. Thanks.

adolfosilva:
Hi @JuliusSweetland. I would like to contribute a bit more to this project. Add some tests, maybe write some documentation. Would you be able to give some mentoring help?

JuliusSweetland:
@adolfosilva Sure. I can't offer too much time I'm afraid, but I can definitely oversee efforts if that sounds alright by you?

adolfosilva:
Yes great, i understand. I just have a question for now. Why is the Enum + Extension Methods pattern used instead of using a Class?

JuliusSweetland:
Can you give me an example of this and I'll explain the design

adolfosilva:
For instance Enums/FunctionKeys.cs and Extensions/FunctionKeysExtensions.cs

JuliusSweetland:
@adolfosilva I think my decision came from the fact that the value being passed around is just an int. Creating a class to hold a single enum value and a few "helper" methods just didn't seem appropriate to me, although this would be perfectly valid. I tend to use Extensions classes where the method is not necessarily core functionality of the data, but is instead more closely related to how that data is being used.

adolfosilva:
@JuliusSweetland Ah makes sense. Thanks.

Issue #305, opened 2017-01-14
Text not entered when using mouse button selection


mostlyjason:
I have mouse button key selection turned on, and when I click on a key it changes focus to the Optikey window. As a result, text is entered into the scratch pad, but not my target window. An improvement could be to prevent changing focus to the Optikey window. This is similar to how Microsoft's on-screen keyboard works. 

One disadvantage of this approach is that bringing up the settings editor requires pressing Alt-M with the Optikey window in focus. We would need to think of an approach to open the settings window, such as adding a button for it in the menu. 

Do you have other suggestions to address these issues?

JuliusSweetland:
@mostlyjason His Jason - it's a design decision really. If mouse events are passed through OptiKey then you cannot interact with it at all. This means no context menu, or if I wanted to allow the user to drag it around (which you cannot currently do, it's just an example). Is it common to accidentally click on OptiKey and not be able to then click away from it again if it was an accident?

mostlyjason:
I'm not accidentally clicking on Optikey, I'm clicking on the keys in order to type them. I'm actually using a SmartNav head mouse which is a bit more accurate for me than the EyeX. When I click on a key I want to enter that key into my target window but it doesn't have focus so typing does not work.

I believe we would still get mouse events to the window, it's just that when we receive a window activation event we tell the Windows API not to activate it. I believe you would still be able to drag the window around, as you can with the Windows on-screen keyboard. I would have to test the context menu to verify that would work as well.

Here is an example:
```private void Window_Loaded(object sender, RoutedEventArgs e)  
        {  
            HwndSource source = HwndSource.FromHwnd(new WindowInteropHelper(this).Handle);  
            source.AddHook(new HwndSourceHook(WndProc));  
        }  
 
        private IntPtr WndProc(IntPtr hwnd, int msg, IntPtr wParam, IntPtr lParam, ref bool handled)  
        {  
            const int WM_MOUSEACTIVATE = 0x0021;  
            const int MA_NOACTIVATE = 3;  
 
            if (msg == WM_MOUSEACTIVATE)  
            {  
                handled = true;  
                return new IntPtr(MA_NOACTIVATE);  
            }  
 
            return IntPtr.Zero;  
        } 
```

I haven't tested this, but if you're open to the idea I could try it out in a separate branch and let you know how it works.

JuliusSweetland:
Interesting. Yes that sounds like a good idea. Let me know how your testing
goes.

[...]

PR #307, opened 2017-01-21
preventing the change of focus when using mouse button key selection

mostlyjason:
This fix should address issue #305. It will prevent the changing of focus when using the mouse button key trigger source. This will allow Optikey to type into the target window even when the user is clicking on it. 

Using this approach, the user will not be able to open the management console using only the on-screen keyboard. Instead, users should open the management console using the context menu or by clicking Optikey in the taskbar to activate the window and then pressing Alt-M. The latter method is similar to how users can activate the window for Microsoft's on-screen keyboard.

This only affects the behavior when using the mouse button key trigger source. Other trigger sources are not affected.


JuliusSweetland:
Great change - thank you.

Issue #272, opened 2016-07-15
After adding a word and back to the Conversation Mode, the Clear and Add buttons seem shifted to the right


JuliusSweetland:
I have verified that this can happen on Windows 10.

Workaround: exit from conversation mode and enter again.

PR #313, opened 2017-03-02
#272: set keyboard back to the origianl keyboard after adding a word to the dictionary

rsrobinett:
Moved the delegate to set the Keyboard back to the original to run at the end of the yes action.


JuliusSweetland:
Great stuff. Thank you.

Issue #254, opened 2016-04-29
OptiKey disconnects from Eye Tribe Tracker after "Calibration" option in Conversation Mode


mrkiefer:
EyeTribe tracker (original model), OptiKey 2.3.1, Windows 7

While in Conversation mode, if the user performs a "Calibration" routine, the calibration finishes, returns the user back to the Conversation mode in OptiKey, but then the eye tracker is no longer connected. 

Workaround: Go to Manage/Options dialog, press OK, communication is restored.

Issue #309, opened 2017-02-08
OptiKey's "Calibrate" for EyeX is breaking the config


alexandre-mbm:
Bad result: https://github.com/ENE/OptiKey-Config/commit/33f915caee2693f8e760afd5a46b1f011381c881

JuliusSweetland:
@alexandre-mbm How was that config changed? Can you provide more detail.

alexandre-mbm:
After calibration OptiKey shows bad behavior. So I make copy of file and then I make diff it with my backup in Git. The changes are these that you are seeing. I make the comparation of them because at first time that this happend I thought the XML had been corrupted.

JuliusSweetland:
Unable to replicate.

PR #314, opened 2017-03-10
#254: fix eye tracker from disconnecting after calibration

rsrobinett:
Unable to test because I do not have an eye tracker.


alexandre-mbm:
Is this related to #309?

JuliusSweetland:
@alexandre-mbm It's more this one: #314

PR #315, opened 2017-03-13
Fixed restoring window to minimised bug

mark-poscablo:
## Bug description
A scenario happens where a user starts out in Conversation mode, and when the user tries to go back (to the Menu), the window state goes to Minimised. Thus at this point the OptiKey window is shrunk to the size of the minimised state even though the application is at the Menu, and so the user cannot use the Menu at all since it's too small.

This scenario can be reproduced by making OptiKey start out in Conversation mode, going back to the Menu, minimising OptiKey, and then exiting the application the "unusual" way (say right click on the taskbar icon and close the application from there). Opening OptiKey again and try to go back - it should resize to the minimised window state.

This happens because after minimising OptiKey and unusually closing it, the saved MainWindowState in user.config is Minimised, and this then becomes the MainWindowPreviousState when OptiKey changes the window state to Maximised at start up (due to Conversation mode). At this point, when the user goes back, MainWindowPreviousState = Minimised is restored.

## Solution description
During startup (specifically, at the constructor of WindowManipulationService), the window state or previous window state is forced to Docked if either one is minimised (in user.config).

In my local manual testing, the changes prevent the scenario as described above from happening.

### DIFF
--- a/src/JuliusSweetland.OptiKey/Services/WindowManipulationService.cs
+++ b/src/JuliusSweetland.OptiKey/Services/WindowManipulationService.cs
@@ -112,17 +112,23 @@ namespace JuliusSweetland.OptiKey.Services
                 screenBoundsBottomRightInDp.Y - screenBoundsTopLeftInDp.Y);
             Log.DebugFormat("Screen bounds in Dp: {0}", screenBoundsInDp);
 
-            //Force the window state/previous window state to Docked if either is Minimised
-            //This is to avoid restoring the window state to Minimised in certain scenarios
+            //Force the window state/previous window state to non-Minimised/non-Hidden
+            //This is to avoid restoring the window state to Minimised or Hidden in certain scenarios
             var windowState = getWindowState();
             var previousWindowState = getPreviousWindowState();
-            if (windowState == WindowStates.Minimised)
+            if (windowState == WindowStates.Minimised ||
+                windowState == WindowStates.Hidden)
             {
-                saveWindowState(WindowStates.Docked);
-            }
-            if (previousWindowState == WindowStates.Minimised)
-            {
-                savePreviousWindowState(WindowStates.Docked);
+                if (previousWindowState == WindowStates.Minimised ||
+                    previousWindowState == WindowStates.Hidden)
+                {
+                    saveWindowState(WindowStates.Docked);
+                    savePreviousWindowState(WindowStates.Docked);
+                }
+                else
+                {
+                    saveWindowState(previousWindowState);
+                }
             }
 
             CoerceSavedStateAndApply();
--- a/src/JuliusSweetland.OptiKey/Services/WindowManipulationService.cs
+++ b/src/JuliusSweetland.OptiKey/Services/WindowManipulationService.cs
@@ -112,25 +112,7 @@ namespace JuliusSweetland.OptiKey.Services
                 screenBoundsBottomRightInDp.Y - screenBoundsTopLeftInDp.Y);
             Log.DebugFormat("Screen bounds in Dp: {0}", screenBoundsInDp);
 
-            //Force the window state/previous window state to non-Minimised/non-Hidden
-            //This is to avoid restoring the window state to Minimised or Hidden in certain scenarios
-            var windowState = getWindowState();
-            var previousWindowState = getPreviousWindowState();
-            if (windowState == WindowStates.Minimised ||
-                windowState == WindowStates.Hidden)
-            {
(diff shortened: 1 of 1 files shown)


alexandre-mbm:
@mark-poscablo, is this related to the issue #309, is not?

JuliusSweetland:
@mark-poscablo I've just thought - can you extend this to also check for the window state 'Hidden' and revert to Docked also? Minimised is OptiKey parlance for "shrunk down" and Hidden is OptiKey parlance for minimised to the taskbar. Both should be reverted.

Secondly rather than always resetting to Docked, could you check the MainWindowPreviousState for Docked or Floating and use that as the new value. If not either of these then please use Docked. This gives us a chance to reset OptiKey to its previous position, rather than always docking it again.

mark-poscablo:
@alexandre-mbm Sorry, I don't think it's related to #309, or at least I can't see how this might be related to that issue..?

@JuliusSweetland That makes sense, I completely forgot about the Hidden window state. Your second comment also makes sense as we shouldn't force the window state when it's not unnecessary. I'll let you know when I make the additional changes. Thanks!

mark-poscablo:
Just added the additional changes :) let me know if that's better or I should edit it more?

By the way, this change allows a previous state of Maximised also to be restored on selecting Back - is that okay or should I strictly force MainWindowState to Docked if the MainWindowPreviousState is Maximised?

JuliusSweetland:
I think maximised should be reverted to docked or floating (if the previous
state is floating), yes. Good point!

mark-poscablo:
Ah good, but I meant more like what if the previous state is maximised? Should the window be restored to maximised when going back to the menu too? Or is such a scenario even possible in normal usage (MainWindowState = Minimised/Hidden/Maximised, MainWindowPreviousState = Maximised)?

JuliusSweetland:
@mark-poscablo No the window should not be restored back to Maximised. I think the logic when starting the WindowManipulationService should be:

1.If MainWindowState != Docked && MainWindowState != Floating then
If MainWindowPreviousState == Docked or Floating then MainWindowState =MainWindowPreviousState 
Else MainWindowState = Docked

2.If MainWindowPreviousState != Docked && MainWindowPreviousState != Floating then
MainWindowPreviousState  = Docked

alexandre-mbm:
Julius, use code boxes please. It is fuzzy.

1.
```vb
If MainWindowState != Docked && MainWindowState != Floating then
  If MainWindowPreviousState == Docked or Floating then
    MainWindowState = MainWindowPreviousState
Else
  MainWindowState = Docked
```
2.
```vb
If MainWindowPreviousState != Docked && MainWindowPreviousState != Floating then
  MainWindowPreviousState = Docked
```

**?**

mark-poscablo:
Hmm, I think that at this point, it's more prudent to agree on the correct behavior of OptiKey rather than focusing on how the logic should be.

Window state at exit | Previous window state at exit | Restored window state going back to Menu
--- | --- | ---
Minimised/Hidden | Docked | Docked
Minimised/Hidden | Floating | Floating
Minimised/Hidden | Hidden | Docked
Minimised/Hidden | Maximised | Docked
Minimised/Hidden | Minimised | Docked

Is this correct so far? The last three cases seem absurd, but the fourth one for example is the case I brought up in my last comment, and we're trying to protect against that just in case it happens. 

Unfortunately, Julius's logic fails the second case - the window is restored to Docked. :(

JuliusSweetland:
@mark-poscablo I think my logic would agree with your table for the second case, no? I wrote part of the line in shorthand, so written fully the first case is:
'''
If MainWindowState != Docked && MainWindowState != Floating then
  If MainWindowPreviousState == Docked or MainWindowPreviousState == Floating then MainWindowState = MainWindowPreviousState
'''

mark-poscablo:
My apologies, I wasn't careful enough to see that the logical operators on the inner if statement was different from the first one. Thank you so much!

The logic does pass all the tests laid out in my table above. I'll push the newest commit shortly..

mark-poscablo:
There. By the way, I extracted this part of the code out of the constructor to a new function, since I felt like it's getting too big to the point of being a distraction inside the constructor. Would that be okay for you?

JuliusSweetland:
@mark-poscablo New method looks good, and yes I agree that it should be refactored out into a new method. Great work, thanks Mark!

mark-poscablo:
No problem, and thanks for all the help! You were the one who was able to fix it at the very end 😄

Issue #218, opened 2016-01-28
Multiple Dictionaries


EddPorter:
The `DictionaryService` and the `IAutoComplete` implementations both track dictionary entries. The auto complete dictionaries use (potentially) different storage methods for lookup, which is where the complication lies. Ideally, there'd be a single data structure containing the word entries and usage counts. Indexes would be created as need for the different systems that need to access it.

Considerations:
- adding new words
- removing words
- updating usage counts

alexandre-mbm:
is it about sharing data between two or more assessors?

Chopinsky:
I will take a look at this issue. From the face value, the different ways of storing word counts might be the cause of issue #217.

PR #326, opened 2017-07-04
Consolidate data structure to store in-memory dictionary entries

Chopinsky:
Related to issue #218 which mentions the dictionary entries are stored at different level and are more or less in duplicates, hence creating maintenance difficulties. This pull request aims to resolve the complexity in the dictionary entries data structure, such that we only use 1 source of truce as where to find the user entries across `DictionaryService ` and the `IAutoComplete` implementations. 

On a side note, VS community seems to render indent a little differently than GitHub, so there're a couple of files that only the indents were updated (and they look ugly in GitHub, but in VSC they look fine).

### DIFF
--- a/src/JuliusSweetland.OptiKey/Services/AutoComplete/BasicAutoComplete.cs
+++ b/src/JuliusSweetland.OptiKey/Services/AutoComplete/BasicAutoComplete.cs
@@ -9,7 +9,7 @@ namespace JuliusSweetland.OptiKey.Services.AutoComplete
 {
 	public class BasicAutoComplete : IManageAutoComplete
     {
-        private readonly Dictionary<string, HashSet<DictionaryEntry>> entriesForAutoComplete = new Dictionary<string, HashSet<DictionaryEntry>>();
+        private readonly Dictionary<string, HashSet<DictionaryEntry>> entries = new Dictionary<string, HashSet<DictionaryEntry>>();
 
         private static readonly ILog Log = LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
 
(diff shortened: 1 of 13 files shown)


src/JuliusSweetland.OptiKey/Services/AutoComplete/BasicAutoComplete.cs
```diff
@@ -50,24 +50,21 @@ public IEnumerable<string> GetSuggestions(string root)
         }
 
         public void AddEntry(string entry, DictionaryEntry newEntryWithUsageCount)
-        {
-
-            //Also add to entries for auto complete
-            var autoCompleteHash = entry.Normalise(log: false);
+		{
+			if (!string.IsNullOrWhiteSpace(entry) && entry.Contains(" "))
```
JuliusSweetland: Was there a reason for switching around the logic to add the new entry as a phrase and as a normal entry?

JuliusSweetland:
@Chopinsky Good stuff. I've asked one question on this commit - https://github.com/OptiKey/OptiKey/pull/326/commits/41affcd6b0ed35e01fa70d8573b242042ec579ec

Let me know and I'm happy to merge.

JuliusSweetland:
@Chopinsky Hey. I've just tested this and everything is good, except when performing multi-key selections. If you try and multi-key select the word "welcome" you'll get unrelated matches. Could you take a look as otherwise your PR is very useful. Thanks.

I'm closing this PR. Please submit a new one.

src/JuliusSweetland.OptiKey/Services/AutoComplete/BasicAutoComplete.cs
Chopinsky: Thought this would make more sense if we keep the logic order the same between  `BasicAutoComplete` and `NGramAutoComplete`, where the phrase was added into the dictionary first. I will take a look at the multi-select issue. Will PR again once that's fixed.

JuliusSweetland:
Understood and agree. Thanks.

Chopinsky:
Hi @JuliusSweetland, just want to double check with you on what's expected when using multi-key selection: when the feature is on, and if I register every letter (or part of this word) of "welcome", then I would expect the dictionary to auto-complete the whole word, is this correct? 

Because right now, I'm having difficulty in using this feature: when I ran OptiKey on version 2.3.16 from the remote/master branch, after I registered key "w" and "e" (with multi-selection locked down), then the word "we" would auto-complete, and if I continue with key "l", the multi-selection would start finding words starting with "l" or " l", instead of completing the "welcome". Please let me know if I'm misunderstanding the behavior here, or if I were using this feature wrongly.

JuliusSweetland:
To register "welcome" firstly make sure multi key selection is enabled,
then select the "w" key fully. Next you only need to register interest
(look at them/move the cursor over them) in "e" then "l" then "c" then "o"
then "m" and then finally select "e". "Welcome" should be output after you
select the final "e"

Issue #327, opened 2017-07-07
Keyboard in block sectors?


OneBadNinja:
Hey there Julius, 

A family member of mine is an ALS patient and she's having a hard time coping with the keyboard (elderly lady) to communicate properly.

I was thinking it would be much easier for people in her case to have a simpler interface - keyboard blocks to communicate with (perhaps a new communication mode from the menu).

For instance there are initially 6 blocks on the screen, each containing a group of letters. The user picks the block that contains their letter, and then that block is further dissected recursively until there are large blocks each containing a single letter (emphasis on large blocks as she struggles with focusing on the smaller blocks)

There is a video on youtube here that contains a similar functionality, whereas you slowly zoom into the desired letter:

https://www.youtube.com/watch?v=LtSCemhpIWk

Only simple letter blocks would be needed. She finds more detailed menus very confusing, and I imagine could potentially be the same for some others her age.

Would this be feasible to make? Or does the software already have this feature and I've missed it?

Thank you so much,
Joey

zelmon64:
Hi Joey,

To my knowledge this hasn't been done before but should be relatively simple to implement. If you specify the arrangement of letters you want I should be able to do it.

As an already existing alternative there is [this Yes/No build](https://github.com/OptiKey/OptiKey/issues/322#issuecomment-301759584) which is usable but not quite finished.

JuliusSweetland:
@OneBadNinja Hi Joey. No this method of input does not currently exist. If @Zelmon64 is kind enough to add it then that would be great. Otherwise there is work underway to allow users to define their own keyboards. It would be possible to create and share the keyboard you are describing using this new method, but it is not ready for release yet. The work is being done by someone else so I cannot give a timeline I'm afraid.

OneBadNinja:
Hi Julius and Zelmon, thank you for both of your replies. I understand the situation now. Also apologies for getting your name wrong earlier Julius. 

Zelmon I would be sincerely grateful if you would be able to help out. Thank you so much for volunteering! To offer some clarity, I have just designed a small mockup of this simple interface, and I have attached the images below. I hope this isn't too troublesome

## The Main Grid

User selects a block containing letters/symbols and is taken to the next pane.

![pane_1](https://user-images.githubusercontent.com/20068286/27964943-13471db4-634b-11e7-8fa9-367b4fc109af.png)

## Secondary Pane

User now further selects a smaller subsection of the initial letter group. 

[...]

PR #329, opened 2017-07-09
Blocky keyboard for #327

zelmon64:
Added a blocky keyboard where there are much fewer keys and the letters are grouped together. The user must select a group (and subgroup) before the desired key can be pressed. This is mainly for when the user needs to use larger than normal keys. This keyboard can only be accessed by setting it as the startup keyboard.

The setting `EnableQuitKeys` has been added to be able to hide the quit keys so that the only way to quit is by right-clicking on the keyboard and selecting Quit.


JuliusSweetland:
@zelmon64 I've reviewed the PR and have a few changes I would like to make. If I list out my initial thoughts can you tell me what you think and whether you'll have a chance to make them? I am happy to work with you, but I won't be able to get any work done until later this week, or this weekend. It's an excellent PR by the way - these are mainly consistency points to keep the code base as clean as possible. I'm a big believer in consistency so people can start to learn the code base and then guess how other parts work and be correct more often than they are wrong...

1.Rename Blocky to something else, e.g. SimplifiedAlpha. The keyboard could be SimplifiedAlpha and should be located in the English folder as it will need to be localised into other languages if people want to use it. Function keys could be SimplifiedAlphaABC etc. Resources such as BLOCKY_ALPHA_LAYOUT should be renamed also.

2.What do you think about having 2 version of the Simplified keyboard; SimplifiedAlpha and SimplifiedConversationAlpha (in the same way as AlphabeticalAlpha and AlphabeticalConversationAlpha). They could be identical for now with SimplifiedAlpha being used to type into other applications. We could later add a few more features to the SimplifiedAlpha, such as a shift key, but we can discuss this later.

3.I think we need a new setting like the "Use alphabetical keyboard layout" (UseAlphabeticalKeyboardLayout) setting (called UseSimplifiedKeyboardLayout) as the Simplified alpha keyboard is really just an alternative alpha keyboard. This, like the UseAlphabeticalKeyboardLayout setting, should only be available if the keyboard language is English, and impacts the KeyboardHost to return the SimplifiedAlpha keyboard when Alpha is selected, exactly like the UseAlphabeticalKeyboardLayout setting. The Simplified keyboards would therefore not be options in the startup keyboard list, instead you would turn on UseSimplifiedKeyboardLayout ("Use simplify keyboard layout" setting text) and set the startup keyboard to Alpha. If UseSimplifiedKeyboardLayout and UseAlphabeticalKeyboardLayout are both on then UseSimplifiedKeyboardLayout takes precedence.

4.Instead of the very lowest levels of simplified keys being function keys where the handlers directly edit the Text property, they can be normal keys with text, i.e remove BlockyA, BlockyAp, BlockyEx, BlockyPe, etc and just place with keys with Text="A" etc. This impacts a couple of things;

a) Casing. You currently handle this by always outputting uppercase letters, but if the KeyboardOutputService is going to handle the output then we need to force caps locks. This opens the possibility of adding a shift key to the Simplified keyboards to let people control casing, or to use the auto-capitalisation logic that is build in. For now I think if the user turns on the UseSimplifiedKeyboardLayout setting then we should prompt the user to also turn on Settings.Default.ForceCapsLock, and vice versa (prompt user to turn off ForceCapsLock when they turn of UseSimplifiedKeyboardLayout). When UseSimplifiedKeyboardLayout setting is turned on then also prompt the user to change the startup keyboard to ConversationAlpha as that is probably what is desired.

image

b) Need new logic to determine which keyboard to return to after outputting the character. This can be moved into the MainViewModel -  you would need to change the IKeyboardOutputService to implement INotifyPropertyChanged (which it does already) then monitor changes to the Text property - if the .LastOrDefault() character is not null and is a letter (char.IsLetter) then return to SimplifiedAlpha, else if it is char.IsNumber then return to the main SimplifiedNumeric keyboard. Symbols might have to be case by case as they are split over those two keyboards.

4.Change casing on Enable_Quit_Keys_LABEL resource to all uppercase for consistency.

5.Rename BlockyLayer setting to "SimplifiedKeyboardCurrentContext"

zelmon64:
@JuliusSweetland I think these are all good points and I should be able to start on them tomorrow. I especially like the part about making it similar to AlphabeticalAlpha. I was trying to think how best to incorporate it more.

JuliusSweetland:
Great stuff. I can help later this week.

zelmon64:
@JuliusSweetland I've done everything on your list apart from prompt the user to turn ForceCapsLock on or off because I'm not sure if it's really necessary as I think the key works quite nicely with ForceCapsLock off. 

I've set all the letter and punctuation keys to return to the main layer and the number keys return to the number layer.

Here's a build in case you (or anyone else) want to try it before merging.
https://drive.google.com/open?id=0B7SYYOqNW4FSOTBNcjlYQnlub1U

JuliusSweetland:
Awesome. Thank you. I'll check it out this evening.

src/JuliusSweetland.OptiKey/UI/ViewModels/Keyboards/SimplifiedAlpha.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using JuliusSweetland.OptiKey.UI.ViewModels.Keyboards.Base;
+
+namespace JuliusSweetland.OptiKey.UI.ViewModels.Keyboards
+{
+    public class SimplifiedAlpha : BackActionKeyboard, IConversationKeyboard
+    {
+        public SimplifiedAlpha(Action backAction)
+            : base(backAction, simulateKeyStrokes: false, multiKeySelectionSupported: true)
```
JuliusSweetland: multiKeySelectionSupported: false

src/JuliusSweetland.OptiKey/UI/ViewModels/Keyboards/SimplifiedConversationAlpha.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using JuliusSweetland.OptiKey.UI.ViewModels.Keyboards.Base;
+
+namespace JuliusSweetland.OptiKey.UI.ViewModels.Keyboards
+{
+    public class SimplifiedConversationAlpha : BackActionKeyboard, IConversationKeyboard
+    {
+        public SimplifiedConversationAlpha(Action backAction)
+            : base(backAction, simulateKeyStrokes: false, multiKeySelectionSupported: true)
```
JuliusSweetland: multiKeySelectionSupported: false

src/JuliusSweetland.OptiKey/Enums/Keyboards.cs
```diff
@@ -18,6 +18,7 @@ public enum Keyboards
         NumericAndSymbols2,
         NumericAndSymbols3,
         PhysicalKeys,
+        SimplifiedAlpha,
```
JuliusSweetland: Is this required?

zelmon64: No it's not. Thanks for spotting it and the others.

JuliusSweetland:
You're quick! I was leaving a few comments for myself as I take a break to
eat. I'll continue in a bit. Looks really good though. Great quality.

src/JuliusSweetland.OptiKey/UI/Views/Keyboards/English/SimplifiedAlpha.xaml
```diff
@@ -0,0 +1,1285 @@
...
+                                          Text="{x:Static resx:Resources.RE_CALIBRATE}"
+                                          SharedSizeGroup="KeyWithSymbol"
+                                          WidthSpan="2" 
+                                          Value="{x:Static models:KeyValues.CalibrateKey}"/>
+                        </Setter.Value>
+                    </Setter>
+                    <Style.Triggers>
+                        <MultiDataTrigger>
+                            <MultiDataTrigger.Conditions>
+                                <Condition Binding="{Binding Source={x:Static properties:Settings.Default}, Path=ForceCapsLock}" Value="False" />
```
JuliusSweetland: Is this a mistake? I can't see why the Calibrate key's visibility is linked to the ForceCapsLock setting. Don't worry about submitting a commit - I will remove if it's not linked - can you just let me know? This is also the same on the SimplifiedConversationAlpha (on 2 Calibration keys).

zelmon64: It's not a mistake. When ForceCapsLock is disabled the shift key uses some of the space the Calibrate key is in. It's the same for the Conversation Confirm keyboard key.

JuliusSweetland: Ah, thanks for clarifying

src/JuliusSweetland.OptiKey/UI/Views/Keyboards/English/SimplifiedConversationAlpha.xaml
```diff
@@ -0,0 +1,1249 @@
...
+                        Grid.RowSpan="1" Grid.ColumnSpan="1">
+            <ContentControl.Style>
+                <Style TargetType="{x:Type ContentControl}">
+                    <Setter Property="Visibility" Value="Hidden" />
+                    <Setter Property="Content">
+                        <Setter.Value>
+                            <controls:Key SymbolGeometry="{StaticResource CalibrateIcon}"
+                                          Text="{x:Static resx:Resources.RE_CALIBRATE}"
+                                          SharedSizeGroup="KeyWithSymbol"
+                                          WidthSpan="2"
```
JuliusSweetland: Should this key be WidthSpan="2"? The other version of this key actually spans 2 columns, this one only spans 1 column.

src/JuliusSweetland.OptiKey/UI/Views/Keyboards/English/SimplifiedConversationAlpha.xaml
```diff
@@ -0,0 +1,1249 @@
...
+                                <Condition Binding="{Binding Source={x:Static properties:Settings.Default}, Path=ConversationConfirmEnable}" Value="True" />
+                            </MultiDataTrigger.Conditions>
+                            <Setter Property="Visibility" Value="Visible" />
+                        </MultiDataTrigger>
+                    </Style.Triggers>
+                </Style>
+            </ContentControl.Style>
+        </ContentControl>
+
+        <ContentControl Grid.Row="3" Grid.Column="0"
```
JuliusSweetland: This is a duplicate key - removing.

src/JuliusSweetland.OptiKey/UI/Views/Keyboards/English/SimplifiedConversationAlpha.xaml
zelmon64: I think you're right and I simply forgot to delete it when I copied it.

JuliusSweetland:
@zelmon64 Merged. An excellent pull request - thank you very much! If there's anything else you would like to work on at any time please let me know.

JuliusSweetland:
@zelmon64 This has been released as part of v2.4.0 - https://github.com/OptiKey/OptiKey/releases/tag/v2.4.0

Issue #334, opened 2017-07-22
Add support for MaryTTS


zelmon64:
[MaryTTS](https://github.com/marytts/marytts) is an alternative speech synthesizer that uses a java server with an HTTP interface instead of SAPI5. It has quite a few languages with rather good sounding voices. It comes with .bat files to start a GUI to install the desired voices and run the server. It's also possible to run MaryTTS as a [socket server](https://github.com/marytts/marytts/tree/master/doc/examples/client/c%2B%2B) so that it doesn't run it's HTTP interface (and uses less resources I think) but I haven't yet figured out how to interface with this.

A quick and dirty integration has been accomplished in 3e38a25f990ed48db01ce2e9e8c930af9d27c2c9 where SoundPlayer has been used to retrieve the speech from the MarryTTS server. In this first attempt speechSynthesiser has simply been replaced with SoundPlayer in the `Speak` function using an `if` statement to choose which is active. All the voice options are currently hard coded but it should be simple to change them to user variables.

There are two drawbacks with the current hacked solution

1. The MarTTS server must be manually started (and closed). I'm not sure if it will be possible to have MaryTTS launch in the background when OptiKey starts (and close when OptKey is closed).
2. The Speak key remains blue after it has been pressed (it returns to white if an alpha/num/sym key is pressed)

zelmon64:
I've managed to retrieve the list of available voices from MaryTTS (842beb686d59378260e1347c9bb238bf31aa7a5b) but for some reason they appear with an extra line in the managment console ComboBox as seen below. It doesn't seem to cause a problem but it's visually displeasing. When I retrieved a similar list for just the locales is didn't have these extra lines so I'm not sure what's causing them

![image](https://user-images.githubusercontent.com/15001389/28498572-4b68b5a0-6f98-11e7-85ff-dfb222fb960e.png)

zelmon64:
@JuliusSweetland I been trying to get the speak key to behave normally when used with MaryTTS but so far what I've tried hasn't completely worked. Here's some of what I've tried and the results:

1. Without any modification the speak key will remain blue until another key is pressed.
2. By making `SpeakNewOrInterruptCurrentSpeech` return false when MaryTTS is used the speak key will immediately turn back to white. 
    1) With asynchronous playback of the voice this gives no visual feedback for when the speech ends.
    2) With synchronous playback of the voice this freezes OptiKey until the speech ends.
[...]

PR #335, opened 2017-07-25
Add support for MaryTTS as discussed in #334

zelmon64:
[MaryTTS](https://github.com/marytts/marytts) is an alternative speech synthesizer that uses a java server with an HTTP interface instead of SAPI5. It has quite a few languages with rather good sounding voices. It comes with .bat files to start a GUI to install the desired voices and run the server.

To use MaryTTS with OptiKey get [marytts-5.2.zip](https://github.com/marytts/marytts/releases/download/v5.2/marytts-5.2.zip) (or later) and extract it to anywhere (such as "C:\MaryTTS\marytts-5.2\"). Then in Optikey's management console, under the Sounds tab, check `Use MaryTTS speech synthesizer` and press `FIND` to navigate to the unpacked `marytts-server.bat` ("C:\MaryTTS\marytts-5.2\bin\marytts-server.bat" in the example). After Optikey restarts a minimised cmd window will also be started containing the MaryTTS server and the available MaryTTS voices will be listed in the management console in the Sounds tab under `MaryTTS voice:`.

To install more MaryTTS voices run `marytts-component-installer.bat` ("C:\MaryTTS\marytts-5.2\bin\marytts-component-installer.bat" in the example).

### DIFF
--- a/src/JuliusSweetland.OptiKey/App.xaml.cs
+++ b/src/JuliusSweetland.OptiKey/App.xaml.cs
@@ -160,15 +160,34 @@ namespace JuliusSweetland.OptiKey
                     {
                         StartInfo = new ProcessStartInfo
                         {
-                            FileName = @".\Resources\marytts\bin\marytts-server.bat",
                             UseShellExecute = true,
-                            WindowStyle = ProcessWindowStyle.Minimized, // cannot close it if hidden
+                            WindowStyle = ProcessWindowStyle.Minimized, // cannot close it if set to hidden
                             CreateNoWindow = true
                         }
                     };
-                    proc.Start();
-                    Log.InfoFormat("Started MaryTTS in '{0}' with process name '{1}'.", proc.StartInfo.FileName, proc.ProcessName);
-                    CloseMaryTTSOnApplicationExit(proc);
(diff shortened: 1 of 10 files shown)


zelmon64:
Sorry I had a change to "src\JuliusSweetland.OptiKey\JuliusSweetland.OptiKey.csproj" which wasn't supposed to be there (line 915 `<LastGenOutput>Resources.Designer.cs</LastGenOutput>` got deleted somehow). It's fixed now.

JuliusSweetland:
@zelmon64 I've only started looking into this PR and have a question for you; if you turn on exception handling in Visual Studio for ContextSwitchDeadlock do you encounter an exception on startup? If I prevent catching this exception in VS then everything looks fine, but this exception suggests something long running on the UI thread. I'll look further when I get a chance...

zelmon64:
Is this the correct way to turn on exception handling in Visual Studio for ContextSwitchDeadlock?
image


I don't seem to be having any errors. Where would they appear?

JuliusSweetland:
That's correct. If you we're encountering the exception it would be caught
by visual studio and displayed. I'll look into who I'm seeing it. Thanks.

JuliusSweetland:
@zelmon64 Awesome PR. Merged. Thank you.

PR #337, opened 2017-07-26
Presage next word prediction

zelmon64:
This is an updated version of #296.

To use Presage select it as the suggestion method in the words tab of the management console.


JuliusSweetland:
Does #296 need to be merged for this to work? I presume not and that this is a stand alone PR.

zelmon64:
I believe this should work like any other stand alone PR.

zelmon64:
I've added a registry check for Presage if it is the selected suggestion method. It switches to Basic if it's not present. Only `presage-0.9.2~beta20150909-32bit` works but I haven't found a way to catch if a different version is installed (currently it just crashes).

JuliusSweetland:
Is it possible to catch the exception when checking if a different Presage
version is installed?

zelmon64:
That registry check returns the installation path of Presage ("C:\Program Files (x86)\presage") which has no version information (apart from the fact that it's a 32bit version on a x64 system). 

I just Looked in the registry and that value is the default but there's another that's named `Start Menu Folder` which contains `presage-0.9.2~beta20150909`. I think these two values can hopefully be combined to know that the version is `presage-0.9.2~beta20150909` and 32bit.

JuliusSweetland:
That sounds reasonable. What is the minimum, or exact version of Presage
that is required?

Also what causes the crash?

zelmon64:
That worked. I've squashed it into the last commit as it has the same purpose.

Only `presage-0.9.2~beta20150909-32bit` will work. `PresageSuggestions` was giving errors if `presage-0.9.2~beta20150909-32bit` isn't installed (either registry error is it's not present or problems with presage.net.dll if a different version).

JuliusSweetland:
Good stuff. I'll check the PR either tomorrow or over the weekend. Thanks
William!

JuliusSweetland:
@zelmon64 Hi William. I've merged the MaryTTS PR (what an excellent TTS engine by the way - great find!). This PR needs some manual intervention after I reverted commit 44d9f60f718b44dfb6bff62949f24d71d00412a8 and merged your other PR. I've had 2 attempts at resolving the conflicts, but I am not 100% sure which changes in the dictionary classes need to be kept. Any chance you can update this PR against the top of the master branch?

zelmon64:
@JuliusSweetland Sure. I'll tackle it tomorrow. If worst come to the worst I'll just rebase the original PR again.

JuliusSweetland:
Great thank you.

zelmon64:
@JuliusSweetland It was rather a tangled mess but I think I've gotten through it in one piece (more or less). Had to tweak some bits in the tests to get them to work. Is it okay that Presage doesn't have a test?

JuliusSweetland:
Yes that's ok. There's no test coverage policy on this project!

Thank you William. I'll review asap.

JuliusSweetland:
@zelmon64 Great PR, but I still cannot get Presage to behave properly. When you test typing a sentence using Presage with SuggestWords=true does it behave itself for you? For me If I write something like "test" and then other words the suggestions are not updated correctly and still reflect "test" suggestions. What happens for you? This is the same sort of error I was getting previously...thanks.

JuliusSweetland:
@zelmon64 FYI the PR now has merge conflicts as I have refactored a few bits.

zelmon64:
@JuliusSweetland That doesn't happen for me. When I start typing a new word the suggestion becomes possible completions of the new word instead of suggestions for the next word. 

Each suggestion method only has one possible setting of `SuggestWords` so I've hard set it to the appropriate mode. Maybe there's a problem with doing that.

JuliusSweetland:
Thanks for three info. I'll have to eat everything and try again.

zelmon64:
> eat everything
Not quite sure what you meant by that. I just cleared my settings in case there were any ghosts but it still works fine for me. 

The Presage dictionary seems a bit small. For instance it didn't seem to have the word `today` but after I typed it, it now comes up as a suggestion if I type `to`.

Shall I try to rebase these new refactors or wait till we get to the bottom of your problem?

JuliusSweetland:
Sorry I'm on mobile so I meant to write "reset".

If you don't mind rebasing that would be great!

That's odd about the missing words. I thought Presage was primed with the
same dictionary from OptiKey?

zelmon64:
@JuliusSweetland I've done the rebase but noticed the speak key no longer stops the speech when it is part way through. Was it probably something in that refactoring or had I just not noticed before? Should I hunt it down and stick a fix into this PR?

I like how you've changed the voice selection to only display the active engine.

JuliusSweetland:
I didn't actively change the way speech is stopped, no. That sounds odd.

Thanks - trying to keep it as simple as possible for the user. It's complex
enough!

zelmon64:
@JuliusSweetland I've just found that while the Presage word prediction works fine for me the NGram and Basic methods act as you described earlier. All but the first suggestion box are unrelated to the current word being typed. If I type a long word some of the suggesions become relevant.

I wonder if what though of as the Presage dictionary being limited is actually the revers of this issue. With long words the suggestions stop after a few characters whereas with NGram and Basic the suggestions start after a few characters?

It also seems not possible to deactivate MaryTTS using the management console.

What a mess I seem to have made :(

JuliusSweetland:
You haven't made a mess at all. On the contrary  you've produced some of
the best PRs this project has seen.

I disagree with not being able to disable MaryTTS. I tested earlier today
being able to enable and disable it. Can you test again and let me know why
you think things are broken?

Regarding Presage - I'll have a play tomorrow and see how it is behaving
with each suggestion method. From a quick test earlier it seemed that basic
and ngram were fine, but Presage was behaving incorrectly. I'll test again
and give more details.

zelmon64:
@JuliusSweetland That's very kind of you to say.

Yesterday when I would disable MaryTTS in the management console, OptiKey would restart with it still enabled. Today it woks fine again so I don't know what was the problem. The MaryTTS speech still won't stop when I press the speak key a second time though.

When I couldn't disable MaryTTS I also tried to set it's path incorrectly so that it would automatically disable but that also didn't work either. That is also fine now.

Today ngram still only gives me suggestions for the first word and basic is not giving any suggestions. Presage is still only giving suggestions during the first few letters of each word.

JuliusSweetland:
Ok. I'll test all of this today and let you know how it is behaving for me.

JuliusSweetland:
Basic:
image


NGram:
image

image

Dictionary words missing - e.g. typing "Julius" - after the "J":
image

after the "u" (Julius has not dropped off the suggestions - even if you scroll to the next page - incidentally there are only 7 suggestions in total, including suggestion 1 which is a mirror of the current word):
image



image

zelmon64:
@JuliusSweetland Thanks for all that.

### MaryTTS

1. Restarting outside VS was probably my problem.

2. Perhaps my original method of retrieving the speech time and asynchronously waiting for that duration instead of synchronously playing the speech may work if I can't find why this no longer stops the speech.

3. With all the rebases a few variables were duplicated. Thanks for spotting that one.

4. I'll keep it in mind.

---

### Presage

This is what I was getting as well.

I've found the reason basic and ngram was broken was because the whole text was being sent to get the suggestion instead of just the last word. The whole text is needed for Presage which is why that was working. It was a simple fix and now basic and ngram both seem to work okay again for me.

I think the default Presage suggestion count is six. This is almost certainly why we're only getting seven suggestion in total. It's definitely capable of giving more I'll just have to find where the setting goes. 

I'm not sure yet why it seems to give up suggesting after four letters have been typed.

zelmon64:
@JuliusSweetland I've found the Presage variables `SUGGESTIONS` and `REPEAT_SUGGESTIONS` control the number of given suggestion and whether the same suggestion is to be offered in subsequent predictions. The latter has a default of no which is why in your example Julius disappears from the suggestions. I've found there are functions to check, change and save the settings but for some reason I haven't been able to use them yet. My play is to check the settings after the registry check and change them accordingly. For the time being if you want to see the correct behaviour you can manually change the settings in `"C:\Program Files (x86)\presage\etc\presage.xml"`.

JuliusSweetland:
Excellent finds. I think we could provide a custom settings file for
Presage as part of the installation instructions, rather than try to change
the file.

zelmon64:
That would be easier. I'll push the basic/ngram fix now then.

JuliusSweetland:
Great thank you

zelmon64:
@JuliusSweetland I tried [my old method of querying how long the MaryTTS speech will take](https://github.com/zelmon64/OptiKey/blob/fe8ca76f0011fab4338a1d4aa1699225c2cfceb6/src/JuliusSweetland.OptiKey/Services/AudioService.cs) and it works. Should I push it into this PR or try more to find out why the `Task.Run(() => player.PlaySync());` is no longer working?

JuliusSweetland:
Using the task.run method how were you cancelling the playback?

zelmon64:
It's the [same stopping method](https://github.com/OptiKey/OptiKey/blob/master/src/JuliusSweetland.OptiKey/Services/AudioService.cs#L188-L191) but a different playing method (`PlaySync` vs `Play`). It may be that I forgot to test stopping the speech with the task.run method and only checked that the speak key stayed blue while speaking.

JuliusSweetland:
Is there a recommended way to cancel in progress MaryTTS speech once
started? The only difference that I would guess between play and playsync
is that the second would be blocking. It's there a cancellation mechanism?

zelmon64:
MaryTTS just produces a .wav audio file which is played by SoundPlayer in OptiKey. I think you are correct that playsync is blocking which is why it can't be stopped.

JuliusSweetland:
Oh I see. Leave that one with me. I'll have a look if I can tonight

JuliusSweetland:
@zelmon64 I've just committed a couple of changes to the OptiKey master branch which includes changes to the AudioService, as well as a new PInvoke method, a new Sound class, and a new SoundPlayerEx which includes a new SoundFinished event. I've wired it all up, but the Sound.GetSoundLength method is returning 0 length for the MaryTTS sound location. As you've mentioned that you got a version of this type of thing working would you take a look? I'm hoping your method can be dropped into the Sound.GetSoundLength method to make this all work?

zelmon64:
@JuliusSweetland very interesting. I've managed to incorporate my method for getting the MaryTTS speech duration into Sound.GetSoundLength. I've put it here: https://github.com/zelmon64/OptiKey/blob/3a78339e519f01cf80e76c458b8940a1547d3a28/src/JuliusSweetland.OptiKey/Static/Sound.cs#L16-L42

It works in turning the speak key white after the speech ends and can stop the speech mid way but the speech can't be restarted if it was stopped manually. Do you want to handle it or shall I look into it?

JuliusSweetland:
If you have a moment to look that would be great as I won't be able to look
until tonight.

zelmon64:
@JuliusSweetland I found the problem was that the CancellationTokenSource/CancellationToken stays set to cancelled after it is triggered so it must be set to a new instance. I've pushed a fix into PR #338. In trying to find the problem I added a few debug logs and I've left a few in. I hope that's okay.

JuliusSweetland:
Excellent. No problem with additional logs, especially if there debug logs.

zelmon64:
@JuliusSweetland I did a rebase on the current master and I didn't have to add any fixes afterwards! Woo hoo! It all seems to be working well.

JuliusSweetland:
Fantastic. Thank you. I'll look at the Presage PR asap.

zelmon64:
@JuliusSweetland I just tried multikey input with presage and after I turned multikey back off the suggestions would override the last word instead of adding a new word unless I pressed space first. I'm not sure what the desired functionality should be.

JuliusSweetland:
Are the suggestions for the next word? If so the desired behaviour would be
to not override the current word. What do you think?

zelmon64:
Yes the suggestions are for the next word. That makes sense.

zelmon64:
@JuliusSweetland I think the problem was with `lastTextChangeWasMultiKey` getting stuck as true. I've push a fix that seems to work okay for what I've tested.

JuliusSweetland:
@zelmon64 Great PR - merged. Thank you.

JuliusSweetland:
@zelmon64 Released in 2.7.0. Instructions on using Presage are here if you're interested (can you sanity check?!): https://github.com/OptiKey/OptiKey/wiki/Next-word-prediction

zelmon64:
@JuliusSweetland Great! The instructions look good to me.

PR #338, opened 2017-07-30
MaryTTS Sound.GetSoundLength fix

zelmon64:
Fix for Sound.GetSoundLength to retrieve the duration of the requested speech from MaryTTS.


JuliusSweetland:
@zelmon64 Looks good to me. Thank you!

Issue #332, opened 2017-07-16
Performance Issues associated with DictionaryService's Disk I/Os. 


Chopinsky:
While working in `DictionaryService`, 2 performance issues have come into my attention:

1. Implementation of `SaveUserDictionaryToFile` seems to append the entire in-memory entries to the end of the user dictionary file stored on disk indiscriminately, rather than updating the entry usage of existing entries. This will cause performance issue over time -- the user dictionary file will be filled with duplicate entries and grow dramatically large after some time. Since the user dictionary file will be loaded every time at the launch of the app, the launching period will grow over period of use as well. Functional-wise, the entry usage won't be updated correctly, because when we load entries from user dictionary file, we will only write the 1st encounter of the entry along with its usage into the memory, but we will not update the entries usage when encountering newer and more up-to-date usages of the same entry.

2. The same function `SaveUserDictionaryToFile` is called whenever entry usage is updated, while this is a synchronize disk I/O, it's pretty expensive and could block UI while updating user dictionary, especially given problem 1, where we write the entire memory back to disk. This won't be noticeable on a i7/16gb/SSD machine, but for a less superior machine, the lagging will be more of a problem. 

So I can envision a solution contains 2 parts: 1. eliminate duplicate logging of the same entry in the user dictionary; 2. call `SaveUserDictionaryToFile` in another thread so we won't block the UI.

JuliusSweetland:
@Chopinsky Hi Jacob - great spot. I completely agree on both suggestions. Can you make these changes?

Chopinsky:
Sure, this's on my to-do list :)

PR #330, opened 2017-07-16
Simplify Dictionary Data Structure

Chopinsky:
This PR aims to resolve issues of storing user entries and language dictionaries in 2 different levels, which will double the maintenance effort as well as storage space. Now entries storage has been all moved into `IAutoComplete` implementations and are exposed with proper APIs, `DictionaryService` will keep a reference to this storage for convenience use only.

PR #339, opened 2017-08-01
Dictionary Service Updates

Chopinsky:
This PR is a combine of PR #330, fixes to resolve the long loading time, and fixes to the dictionary I/O problems as mentioned in Issues #332. 

This update include changes in the following area:
- As part of PR #330, `DictionaryService` will no longer hold any actual entries storage. All entries will now live in the implementations of `IManageSuggestions`, and they're indexed.
- The workflow to load original or user dictionary from file, and rebuild it in the memory has been optimized, such that the loading time for the `DictionaryService` has been reduced to around 1 sec for EnglishUK dictionary. 
- Saving dictionary to file will now be executed in another thread such that disk i/o won't block user UIs. I've considered using async/await patterns on this type of actions, however this pattern doesn't play well with onClosing app events.
- Number of saving in-memory dictionary to file has been minimized -- we only update the on-file dictionary when absolutely necessary, and when closing the app. This will help extend your HDD's lifespan as well :)
- Root out a bug introduced in PR #330 where duplicate entries could have been saved to user dictionary file. Now we should have a slimmer foot-print of user dictionary.


JuliusSweetland:
@Chopinsky Thank you for this, but there are still a few problems. 

1. Suggestions are not working (I only tested NGram and got the below screenshot, but this is the same issue as last time I believe):

image

2. Switching to another keyboard language (form the Menu, open the Language keyboard and select French, for example) is taking a long time. OptiKey is locked up during this time.

Any chance you could take a look? Thank you. Your help is appreciated.

Chopinsky:
The lag when switching language have been fixed, but I'm not sure why predictions doesn't work from your testing, at least from testing my fork's master branch, I can see predictions from either normal entries or multi-selections:

- Multi-Selection
Multi-Selection

- Normal Entry
Normal Entry

I'm starting a new PR that will contain the fix to the language-switching lags, but I may need more information on what you saw to reproduce the issue on my machine (e.g. the counts of `entries` in `DictionaryService.cs`, the language used for testing, etc.).

PR #342, opened 2017-08-04
Add toast notifications for MaryTTS and Presage launch errors

zelmon64:
An error toast notification now appears if there is a problem with MaryTTS
or Presage at launch. The notifications include what the respective
settings have been automatically changed to.

### DIFF
--- a/src/JuliusSweetland.OptiKey/App.xaml.cs
+++ b/src/JuliusSweetland.OptiKey/App.xaml.cs
@@ -137,6 +137,7 @@ namespace JuliusSweetland.OptiKey
                 //Create services
                 var errorNotifyingServices = new List<INotifyErrors>();
                 IAudioService audioService = new AudioService();
+                bool PresageFail = false;
                 if (Settings.Default.SuggestionMethod == SuggestionMethods.Presage)
                 {
                     string resultpath = null;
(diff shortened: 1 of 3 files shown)


zelmon64:
I have not as yet been able to get OptiKey to crash when Presage is supposedly not running. After reinstalling Presage there is a checkbox for whether to start the the Presage service which I specifically unchecked but the next word suggestions were still available in OptiKey.

JuliusSweetland:
@zelmon64 I got OptiKey crashing if you rename the C:\Program Files (x86)\Presage installation directory to something else. When the Presage class is instantiated it attempts to link the installed presage libraries and throws a DllNotFoundException. I've modified the checking code to cater for this. See what you think.

Also, I've noticed that Presage is not trained on the OptiKey dictionary. Do you know how it is set up and what would happen if a different language was used in OptiKey?

zelmon64:
@JuliusSweetland This seems good to me. Presage queries the registry for it's install location when it's called so it should work if it's initial install location is set to somewhere other than `C:\Program Files (x86)\Presage` if the user wants to put it somewhere else.

The language files are in `C:\Program Files (x86)\presage\share\presage` (available are en, es and it). The setting is in the config file under:
```
    <Predictors>
        <DefaultSmoothedNgramPredictor>
            <DBFILENAME>C:\Program Files (x86)\presage\share\presage\database_en.db</DBFILENAME>
```

When I tried Spanish Optikey and Preage it seemed to work reasonably well but then I don't speak Spanish so I only used very simple phrases.

JuliusSweetland:
Great. Thanks for the extra Presage info.

PR #345, opened 2017-08-09
Dictionary Service Updates

Chopinsky:
This PR is a combine of PR #330, fixes to resolve the long loading time, and fixes to the dictionary I/O problems as mentioned in Issues #332.

This update include changes in the following area:

- As part of PR #330, DictionaryService will no longer hold any actual entries storage. All entries will now live in the implementations of IManageSuggestions, and they're indexed.
The workflow to load original or user dictionary from file, and rebuild it in the memory has been optimized, such that the loading time for the DictionaryService has been reduced to around 1 sec for EnglishUK dictionary.
- Saving dictionary to file will now be executed in another thread such that disk i/o won't block user UIs. I've considered using async/await patterns on this type of actions, however this pattern doesn't play well with onClosing app events.
- Number of saving in-memory dictionary to file has been minimized -- we only update the on-file dictionary when absolutely necessary, and when closing the app. This will help extend your HDD's lifespan as well :)
- Root out a bug introduced in PR #330 where duplicate entries could have been saved to user dictionary file. Now we should have a slimmer foot-print of user dictionary.
- Optimizing the loading time between switching languages.


JuliusSweetland:
@Chopinsky Hi Jacob. I've resolved the merge conflicts (just let BeyondCompare do its thing!)... I'm still encountering issues with English (UK) when using Basic or NGram. I get this sort of behaviour (no suggestions):
image

Oddly, when I switch to French both Basic and NGram work fine.

FYI Presage prediction works for both languages.

Any ideas?

Chopinsky:
Could you please clear everything under "C:\Users\<UserName>\AppData\Roaming\JuliusSweetland\OptiKey", and then test again?

I'm suspecting that the EnglishUK user dictionary was corrupted -- if the first time we ran OptiKey without loading the EnglishUK dictionary successfully, the corrupted or empty dictionary will be saved into the aforementioned folder and used every time from then on, meaning that our keyboard won't ever work again in that language. 

And the case that French is working is probably because the dictionary is loaded and saved to the user dictionary folder correctly, and it will keep working that way ever since. 

If this is the actual cause of what you saw, then we may need to come up with a strategy to allow resetting the user dictionary, since the user dictionary could be corrupted in a variety of ways, and that could lead to a broken dictionary service.

zelmon64:
@Chopinsky Deleting `EnglishUK.dic` does fix it. It seems that if OptiKey is closed when the suggestion method is set to Presage, the contents of `EnglishUK.dic` is deleted but the file remains. I don't think Presage used to delete the contents of `EnglishUK.dic` (it doesn't in release 2.10.0).

JuliusSweetland:
That's interesting - great find. Does it delete the dictionary with NGram
or Basic suggestions?

zelmon64:
@JuliusSweetland No, only Presage deletes it's contents. With NGram and Basic `EnglishUK.dic` stays at 689 KB but according to Notepad++ something is changed with that file when OptiKey is closed.

JuliusSweetland:
@zelmon64 @Chopinsky I agree with William's findings. Everything works with Basic and NGram. When you change to Presage it works (as the dictionary in use is baked into Presage - the OptiKey dictionary isn't actually used). At some point the EnglishUK.dic file in the roaming folder (C:\Users\Julius\AppData\Roaming\JuliusSweetland\OptiKey\Dictionaries on my computer) gets deleted.  When you change back to Basic or NGram suggestions don't work, presumably because the dictionary is empty.

Chopinsky:
Good catch on the source issue! So this is what's happening under the hood: 

PresageSuggestions.cs implemented the `IManagedSuggestions` interface, however it's not actually using the in-memory dictionary but an external dictionary for words suggestions. With that said, the in-memory dictionary is empty, and when leaving OptiKey where the in-memory dictionary is written to user file, all the user entries would be lost from then on.

Basically speaking, it doesn't make sense to load/save user dictionary when using Presage, so I'm adding safe guard around the dictionary file I/O code to prevent empty in-memory entries being saved to user dict files. 

So please try toggle between different suggestion methods, and make sure that the user dictionary file won't disappear along the way, and that the word suggestions will work fine among different methods.

zelmon64:
@Chopinsky That's fixed the empty `EnglishUK.dic` file. Great work!

One thing I noticed while switching between suggestion methods is that if I type a few words then change the suggestion method, the usage count for the words I just typed aren't added into the dictionary file. Is this a problem?

JuliusSweetland:
I think that's fine. I wouldn't expect users to switch between them in
general usage and if they do they can be considered completely separate
engines with their own word databases.

JuliusSweetland:
@zelmon64 @Chopinsky Everything is looking good now. Thank you - this is a really useful PR and will massively improve the experience for users running overly zealous virus scanners.

PR #346, opened 2017-08-11
Dynamic keyboards

kirstymcnaught:
Core functionality to allow keyboards to be defined in XML files and dynamically loaded. 

No visible changes in OptiKey - you'd have to add a key to load the dynamic keyboard menu if you wanted to access these keyboards. (and include some XML definitions).

You may wish to remove c7cb955 (Add "Back" to context menu) if you don't think it's appropriate - it's an important addition for when you're loading dynamic keyboards since it prevents you getting stuck in a keyboard without a back key, and it may also be relevant to the Communikate work. But I won't be offended if you take it out for now.


kirstymcnaught:
Just spotted some unsafe resource handling in the Xml loading - if an exception is thrown here, the file won't be disposed. 
`keyboard = (XmlKeyboard)serializer.Deserialize(readStream);`

I will replace this with a `using` statement before the PR is accepted - but please let me know if you have any other feedback first so I can fix it all in one go.

src/JuliusSweetland.OptiKey/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -212,6 +212,7 @@ private void HandleFunctionKeySelectionResult(KeyValue singleKeyValue)
                     }
                     else
                     {
+                        // TODO: Shouldn't this take you to 'default' keyboard? (may not be alpha)
```
JuliusSweetland: Yes, very good point. Could you change this?

src/JuliusSweetland.OptiKey/UI/ViewModels/MainViewModel.cs
```diff
@@ -127,6 +127,21 @@ public IKeyboard Keyboard
...
+            Log.Info("Navigating back from keyboard via contextmenu.");
+            var navigableKeyboard = Keyboard as IBackAction;
+            if (navigableKeyboard != null && navigableKeyboard.BackAction != null)
+            {
+                navigableKeyboard.BackAction();
+            }
+            else
+            {
+                // TODO: Shouldn't we always remember 'last' keyboard?
+                Log.Error("Keyboard doesn't have backaction, can't go back");
```
JuliusSweetland: Instead this should mirror the "case FunctionKeys.BackFromKeyboard:" handler and go to the default/startup keyboard.

JuliusSweetland:
@kirstymcnaught I'm going to look over the code and add a few comments now. I'll comment again here when I'm finished. One thing: the PR breaks 3 unit test (https://ci.appveyor.com/project/OptiKey/optikey-53pu1/build/2.2.5.287/job/mhah8g48c5x1wea5) - looks to be a change in the payload of one of the events.

JuliusSweetland:
@kirstymcnaught Would it be better to add a new setting to the Visuals tab in Management Console to enable the 'Dynamic' key and also to select the xml keyboard files location? You may have done this - I haven't gone through all the code or tested yet.

src/JuliusSweetland.OptiKey/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
kirstymcnaught: Sure, will do.

kirstymcnaught:
I'll take a look at the unit tests. I haven't currently added anything in Management Console, so the option is unavailable. Good idea on setting the xml keyboards location. My main priority is to get the changes to the KeyValue payloads into the master branch since this is the one thing that causes somewhat painful merges.

kirstymcnaught:
The commit "Add menu button for dynamic keyboards page" is intended to be a temporary convenience commit to facilitate your testing; we should remove it before the PR is accepted.

kirstymcnaught:
You can download some simple working XML keyboard definitions here if you want for testing: 

[download sample keyboards](https://www.dropbox.com/s/pmb9bqq4yxh0q9a/sample_keyboards.zip?dl=1)

JuliusSweetland:
@kirstymcnaught I have yet to complete my review of your commits, but I aim to over the next week. In the mean time I've merged a PR which has resulted in a few conflicts with your PR (files listed below). Could you please resolve this conflict? Your PR is next on my list...

src/JuliusSweetland.OptiKey/JuliusSweetland.OptiKey.csproj
src/JuliusSweetland.OptiKey/Properties/Resources.resx
src/JuliusSweetland.OptiKey/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs

kirstymcnaught:
My attempt at resolving merge conflicts within the github PR UI didn't work, so I'll fix offline and force push a correct merge.

JuliusSweetland:
Ok thanks. I'm slowly working my way through your commits by the way. I'm
not ignoring you!

src/JuliusSweetland.OptiKey/Services/WindowManipulationService.cs
```diff
@@ -415,6 +415,15 @@ public void Move(MoveToDirections direction, double? amountInPx)
             }
         }
 
+        public void TemporarilyResizeDockToSpecifiedHeight(double heightAsPercentScreen)
```
JuliusSweetland: Could this signature be "ResizeDockToSpecificHeight(double heightAsPercentScreen, bool persistNewSize)"

src/JuliusSweetland.OptiKey/Enums/FunctionKeys.cs
```diff
@@ -32,13 +32,16 @@ public enum FunctionKeys
...
-        Currencies2Keyboard,
-        CzechCzechRepublic,
+        Currencies2Keyboard,        
+	    CzechCzechRepublic,
         DanishDenmark,
         DecreaseOpacity,
         Delete,
         DutchBelgium,
         DutchNetherlands,
+        DynamicKeyboardMenu,
```
kirstymcnaught: Rename to DynamicKeyboard

JuliusSweetland: Rename to DynamicKeyboard

src/JuliusSweetland.OptiKey/Enums/Keyboards.cs
```diff
@@ -8,6 +8,7 @@ public enum Keyboards
         ConversationNumericAndSymbols,
         Currencies1,
         Currencies2,
+        DynamicKeyboardMenu,
```
JuliusSweetland: Rename to DynamicKeyboard

src/JuliusSweetland.OptiKey/Extensions/XElementExtensions.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Globalization;
```
JuliusSweetland: Check this file is used

src/JuliusSweetland.OptiKey/Models/DynamicKeyboardFolder.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            var applicationDataPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), ApplicationDataSubPath);
+            Directory.CreateDirectory(applicationDataPath); //Does nothing if already exists
+
+            Log.DebugFormat("GetUserKeyboardFolder: {0}", applicationDataPath);
+
+            return applicationDataPath;
+        }
+
+        // get name from XML if present
```
JuliusSweetland: Could you add a setting to the management console (under the Visuals tab) for the user to specify the directory. Default to NULL, but on startup set this to the ApplicationData...Keyboards folder.

src/JuliusSweetland.OptiKey/Models/KeyValueLink.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Text;
+using System.Xml;
+using System.Xml.Serialization;
+using JuliusSweetland.OptiKey.Enums;
+using JuliusSweetland.OptiKey.Extensions;
+
+namespace JuliusSweetland.OptiKey.Models
+{
+    [TypeConverter(typeof(KeyValueLinkConverter))]
+    public class KeyValueLink : KeyValue, IEquatable<KeyValueLink>
```
JuliusSweetland: Rename to ChangeKeyboardKeyValue.
Could you extend this so that this key type *could* link to a built in keyboard if required, e.g. to "Alpha"

src/JuliusSweetland.OptiKey/Models/KeyValueLink.cs
```diff
@@ -0,0 +1,157 @@
...
+                stringBuilder.Append(Keyboard);
+            }
+            stringBuilder.Append(" replace = ");
+            stringBuilder.Append(this.replaceKeyboard);
+            
+            return stringBuilder.ToString();
+        }
+    }
+
+    public sealed class KeyValueLinkConverter : TypeConverter
```
JuliusSweetland: Split into a new file please

src/JuliusSweetland.OptiKey/Models/KeyValuePress.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Security.Permissions;
+using System.Text;
+using System.Xml;
+using System.Xml.Serialization;
+using JuliusSweetland.OptiKey.Enums;
+using JuliusSweetland.OptiKey.Extensions;
+
+namespace JuliusSweetland.OptiKey.Models
+{
+    public class KeyValuePress : KeyValue, IEquatable<KeyValuePress>
```
JuliusSweetland: Rename to KeyPressKeyValue

src/JuliusSweetland.OptiKey/Models/XmlKeyboard.cs
```diff
@@ -0,0 +1,246 @@
...
+        public int Count
+        {
+            get
+            {
+                return BehaviourKeys.Count + TextKeys.Count + LinkKeys.Count;
+            }
+        }
+    }
+
+    public class XmlLinkKey : XmlKey
```
JuliusSweetland: Rename to XmlChangeKeyboardKey?
Also split out into separate files per class (bundle into a new folder?)

src/JuliusSweetland.OptiKey/Models/XmlKeyboard.cs
```diff
@@ -0,0 +1,246 @@
...
+        
+        [XmlElement("ReturnToThisKeyboard")]
+        public string ReturnToThisKeyboardAsString
+        {
+            get { return this.ReturnToThisKeyboard ? "True" : "False"; }
+            set { this.ReturnToThisKeyboard = XmlUtils.ConvertToBoolean(value); }            
+        }
+    }
+
+    public class XmlBehaviourKey : XmlKey
```
JuliusSweetland: Keyboard/Key/Concept renames:
Rename "BehaviourKey" to "ActionKey"
Rename "FunctionKey" to "Action"
Rename "LinkKey" to "ChangeKeyboardKey"
...and "Link" to "DestinationKeyboard"

src/JuliusSweetland.OptiKey/Services/KeyStateService.cs
```diff
@@ -68,30 +68,33 @@ public bool SimulateKeyStrokes
 
         public void ProgressKeyDownState(KeyValue keyValue)
         {
-            if (KeyValues.KeysWhichCanBePressedDown.Contains(keyValue)
-                && KeyDownStates[keyValue].Value == Enums.KeyDownStates.Up)
+            if (null != keyValue)
```
JuliusSweetland: odd. odd odd odd

src/JuliusSweetland.OptiKey/Services/KeyboardOutputService.cs
```diff
@@ -371,6 +371,13 @@ public void ProcessSingleKeyText(string capturedText)
             }
         }
 
+        public void ProcessSingleKeyPress(string key, KeyValuePress.KeyPressType type, int delayMs = 0)
+        {
+            // TODO: This is a stub for future use
+            throw new NotImplementedException();
```
JuliusSweetland: Could you pls test that this is caught and displayed as an error popup. If so could you add the text "This feature is not currently supported"

src/JuliusSweetland.OptiKey/UI/ViewModels/Keyboards/CustomKeyboard.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using JuliusSweetland.OptiKey.UI.ViewModels.Keyboards.Base;
+
+namespace JuliusSweetland.OptiKey.UI.ViewModels.Keyboards
+{
+    public class CustomKeyboard : BackActionKeyboard
```
JuliusSweetland: Is this used? Is it a dynamic keyboard?

src/JuliusSweetland.OptiKey/UI/Views/Keyboards/Common/DynamicKeyboard.xaml.cs
```diff
@@ -0,0 +1,326 @@
...
+            var allLines = info.Split(space)
+                .GroupBy(w =>  (int)((charCount += w.Length + 1) / maxLineLength))
+                .Select(g => string.Join(" ", g));
+
+            return String.Join(Environment.NewLine, allLines);
+        }
+
+        private void SetupErrorLayout(string heading, string content)
+        {
+
```
JuliusSweetland: delete white space

src/JuliusSweetland.OptiKey/UI/Views/Keyboards/Common/DynamicKeyboard.xaml.cs
```diff
@@ -0,0 +1,326 @@
...
+            {
+                Key newKey = new Key();
+                this.PlaceKeyInPosition(newKey, 3, 0, 1, 1);
+            }
+            {
+                Key newKey = new Key();
+                this.PlaceKeyInPosition(newKey, 3, 1, 1, 2);
+            }
+        }
+
```
JuliusSweetland: white space

src/JuliusSweetland.OptiKey/UI/Views/Keyboards/Common/DynamicKeyboard.xaml.cs
```diff
@@ -0,0 +1,326 @@
...
+            {
+                Key newKey = new Key();
+                this.PlaceKeyInPosition(newKey, 3, 1, 1, 2);
+            }
+        }
+        
+
+        private void SetupKeys()
+        {
+
```
JuliusSweetland: white space

src/JuliusSweetland.OptiKey/UI/Views/Keyboards/Common/DynamicKeyboard.xaml.cs
```diff
@@ -0,0 +1,326 @@
...
+            foreach (XmlTextKey key in keys.TextKeys)
+            {
+                AddTextKey(key);
+            }
+
+            foreach (XmlLinkKey key in keys.LinkKeys)
+            {
+                AddLinkKey(key);
+            }
+
```
JuliusSweetland: white space

src/JuliusSweetland.OptiKey/UI/Views/Keyboards/Common/DynamicKeyboard.xaml.cs
```diff
@@ -0,0 +1,326 @@
...
+                else
+                {
+                    resizeAction(keyboard.Height.Value);
+                }
+            }
+
+            // Get border thickness, if specified, to override
+            if (keyboard.BorderThickness.HasValue)
+            {
+                Log.InfoFormat("Setting border thickness for custom keyboard: {0}!!",
```
JuliusSweetland: been drinking!!!

src/JuliusSweetland.OptiKey/UI/Views/Keyboards/Common/DynamicKeyboardSelector.xaml.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        private void AddKeyboardKey(KeyboardInfo info, int row, int col)
+        {
+            Key lKey = new Key();
+            lKey.Value = new KeyValueLink(info.fullPath);
+            lKey.SharedSizeGroup = "KeyboardKey";
+            lKey.Text = info.keyboardName;
+            if (info.symbolString != null)
+            {
+                Geometry geom = (Geometry)this.Resources[info.symbolString];
```
JuliusSweetland: Could you add error handling that the key exists in the Resources dictionary?

src/JuliusSweetland.OptiKey/UI/Views/Keyboards/Common/Menu.xaml
```diff
@@ -48,10 +48,9 @@
                       SharedSizeGroup="KeyWithSymbolAndText"
                       Value="{x:Static models:KeyValues.NumericAndSymbols1KeyboardKey}"/>
         <controls:Key Grid.Row="0" Grid.Column="2"
-                      SymbolGeometry="{StaticResource DiacriticsIcon}"
-                      Text="{x:Static resx:Resources.DIACRITICS}"
+                      Text="Dynamic!"
```
JuliusSweetland: Could you please revert for now

JuliusSweetland: I had another thought on this...

We could retain the current Menu keyboard, but also define exactly the same keyboard in xml as a dynamic keyboard called Menu.xml (or similar). If detected in the keyboard directory (defined in the setting) then this is used to render the menu, giving the end user control of how their dynamic keyboards are used and which keys they replace to get to their other dynamic keyboards. If not detected then OptiKey can fall back to the built in Menu.xaml keyboard.

What do you think?

src/JuliusSweetland.OptiKey/UI/ViewModels/Keyboards/CustomKeyboard.cs
kirstymcnaught: not used, removed it

src/JuliusSweetland.OptiKey/Extensions/XElementExtensions.cs
kirstymcnaught: not used, removed

JuliusSweetland:
@kirstymcnaught what do you think of turning the menu keyboard into a dynamic keyboard?

kirstymcnaught:
Sounds good to me, yes. It would also help with Communikate since people could choose to sacrifice a menu item for the CK link.

kirstymcnaught:
(I'm still dealing with the last few items)

JuliusSweetland:
Cool. Agree re CommuniKate, except the icon is hard coded into the alpha
and alpha communication keyboards so we'd need another setting to show the
CK keys on your keyboards, defaulted to true. Are you happy to add that?

src/JuliusSweetland.OptiKey/Models/KeyValueLink.cs
kirstymcnaught: This wasn't used so I removed it

src/JuliusSweetland.OptiKey/Services/KeyboardOutputService.cs
kirstymcnaught: Good spot, this is not caught. Shall I just remove the not-implemented processing stubs, or also the KeyPressKeyValue classes?

kirstymcnaught:
I just tested CommuniKate and it's not working since merge commit 82981b5. 

If I set "use CK instead of alpha" then the top level CK keyboard loads okay, but all folder/link keys share the same progress graphic, and OptiKey crashes if you click any of them.

It looks like there's an issue in CK folder/link keys not being handled correctly in NotifyingConcurrentDictionary<KeyValue, double> KeySelectionProgress of IKeyStateService. I also notice there's a lot of binding errors in the console relating to CK bindings.

Would you be able to take a look?

[OptiKey.log](https://github.com/OptiKey/OptiKey/files/1290069/OptiKey.txt)
[console output](https://github.com/OptiKey/OptiKey/files/1290070/console.txt)

JuliusSweetland:
Thanks Kirsty, I'll investigate.

src/JuliusSweetland.OptiKey/Models/DynamicKeyboardFolder.cs
kirstymcnaught: I've added this. Some comments:
- defaults to empty string, then on startup it replaces with ApplicationData/etc/etc folder
- in management console, user sees current location (as a string) and selects new location from folder dialog
- If the directory is empty or doesn't exist any more, when launching the dynamic keyboards menu, extra diagnostic information is displayed to user on the empty keyboard selection grid (i.e. this is where I'm looking for keyboards, you can change it in management console/visuals)

JuliusSweetland:
@kirstymcnaught I also see the bug when I test your PR in isolation. This bug does not exist on the current master branch.

There are a couple of merge commits pulling master into your branch. Would it be possible to cherry pick your changes onto master one by one and see where it breaks? Debugging merge commits can be a bit difficult as everything gets munged together.

kirstymcnaught:
I'll take a look

JuliusSweetland:
@kirstymcnaught Thanks

src/JuliusSweetland.OptiKey/Services/KeyboardOutputService.cs
kirstymcnaught: I've removed the calling code. We now only process ChangeKeyboardKeyValues and classical KeyValues.

kirstymcnaught:
I've dealt with everything in the review except that I haven't implemented the suggested Menu.xml. I think it's a great idea, I don't entirely understand the problems with Communikate, but either way I was hoping we could discuss it/add it separately from this PR ?

kirstymcnaught:
I've added an option for the startup keyboard to be a specific XML file. The location of this is independent from the folder of dynamic keyboards, so you could have a Menu.xml that happens to live in the dynamic keyboards folder, and then choose this file as the startup keyboard, but it won't happen automatically.

JuliusSweetland:
@kirstymcnaught Hi Kirsty. Two questions:

1.It looks like this PR is in its final state - is that right?
2.Any luck identifying the issue "I just tested CommuniKate and it's not working since merge commit 82981b5."?

kirstymcnaught:
Yes, it's ready to merge. The CK fix is in 2d89f5c.

src/JuliusSweetland.OptiKey/UI/Controls/Key.cs
```diff
@@ -47,35 +47,41 @@ private void OnLoaded(object sender, RoutedEventArgs routedEventArgs)
...
+                    });
+                onUnloaded.Add(keySelectionProgressSubscription);
+            }
+            var progress = (Value == null) ? 0 : keyStateService.KeySelectionProgress[Value].Value;
             SelectionProgress = progress;
             SelectionInProgress = progress > 0d;
 
             //Calculate IsEnabled
-            Action calculateIsEnabled = () => IsEnabled = keyStateService.KeyEnabledStates[Value];
+            Action calculateIsEnabled = () => IsEnabled = Value != null && keyStateService.KeyEnabledStates[Value];
```
zelmon64: @kirstymcnaught this alteration is causing the blank CK keys to be black instead of coloured. Changing this line to 
```
Action calculateIsEnabled = () => IsEnabled = (Value != null || HasImage) && keyStateService.KeyEnabledStates[Value];
```
fixes this and I believe it should not effect anything else.

zelmon64: @kirstymcnaught I made a PR (https://github.com/kirstymcnaught/OptiKeyMinecraft/pull/1) with this addition along with a modification to the debug logging for the CK Back key. I'm not completely sure if that is the correct procedure so if it's not please ignore it.

kirstymcnaught:
@zelmon64 , sorry about turning your blank buttons black! How do you feel about the fix in 1cbd854 instead of your proposal? It seems best to keep the concept of "enabled" key state away from the styling of keys, as well as not encoding special enabled-ness to null keyvalues (I don't think your fix causes any issues now, but I can imagine a time where a null keyvalue exists outside of CK and should be treated as a disabled key, not an enabled one). Does that make sense?

src/JuliusSweetland.OptiKey/UI/Controls/Key.cs
kirstymcnaught: Thanks! I've answered below.

JuliusSweetland:
@kirstymcnaught Null key values exist now where a key is used to fill in unused space on a keyboard, e.g. at the end of a row. The value is not set in XAML so the key should behave as if it is disabled.

zelmon64:
@JuliusSweetland does that mean it should be changed back to the original 
`Action calculateIsEnabled = () => IsEnabled = keyStateService.KeyEnabledStates[Value];`
code or that it's safe to use my 
```
Action calculateIsEnabled = () => IsEnabled = (Value != null || HasImage) && keyStateService.KeyEnabledStates[Value];
```
 modification?

@kirstymcnaught no worries. 1cbd854 unfortunately doesn't preserve blacking out the CK keys when the Sleep key is selected but I'm happy with fixing it in the themes rather than Key.cs (I only did it in Key.cs because it was the smallest change).

JuliusSweetland:
I'm not in front of my computer. Does
keyStateService.KeyEnabledStates[Value]; when Value == null return false or
throw an exception?

If value is null for keys with images then how are they ever renamed as the
lookup keyStateService.KeyEnabledStates[Value]; should always fail right?

zelmon64:
@JuliusSweetland `keyStateService.KeyEnabledStates[Value]` when `Value == null` doesn't seem to throw an exception so I assume it returns false (I can't quite tell by looking at the code sorry).

The value isn't null for all CK keys, only the blank ones that have no function (they just have a background colour).

zelmon64:
@JuliusSweetland changing it back to 
```Action calculateIsEnabled = () => IsEnabled = keyStateService.KeyEnabledStates[Value];```
does fix the CK blank keys but I can't tell if it effects the dynamic keyboards.

JuliusSweetland:
@kirstymcnaught can you comment on whether changing it back works with the
dynamic keyboards?

kirstymcnaught:
Hi Julius, 

The method keyStateService.KeyEnabledStates[Value] usually returns **true** for Value=null, which I think is itself a bug. The only exception is when OptiKey is sleeping in which case KeyEnabledStates[] returns false for all key values, including null. Note that KeyEnabledStates is a class with a custom []operator containing lots of state logic, not any kind of standard collection. 

Communikate is relying implicitly on the fact that KeyEnabledStates[null] = true when not sleeping and false when sleeping, so that an empty CK key is usually styled as an enabled key rather than a disabled one, even though it has no payload. Only when sleeping does it get greyed out as a disabled key. I think this is confusing behaviour and liable to cause problems down the road. 

I thought that we could fix this by binding the "greying out" styling to (!IsEnabled and !HasImage) as in 1cbd854, but as William pointed out, this breaks the special case for sleeping. Really the logic that CK wants is "grey out if (disabled but isn't a CK key) or (OptiKey is sleeping)". I'm not sure how best to make this logic explicit. 

What do you think? If you prefer we could revert to 
`Action calculateIsEnabled
= () => IsEnabled = keyStateService.KeyEnabledStates[Value];` to preserve the behaviour for now (this isn't a problem with dynamic keyboards) and then separately log a bug for the unexpected behaviour of KeyEnabledStates and nulls ? 

PS - 
@zelmon64 , I'm a little confused by the Key property "HasImage". It doesn't seem to relate to whether or not the key has an image or not (since empty keys don't), but is set to true for all communikate keys. Is this correct? Is it worth renaming to "IsCommunikateKey" (or similar) if that is indeed what it represents? (currently it does seem to be used as a proxy for being a CK key).

zelmon64:
@kirstymcnaught sorry for the confusion. I initially labelled it as `HasImage` before I realised some keys don't actually have an image and that the blank ones should be coloured. Technical it's used as "key has transparent background and black (or white depending on the theme) selection animation". I suppose either `IsCommunikateKey` or `IsTransparent` would be a better option than `HasImage`. 

What I liked about using `Value = null` for the blank CK keys was that they don't have a selection animation but they are enabled (coloured). 

On the other hand, an alternative approach would be to instead have the option to specify the key's background colour. This would then be used to override the theme's default background colour. This approach may also benefit your dynamic keyboards unless they already have a method for changing the colours (apologies if I missed that feature).

Getting back to the dynamic keyboards, are the [example xml files you posted here](https://github.com/OptiKey/OptiKey/pull/346#issuecomment-322534229) still supposed to work correctly? When I tried them at the end of August they worked fine. Now if I set one as the start-up keyboard or reinstate the dynamic key on the menu keyboard, they either say not found or don't display all the keys. Has the syntax been changed slightly or am I missing something else?

JuliusSweetland:
@zelmon64 @kirstymcnaught I'm still playing catch up here...

1.Are we only discussing how empty CK keys are styled?
2.If we reinstate "Action calculateIsEnabled = () => IsEnabled = keyStateService.KeyEnabledStates[Value];" and fix the logic in KeyEnabledStates[Value] to return false if the Value is null, what impact does this have on:
a)Empty (spacer) keys, e.g. the spacer keys on the left of the "Q" on the english keyboard layouts?
b)Empty CK keys?
c)All other keys?

kirstymcnaught:
@JuliusSweetland 

1. We are almost exclusively talking about how CK keys are styled, but CK key styling hooks into usual OptiKey enabled/disabled key styling to some extent. 

2. In the case you describe:
a - _Empty CK keys_ get turned black always, because their null key value makes them "disabled" which greys them out. The intention for CK was for blank keys to be coloured, not black (except when sleeping), but the "greying out" of disabled keys covers up the coloured background. 
b - _Empty (spacer) keys_ in normal OptiKey behave as they should (but they also behave okay regardless of how we handle nulls; they always appear black regardless of whether they are "greyed out" or not since they have no text/symbol)
c - _Empty dynamic keys_ behave as they should - they get disabled and greyed out.
d - Keys with payloads all behave as expected in either case.

(a) is considered the original bug that my PR introduces to the CK behaviours.
(b) and (d) are irrelevant I think, OptiKey blank keys and other keys are fine whatever.
(c) is my desired behaviour for dynamic keys. If we allow the previous behaviour of "null keys are enabled unless sleeping" (which underpins the CK behaviour) then I lose this. Dynamic keys with no payload would look like enabled keys, but they don't behave like them (no progress indicator) so it's acceptable for now if necessary. I could also fix it by making sure all dynamic keys have *some* KeyValue, but when it's not defined, it will be an empty one (e.g. no text or functionkey) which we can check for explicitly.

kirstymcnaught:
@zelmon64 You are right, the spec has changed since I uploaded them (we renamed a few things as part of the PR code review). Here's a link to updated versions. Note that a few buttons intentionally have missing data - currently since null values are enabled, they appear enabled but aren't clickable. Previously they looked greyed out. 

[new sample keyboards](https://www.dropbox.com/s/qvgi8d3d7oq8a8t/sample_keyboard_2.zip?dl=1)

JuliusSweetland:
@kirstymcnaught Thanks. I think we should reinstate "Action calculateIsEnabled = () => IsEnabled = keyStateService.KeyEnabledStates[Value];" and fix the logic in KeyEnabledStates[Value] to return false if the Value is null.

@zelmon64 If we do this we need another mechanism to enable CK keys without a key value to be coloured. Can they expose a new property (I guess this would replace HasImage?) such as IsCommuniKateKey, which could be styled appropriately in the themes?

Does that suit everyone and solve all the outstanding issues?

kirstymcnaught:
I don't have any outstanding issues. 

Do you want me to squash the last few commits and force push them here, or do you prefer to preserve the historical to and fro?

zelmon64:
@kirstymcnaught thanks for the new samples.

@JuliusSweetland that should work out. However, would exposing the key's background colour (for both when enabled and disabled) work better instead? What is the view on customising the theme? Is there an option for alternative colours with the dynamic keyboards?

JuliusSweetland:
@kirstymcnaught I'm happy to preserve the to and fro. The history on this project is hardly a work of art as I commit and roll back lots of things anyway.

@zelmon64 It looks like CKKey.xaml sets the grid background colour like this:
`<Grid Background="{Binding Path=CKBaCo, RelativeSource={RelativeSource AncestorType={x:Type controls:CKKey}}}">`
I would be happy for you to rename HasImage to something like IsCommuniKate key and handle that in the themes to prevent the disabled brushes from being applied (as you do now with HasImage).

Also is it necessary to set SymbolGeometry="" on the keys, or could you just leave that out? I'm only interested on this question.

kirstymcnaught:
@zelmon64 Currently dynamic keyboards get almost entirely themed by OptiKey. There are a few small modifications that you can define in the XML - but only keyboard-level things like whether there's an extra border around the whole keyboard, and whether there's extra padding around all symbols. There is no support for images (you have to define an existing symbol from OptiKey) and you have no control over colours of anything. I don't have particular plans to extend this, but it's reasonable that somebody might want to extend in this way.

zelmon64:
@JuliusSweetland yes that is how the colour is currently set but I was wondering if it would be better to instead have an optional background colour property exposed which would override the theme if set. This property could then also be used by the dynamic keyboards. 

`SymbolGeometry=""` was used initially in CKKey.xaml to hide the text. CKKey.xaml was meant to be a template linked to in CK20Page.xaml but I couldn't get that to work. `SymbolGeometry=""` isn't used in the final version of CK20Page.

@kirstymcnaught thanks for the information. Am I right in thinking that if the properties `EnabledBackgroundColour` and `DisabledBackgroundColour` were added to Key.cs then it would be relatively straightforward to hook the dynamic keyboards into it?

JuliusSweetland:
@zelmon64 I'm happy for there to be string BackgroundColourOverride and DisabledBackgroundColourOverride dps, sure.

kirstymcnaught:
@zelmon64 Yes, any accessible members of Key.cs should be able to be set/overridden when the keyboard is parsed, in DynamicKeyboard.xaml.cs. It's also fairly trivial to add the Xml definitions for a new property in the XmlKeyboardModels files.

zelmon64:
@JuliusSweetland thank you. After this PR is merged I'll try that then. (I'm assuming "dps" was a typo unless it's an acronym I don't know about)

@kirstymcnaught excellent. I'll have a go at adding the colour overrides to the Dynamic Keyboards when I make them then if that's okay.

JuliusSweetland:
@zelmon64 DependencyProperties

JuliusSweetland:
@kirstymcnaught Merged - thank you!
@zelmon64 This PR has been merged if you have a second to correct the CommuniKate key backgrounds on a new PR? Thanks!

zelmon64:
@JuliusSweetland great thanks! Thanks also for creating the new issue. I'll work on that tomorrow.

JuliusSweetland:
Great! Thank you.

Issue #353, opened 2017-08-22
MaryTTS no voices choice.


Sheeva007:
I have installed MaryTTS as preconized and checked "use MaryTTS" in sound settings window but, when i try to select voice to use this stay empty.

So, i have started marytts-server.bat from the console, everythink start correctly. Then, i try to run component-installer ,in my case french pierre. And again everything is available and operational. I can ear this very realistic french voice. Under Optikey... Nothing append. No error. No voice.

Thx.

zelmon64:
If it's configured correctly, marytts-server.bat should be started and closed automatically with OptKey. When you checked `use MaryTTS` you should then have clicked on the button below and located the marytts-server.bat file. Then after closing the management console OptiKey should restart and the available MaryTTS voices should be listed in the management console.

Sheeva007:
@zelmon64 Same behavior here. No MaryTTS server console started While optikey run. I give up for now. i will continue with **eSpeak** crapy voices.

JuliusSweetland:
@Sheeva007 Can you try manually running marytts-server.bat before OptiKey starts (and have MaryTTS enabled in the Management Console) - let us know if that works.

Sheeva007:
@JuliusSweetland It's working manually. With english-female voice (default voice i présume)

JuliusSweetland:
Thank you. If you close MaryTTS and then let OptiKey start it up, by
starting OptiKey, do you see a command prompt open with MaryTTS? If not
could you upload or email your logs?

Sheeva007:
@JuliusSweetland I will try to be clear... If i start MaryTTS server manually then optikey it's working like i said before. But, if i go to management window in sound section i can select other voices included with MaryTTS package. So i'm feeling like a super hero :-) t'ill i press ok ...optikey very kindly crash.
I confirm. The MaryTTS server is not started by Optikey. I send you log file later.(my wife is waiting for me)

JuliusSweetland:
Ok thank you. Your log file will be very helpful. When OptiKey crashes
(actually when you next start it) it emails me the crash log so I can see
what happened - I believe I can see your crash. It doesn't look like it has
anything to do with MaryTTS. Can you try changing any other (non-MaryTTS)
setting in the Management Console and clicking OK - I suspect it will crash
also.

JuliusSweetland:
[...]

PR #355, opened 2017-08-26
Added more startup checks to MaryTTS

zelmon64:
* After trying to start the MaryTTS server, it is now checked that it was
started and hasn't been closed.

* If the MaryTTS server is closed before OptiKey, OptiKey nolonger tries
to close it.

* If MaryTTS is not available when the speak key is pressed, an attempt is
made to restart it. If this fails then MaryTTS is disabled.

@JuliusSweetland I only made these additions for #353 but they might be useful for future debugging. What do you think?

### DIFF
--- a/src/JuliusSweetland.OptiKey/App.xaml.cs
+++ b/src/JuliusSweetland.OptiKey/App.xaml.cs
@@ -963,7 +963,21 @@ namespace JuliusSweetland.OptiKey
                     }
                 };
 
-                if (Settings.Default.MaryTTSLocation.EndsWith(ExpectedMaryTTSLocationSuffix))
+                if (!File.Exists(Settings.Default.MaryTTSLocation))
+                {
+                    Log.InfoFormat("Failed to started MaryTTS (setting MaryTTSLocation does represent an existing file). " +
+                        "Disabling MaryTTS and using System Voice '{0}' instead.", Settings.Default.SpeechVoice);
+                    Settings.Default.MaryTTSEnabled = false;
+
+                    mainViewModel.RaiseToastNotification(OptiKey.Properties.Resources.MARYTTS_UNAVAILABLE,
+                        string.Format(OptiKey.Properties.Resources.USING_DEFAULT_VOICE, Settings.Default.SpeechVoice),
+                        NotificationTypes.Error, () =>
+                        {
+                            inputService.RequestResume();
+                            taskCompletionSource.SetResult(false);
+                        });
+                }
+                else if (Settings.Default.MaryTTSLocation.EndsWith(ExpectedMaryTTSLocationSuffix))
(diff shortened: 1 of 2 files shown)


JuliusSweetland:
@zelmon64 Couple of thoughts:

1. In CloseMaryTTSOnApplicationExit() why check again if the MaryTTS setting is enabled before closing the proc? This should always be safe to attempt, no?

2. In the Sound class there is an attempt to restart MaryTTS, but if this occurs then the process will not be closed when OptiKey exits. What do you think about refactoring out an extension method on the Process class which handles this for both scenarios (and any other processes that we want to spawn in future). Something like this;

`        private static void CloseOnApplicationExit(this Process proc, string description)
        {
            Current.Exit += (o, args) =>
            {
                if (proc.HasExited)
                {
                    Log.InfoFormat("{0} has already been closed.", description);
                }
                else
                {
                    try
                    {
                        proc.CloseMainWindow();
                        Log.InfoFormat("{0} has been closed.", description);
                    }
                    catch (Exception ex)
                    {
                        Log.Error(string.Format("Error closing {0} on OptiKey shutdown", description), ex);
                    }
                }
            };
        }`

JuliusSweetland:
Actually I'll make this change as I want to test it quickly...

JuliusSweetland:
I've added this change, and merged in @Chopinsky 's PR to update usage count when using Presage. Your fix when unpacking the CommuniKate boards has also been merged in. If you have a second could you please test the new v2.11.0 pre-release (3) here: https://github.com/OptiKey/OptiKey/releases

zelmon64:
@JuliusSweetland v2.11.0 pre-release (3) seems to work very well. Thanks for finding a way to close the MaryTTS server when it has been restarted.

One thing I noticed though was that the dictionary list in the management console has repeated entries when set to Basic or NGram. The number each entry is repeated seems random. The EnglishUK.dic file is fine however.

JuliusSweetland:
@Chopinsky do you know why the entries may be duplicated in the management
console? Thanks

Chopinsky:
I think this may have something to do with normalizing entries. For example, Mike's and Mikes will be normalized into the same index and saved in the dictionary, and I think historically this is made by design to improve the search performance. 

I will take a look at how they're presented in the console, and that may very well be showing the entry index value instead of the entry object value.

JuliusSweetland:
That sounds likely. Thanks Jacob.

Chopinsky:
Looks like we're pulling all the normalized entries for NGram as well, that polluted the dictionary tab. I'm creating new PR to fix the issue. Base seems unaffected by this bug though.

zelmon64:
@Chopinsky I just double checked and you are right about Basic being unaffected by this. My mistake.

PR #357, opened 2017-08-27
Stop displaying NGram's internal entries in config menu's dict tab

Chopinsky:
Currently if using NGram, config menu's dictionary tab will pull in helper entries as well. They're created for internal use only, and shouldn't being pulled out for user-facing UIs.

Files modified:
- _DictionaryService.cs::GetAllEntries()_: now we will filter all the helper entries when returning the dictionary entries. Currently this function is called by user facing UI view's to display the content of the dictionary entries, hence we should filter those internal-only entries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 *.suo
 *.user
 *.sln.docstates
+.vs
 
 # Release deployment, including certificates, etc
 deployment/
--- a/src/JuliusSweetland.OptiKey.UnitTests/packages.config
+++ b/src/JuliusSweetland.OptiKey.UnitTests/packages.config
@@ -3,7 +3,7 @@
   <package id="Castle.Core" version="4.1.1" targetFramework="net46" />
   <package id="CommonServiceLocator" version="1.3" targetFramework="net46" />
   <package id="Moq" version="4.7.99" targetFramework="net46" />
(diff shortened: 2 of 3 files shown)


JuliusSweetland:
@Chopinsky Thanks very much!

JuliusSweetland:
@Chopinsky @zelmon64  I've merged this in, but noticed 2 issues:

1.With NGram suggestions the duplicates in the Management Console Dictionary tab are still there, e.g. aardvark appears twice.

2.With Basic suggestions the duplicates in the Dictionary tab are gone, but some suggestions appear not to work, e.g. aardvark is not suggested if I type "aa"

Chopinsky:
I will investigate this issue.

Chopinsky:
The Basic suggestions do contain a bug, where words with same characters in neighboring positions may not be found correctly due to normalization, I'm fixing it with my next PR.

Sorry for all these ad hoc updates, the dictionary service are much more complex than I initially thought (and retrospectively speaking, I may not pick up this piece for starter at the first place -_-|||). 

Though I didn't notice the first issue you mentioned, here're a screen shot of my dictionary tab when using NGram suggestions:
image
 
Could you please provide more info on the steps you've taken prior to seeing the duplicates?

PR #358, opened 2017-08-28
Fixing word suggestions when using Basic suggestions

Chopinsky:
Currently when using Basic suggestions, some of the words with same characters in neighboring positions (e.g. _aardvark_) will not show in the suggestions. This issue is caused by inappropriately indexing the words. 

Files modified:
- _BasicAutoComplete.cs_: now we are going to index the word using _normalize_ method (previously we were using _NormaliseAndRemoveRepeatingCharactersAndHandlePhrases_ method to index the entry).


JuliusSweetland:
@Chopinsky @zelmon64 This looks good now - thank you. Pre-release here if anyone wants to try it out: https://github.com/OptiKey/OptiKey/releases/tag/v2.11.0_pre-release_4

PR #361, opened 2017-09-11
Remove unused files

kirstymcnaught:
These files don't seem to be used but duplicate a lot of the functionality in CKPage20.xaml(.cs) which is very confusing for debugging.


zelmon64:
@kirstymcnaught sorry about this. I think `CKKey.xaml` and `CKKey.xaml.cs` may also be currently unused. The idea was that `CKKey` would be used twenty times in `CKPage20` and twelve times in a 4*3 layout (which would be `CKPage12`). I couldn't get this to work so I used the long method in `CKPage` for `CKPage20` and forgot to delete `CKPage.xaml(.cs)` when it was tidied up (there were many more remnants of failed attempts that were deleted).

I'd prefer `CKKey` to remain for when I have another shot at that method but if it's in the way I'm okay with it being removed. `CKPage.xaml(.cs)` can certainly be removed either way.

Issue #365, opened 2017-09-13
Communikate folder location changes can't be cancelled


kmcnaught:
- Open Management Console -> Words -> Enable CK -> Location of CK top .OBF file, change to a new location, confirm file dialog.
- Cancel Management Console
- The CK location has been changed (it shouldn't have been because the operation was cancelled)
- Launch Management Console again
- Without changing anything, click "Ok"
- You get the message "Restart required"

There's also some confusion in the labels vs. behaviour. It says "Location of CommuniKate top page .obf file", but the default is an ".obz" file and the file selector only allows other .obz files, which represent an entire pageset and not a top page. Is it supposed to work with both? (zipped obz and top page obf).

Chopinsky:
I will take care of this

JuliusSweetland:
@Chopinsky Thanks

kmcnaught:
@Chopinsky @JuliusSweetland I just made a comment about this on the closed PR rather than the open issue - see https://github.com/OptiKey/OptiKey/pull/371

JuliusSweetland:
@kmcnaught You make a good point. 

@Chopinsky I missed the call to "Settings.Default.Reload();" when I reviewed this. As Kirsty suggests I think a solution would be to remove the call to "Settings.Default.Reload();" when cancelling as the changes should NOT have been applied at that point. This requires a bit of clean up which Kirsty also describes:

"Communikate breaks this by making immediate changes to the settings rather than storing locally in the viewmodel until ApplyChanges() is called.

Is it possible to add a test that checks each view upholds this contract? Then we could just fix Communikate settings and any future breaking changes would be flagged by the test."

What do you think?

Chopinsky:
As Kirsty mentioned in the comment to the PR, the reason to call `Reload()` is because that the code behind the "Location of CK top .OBF file" setting _will_ in fact save the file location to the persistent data once the change's been made, so to revert that change, we need to reload the settings with persistent data. 

So I can think of 2 fixing strategy:

1. As Kirsty suggested, we use `ApplyChanges` pattern in the "Communicate Settings" as well, and remove the call to `Reload`.

2. When we click "Ok", we find a way to save the changes to the persistent data, in that case the next time we call `Reload`, we will revert to the most recent updated settings. 

I'm find with either way, we could do both as well. Once we can reach a conclusion, I can make those updates tonight. Creating a test on this could be a bit difficult, at least I need to think more about how could make it work.

[...]

PR #371, opened 2017-09-19
Cancel Unsaved Settings Shall Restore Original Settings

Chopinsky:
As noted in issue #365, if canceling unsaved changes, not all settings will be restored to their previous values. This could cause problems if users accidentally changed some settings (e.g. checked a checkbox without noticing that) but don't have a way to back out what they have changed.

Now, new settings view will always check if a change has been made, and if so, when user click "Cancel" button to close the modal view, they will be asked if they want to discard the changes -- if they choose to proceed, all changes will be restored to the previous values; otherwise, the settings modal view will stay.

Be noted that `BasicAutoComplete.cs` isn't actually changed at all, but VS 2017 behaved weird again by changing all the spacing again, which made it look like being altered, but this PR does not change any of its code (but please double check and feel free to revert any unintended code changes).

### DIFF
--- a/src/JuliusSweetland.OptiKey/UI/ViewModels/ManagementViewModel.cs
+++ b/src/JuliusSweetland.OptiKey/UI/ViewModels/ManagementViewModel.cs
@@ -114,27 +114,20 @@ namespace JuliusSweetland.OptiKey.UI.ViewModels
 
         private void Cancel(Window window)
         {
-            if (ChangesRequireRestart)
-            {
-                //Warn if wants to discard all the changes
-                ConfirmationRequest.Raise(
-                    new Confirmation
+            //Warn if wants to discard all the changes
+            ConfirmationRequest.Raise(
+                new Confirmation
+                {
+                    Title = Resources.SETTINGS_CHANGED,
+                    Content = Resources.DISCARD_CHANGED_SETTINGS
+                }, confirmation =>
+                {
+                    if (confirmation.Confirmed)
                     {
-                        Title = Resources.SETTINGS_CHANGED,
-                        Content = Resources.DISCARD_CHANGED_SETTINGS
-                    }, confirmation =>
-                    {
-                        if (confirmation.Confirmed)
-                        {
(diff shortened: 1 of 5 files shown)


src/JuliusSweetland.OptiKey/UI/ViewModels/ManagementViewModel.cs
```diff
@@ -112,9 +112,29 @@ private void Ok(Window window)
             }
         }
 
-        private static void Cancel(Window window)
+        private void Cancel(Window window)
         {
-            window.Close();
+            if (ChangesRequireRestart)
```
JuliusSweetland: ChangesRequireRestart does not detect if ANY changes have been made, only if changes which require OptiKey to be restarted have been made. Should this not check to see if any changes exist?

JuliusSweetland:
@Chopinsky Just leaving another comment in case you weren't alerted to my comment feedback above.

src/JuliusSweetland.OptiKey/UI/ViewModels/ManagementViewModel.cs
Chopinsky: My mistake, I thought that property represents if any changes have been made in the settings modal view. In this case, I think we would probably just warn users regardless to keep the scope manageable.

kmcnaught:
I just took a look at this proposed fix, and we now get the problem that if you change some settings, save them, and then later you cancel a change, you will lose all the earlier changes you made (and confirmed) as the settings file isn't saved until the app closes.

Previously the strategy was for each tab in the management console to apply it's changes only on ApplyChanges() when the management console was "OK-ed", This seems to work well so long as the sub-views respect the protocol - currently Communikate breaks this by making immediate changes to the settings rather than storing locally in the viewmodel until ApplyChanges() is called.

Is it possible to add a test that checks each view upholds this contract? Then we could just fix Communikate settings and any future breaking changes would be flagged by the test.

PR #373, opened 2017-09-21
Save file selections in the Settings to view model properties instead

Chopinsky:
As discussed in issue #365, `SoundsView` and `WordsView` in the settings modal view have broken the `ApplyChanges()` pattern by saving file selections to the Settings object directly. In specific, when updating "Mary TTS Location" setting in `SoundsView`, or "CommuniKate Pageset Location" in `WordsView`, the changes will be applied to Settings directly, such that even if users choose to click the "Cancel" button to discard all the changes, these two file locations would still be updated -- they're bypassing the `ApplyChanges()` call, which are triggered by the "Ok" button only. 

This actually lead to the original issue as raised by issue #365. The solution would be enforcing the `ApplyChanges()` patter: instead of saving file selections directly to Settings, we will apply the settings to the model properties. These properties will be used to update Settings when `ApplyChanges()` is invoked.


zelmon64:
@Chopinsky just wanted to say thank you very much for fixing this! It was very kind of you to handle this since it was my fault.

JuliusSweetland:
It's no one's fault. The code base is big and complex. But thanks for
making the change.

Issue #372, opened 2017-09-19
CommuniKate keys with no Value no longer have a background colour


JuliusSweetland:
@zelmon64 As discussed - this is an issue for the outstanding CK key issue.

JuliusSweetland:
@zelmon64 Could you explain ShowOverScratchpadOveride more (also "Overide" is slightly misspelled). Shouldn't the sleep key behave like any other normal (not CK) key?

When you say the "sleep colour" this isn't a separate colour or behaviour is it? This should just be the disabled state of each key. Could I suggest making each key have a grey background (like the foreground colour of normal disabled keys) and a foreground colour of black (or a darker grey). The blank CK keys in your screenshots might confuse some users into thinking the keyboard is broken and not disabled.

Lastly regarding the dps:
1.Could you please remove the underscores? I don't tend to use them in property names.
2.Could you remove the Has_ColourOverride and the brush _ColourOveride properties and instead use value converters in the theme? You could use the same one to convert a string to a Brush class. As it's a class it is nullable so you could test the converted brush to see if it's null (which gives the same functionality as Has_ColourOverride) or use the brush (which gives the same functionality as _ColourOveride).

Happy to discuss.

zelmon64:
@JuliusSweetland 

> @zelmon64 Could you explain ShowOverScratchpadOveride more (also "Overide" is slightly misspelled). Shouldn't the sleep key behave like any other normal (not CK) key?

This is for the **Speak** key, not the **sleep** key. In other CommuniKate / OBF communication boards the text is spoken by clicking on the scratchpad-like area. To replicate this with OptiKey I placed the **speak** key over the scratchpad. Normally this would completely hide the scratchpad. To make the scratchpad visible I removed the background from the speak key (made it transparent) and also only show the icon when the key is either current, selection is in progress or is is down (speaking). 

> When you say the "sleep colour" this isn't a separate colour or behaviour is it? This should just be the disabled state of each key. Could I suggest making each key have a grey background (like the foreground colour of normal disabled keys) and a foreground colour of black (or a darker grey). The blank CK keys in your screenshots might confuse some users into thinking the keyboard is broken and not disabled.

The original "sleep colour" was black for all CK keys (like the alpha keys). With how `KeyEnabledStates` was altered I could not find a way to distinguish between blank (disabled) keys and sleeping keys. Instead of adding some convoluted way to distinguish between the two I decided to make both the sleeping keys and the blank keys a solid colour. 

[...]

PR #376, opened 2017-09-21
Fix for CommuniKate formatting

zelmon64:
As discussed in #372, the CommuniKate formatting of blank and sleeping keys needed to be fixed.

### DIFF
--- a/src/JuliusSweetland.OptiKey/Resources/Themes/Android_Dark.xaml
+++ b/src/JuliusSweetland.OptiKey/Resources/Themes/Android_Dark.xaml
@@ -39,13 +39,6 @@
             <SolidColorBrush x:Key="KeyDisabledBorderBrush" Color="{StaticResource Grey}" />
             <SolidColorBrush x:Key="KeyDisabledBackgroundBrush" Color="{StaticResource Black}" />
 
-            <SolidColorBrush x:Key="KeyHasImageForegroundBrush" Color="{StaticResource Transparent}" />
-            <SolidColorBrush x:Key="KeyHasImageBorderBrush" Color="{StaticResource Transparent}" />
-            <SolidColorBrush x:Key="KeyHasImageBackgroundBrush" Color="{StaticResource Transparent}" />
-            <SolidColorBrush x:Key="KeyHasImageHoverForegroundBrush" Color="{StaticResource Black}" />
-            <SolidColorBrush x:Key="KeyHasImageSelectionProgressBrush" Color="{StaticResource Black}" />
-            <StaticResource x:Key="KeyHasImageSelectionBackground" ResourceKey="Transparent"/>
-
             <StaticResource x:Key="ScratchpadCaretBackgroundBrush" ResourceKey="Blue"/>
             <SolidColorBrush x:Key="ScratchpadForegroundBrush" Color="{StaticResource Blue}" />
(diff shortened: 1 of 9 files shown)


src/JuliusSweetland.OptiKey/Resources/Themes/Android_Dark.xaml
```diff
@@ -234,36 +238,16 @@
                         </MultiDataTrigger.Setters>
                     </MultiDataTrigger>
 
-                    <!--Show symbol if there is an image and the key is current-->
+                    <!-- ...or if Show Over Scratchpad-->
```
JuliusSweetland: Minor change as I'll forget what this means. Could we rename this property to "OnlyVisibleWhenInUse" or something similar?

src/JuliusSweetland.OptiKey/UI/Controls/CKKey.xaml
```diff
@@ -49,7 +49,7 @@
                     <Setter Property="Content">
                         <Setter.Value>
                             <controls:Key SharedSizeGroup="KeyWithText"
-                                          HasImage="True" SymbolGeometry="" 
+                                          SymbolGeometry=""
```
JuliusSweetland: Does the SymbolGeometry have to be set at all?

src/JuliusSweetland.OptiKey/UI/ValueConverters/OverrideHasValue.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Windows;
+using System.Windows.Data;
+
+namespace JuliusSweetland.OptiKey.UI.ValueConverters
+{
+    public class OverrideHasValue : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            return value != null && !value.Equals("") && value.ToString() != "";
```
JuliusSweetland: Could this be replaced with !string.IsNullOrEmpty?

src/JuliusSweetland.OptiKey/UI/Controls/CKKey.xaml
zelmon64: Probably not. CKKey isn't actually used. I've left it in as a placeholder because I would like to use it as a template in CK20Page like how Key is used but my attempts at doing it that way haven't worked so far. It had to be edited because `HasImage` was removed.

If it's too distracting to keep the file I'll delete it.

JuliusSweetland:
No that's fine

src/JuliusSweetland.OptiKey/UI/ValueConverters/OverrideHasValue.cs
zelmon64: Yes, that seems to work. I've replaced it with
`return value != null && !string.IsNullOrEmpty(value.ToString());`
is that correct?

src/JuliusSweetland.OptiKey/Resources/Themes/Android_Dark.xaml
zelmon64: I've changed it to `OnlyVisibleWhenInUse`. I somehow missed this comment when I did the others earlier.

PR #387, opened 2017-10-02
Disable combining keys when MultiKeys mode is on

Chopinsky:
This PR will disable accent combining keys if MultiKeys mode is turned on (either in single word mode, or locked down mode) in many keyboards we support. Due to the natures of how these accent combining keys work, they couldn't really fit into the MultiKey selection workflow, hence we will disable those keys to avoid confusions.

However, there are quite a number of keyboards that contain _on-board_ accent keys, for example, the "Ç" and "é" in French keyboard. In some of the keyboards, these on-board keys are enabled for multi-key selection, while some are not. To be consistent, this PR will treat all on-board accent keys as multi-key enabled (and also adding them to the multi-key enabled key list). 

In addition, we're going to disable the left "Control", "Alt", and "Win" keys. These keys are in grey area since they aren't really useful in multi-key selection mode, so they should be turned off as well. But we can revert this change if necessary.

### DIFF
--- a/src/JuliusSweetland.OptiKey/Models/CapturingStateManager.cs
+++ b/src/JuliusSweetland.OptiKey/Models/CapturingStateManager.cs
@@ -17,6 +17,8 @@ namespace JuliusSweetland.OptiKey.Models
         }
 
         private bool capturingMultiKeySelection;
+        private bool multiKeyDownOrLocked;
+
         public bool CapturingMultiKeySelection
         {
             get { return capturingMultiKeySelection; }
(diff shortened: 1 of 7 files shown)


JuliusSweetland:
Thanks! Could we keep the ctrl, win, etc keys enabled when multi key selection is down? The logic governing held down keys as you charge keyboards could be tricky if they are disabled/released automatically.

Chopinsky:
Sure thing, I've updated the list of keys to be disabled, and now we're excluding the left control/alt/win keys.

JuliusSweetland:
@Chopinsky Nice change. I've refactored it slightly - there's actually no need to bundle the key state changes into the CapturingStateManager - you can subscribe directly to the KeyDownStates changes in the KeyEnabledStates class. Thanks!

Issue #370, opened 2017-09-18
Catch exceptions from copyfromscreen and display error instead


JuliusSweetland:
There is an unhandled exception that can be thrown from the CopyFromScreen method (details below):

This occurs in the MagnifyPopup control and I imagine handling this by wrapping the call to DisplayScaledScreenshot(), logging the error, and raising an error toast to the user before continuing.

2017-09-07 16:01:39,244 [ 1] ERROR JuliusSweetland.OptiKey.App: A DispatcherUnhandledException has been encountered...
System.ComponentModel.Win32Exception (0x80004005): 작업을 완료했습니다
   위치: System.Drawing.Graphics.CopyFromScreen(Int32 sourceX, Int32 sourceY, Int32 destinationX, Int32 destinationY, Size blockRegionSize, CopyPixelOperation copyPixelOperation)
   위치: System.Drawing.Graphics.CopyFromScreen(Point upperLeftSource, Point upperLeftDestination, Size blockRegionSize)
   위치: JuliusSweetland.OptiKey.UI.Controls.MagnifyPopup.CaptureScreenshot(Point point) 파일 F:\OptiKey\src\JuliusSweetland.OptiKey\UI\Controls\MagnifyPopup.cs:줄 160
   위치: JuliusSweetland.OptiKey.UI.Controls.MagnifyPopup.DisplayScaledScreenshot(Point point) 파일 F:\OptiKey\src\JuliusSweetland.OptiKey\UI\Controls\MagnifyPopup.cs:줄 123
   위치: JuliusSweetland.OptiKey.UI.Controls.MagnifyPopup.&lt;&gt;c__DisplayClass5_0.&lt;OnLoaded&gt;b__0(Nullable`1 sourcePoint) 파일 F:\OptiKey\src\JuliusSweetland.OptiKey\UI\Controls\MagnifyPopup.cs:줄 55
   위치: System.Reactive.AnonymousObserver`1.OnNextCore(T value)
   위치: System.Reactive.ObserverBase`1.OnNext(T value)
   위치: System.Reactive.AutoDetachObserver`1.OnNextCore(T value)
   위치: System.Reactive.ObserverBase`1.OnNext(T value)
   위치: System.Reactive.SafeObserver`1.OnNext(TSource value)
   위치: System.Reactive.Linq.ObservableImpl.Select`2._.OnNext(TSource value)
   위치: System.Reactive.Linq.ObservableImpl.Where`1._.OnNext(TSource value)
[...]

PR #391, opened 2017-10-05
Catch exceptions from copyfromscreen and display error instead

bswearingen:
Bug fix for #370.

Now when the Win32Exception is thrown, it is caught and logged, the user is notified by a toast, and the mouse simulation workflow is exited. The toast text is pretty generic, so let me know if you want me to spend some time coming up with a more specific message.

I didn't spend too much time trying to duplicate the bug, but didn't come across it while familiarizing myself with how OptiKey works. To simulate, I threw a new Win32Exception at line 184 in MagnifyPopup.cs.

Thanks,
Ben


JuliusSweetland:
@bswearingen Good stuff, thanks Ben!

Issue #393, opened 2017-10-09
Play a sound before saying the sentence.


Sheeva007:
we are a family making a lot of noise  while speaking  together ( like many othersi i presume)
So ,it's always difficult for me to let ear my "synthetic" voice. So, playing a beep or a sample (much more funny)  can be a good improvemennt.

Dominique.

zelmon64:
@Sheeva007 I believe this is related to #341.

@JuliusSweetland & @csharperperson what approach was decided for #341 and has any progress been made? I'd like to have a go at adding the attention sound key if it hasn't been started already.

JuliusSweetland:
@zelmon64 I agree a call button would be useful, rather than playing a sound before **every** spoken output (which would get a bit grating I think).

Where this key lives needs a bit of discussion. The Alpha keyboard is crowded, but it could be very useful there.

zelmon64:
@JuliusSweetland I was thinking to either put it right of the menu key or behind the scratchpad like the CK speak key. Either way I thought to probably have it hidden by default and enabled it in the management console.
![image](https://user-images.githubusercontent.com/15001389/31333202-06684336-ace1-11e7-9831-840ffe7a929e.png)

JuliusSweetland:
I think the last button on the top row would be a good starting point.

zelmon64:
Great, I'll get started on it then.

JuliusSweetland:
Great thanks

zelmon64:
@JuliusSweetland I've used a Times New Roman exclamation mark for the icon and let it use the GeneralSoundFiles (default tone1).

![image](https://user-images.githubusercontent.com/15001389/31349559-a61b6ad6-ad1b-11e7-928a-8dfe38f76b67.png)

@Sheeva007 Here's a test build. Check `Enable attention key?` on the **Visuals** tab of the **management console**. The attention sound can then be changed on the **Sounds** tab.

https://drive.google.com/open?id=0B7SYYOqNW4FSSFpKR2Ixb3pQN1E

JuliusSweetland:
Hi William,

That looks good. I'll take a look this week. One request though; there is
an actual exclamation mark key so I think it would be clearer to use an
"alert" icon, e.g. a bell icon. I am happy to find one, or if you're
interested I tend to source icons from the noun project.

zelmon64:
@JuliusSweetland I'm happy to use a different icon. The exclamation mark was just a quick placeholder. I'll see what I can find.

[...]

PR #395, opened 2017-10-12
Add Attention Sound key

zelmon64:
As discussed in issue #393, this key will play a custom attention sound in order to get the attention of nearby people before speaking. The sound is enabled and configured in the management console. The key is positioned left of the scratchpad by shortening the scratchpad by one key space when it is enabled.


JuliusSweetland:
@zelmon64 Thanks William. I will review tomorrow.

JuliusSweetland:
@zelmon64 Nice work. I've done a little bit of refactoring which you can see in commit d0e5de0 if you are interested.

PR #397, opened 2017-10-13
Fixes #396 - OptiKey starts holding down shift key

### LINKED ISSUE #396: OptiKey starts holding down shift key
When OptiKey starts, the left shift is down for ShiftAware keyboards. It results in a shift hold that affects all other programs on the system.

I guess is a good idea to start the program with shift pressed to allow first letter capitalization, but this behavior (shift down) should not affect other programs since the program was just launched and no other key was pressed on it.

This problem happens on version 2.11.0.0.

eulercamposbarros:
When Reinstate Auto Capitalisation, do not keep left shift down. Register the model with the down value but publish an up value.


JuliusSweetland:
Closing for now while we discuss on the issue.

Issue #383, opened 2017-09-28
Simplify keyboard selection in settings


kmcnaught:
Currently if your language is english, the management console allows you to replace the alpha keyboard with  a simplified keyboard layout, an alphabetical layout, or a Communikate layout. Each of these options appears/disappears based on other settings. It would be clearer from a user's perspective to consolidate these into a single option such as "Keyboard style for typing" where the options are qwerty/alphabetic/simplified/communikate.

JuliusSweetland:
I agree with this idea

zelmon64:
I'm going to have a go at this today if that's alright. I thought it would be nice to have for the next release.

JuliusSweetland:
Great thanks

PR #400, opened 2017-10-15
Simplify keyboard selection in settings

zelmon64:
As discussed in #383 the checkboxes for keyboard selection have been
replaced with a single drop down list.


JuliusSweetland:
@zelmon64 This looks good. I'll have to sit down and test it properly this week. Thank you!

Issue #408, opened 2017-10-20
Manual mode isn't working if MyGaze/EyeTribe isn't connected


kmcnaught:
Version: 2.11.0

- Set up OptiKey to use MyGaze or EyeTribe (but these aren't installed/connected)
- OptiKey restarts to use the chosen eye tracker
- Obviously eye tracking doesn't work
- Right click -> Manual mode
- OptiKey enters manual mode
- Manual mode doesn't work (no mouse progress is registered, you cannot interact). 

If you do the same thing with GazeTracker or EyeX selected, then manual mode works okay. I haven't tested all the other input options.

I've also replicated this on my Win8.1 PC and another running Win 10.

JuliusSweetland:
Fix in 2.12.0 pre_release_2 (https://github.com/OptiKey/OptiKey/releases/tag/v2.12.0_pre-release_2)

kmcnaught:
I've just tested this on v2.12.0_pre-release_2 and the bug still exists but later in the workflow:

- Set up to use MyGaze (which isn't installed or connected)
- Launch OptiKey
- You now get the "Uh oh" error message, great
- Turn on manual mode
- Manual mode works
- Turn off manual mode
- Turn on manual mode
- Manual mode does **not** work this time.

kmcnaught:
Might be worth noting:
- the above behaviour also occurs with EyeTribe as input
- With the EyeX or Gaze tracker as input the behaviour is as follows: (in the case of EyeX, it is installed but not connected)
    - set up to use [eyex or gazetracker]
    - launch optikey
    - you do not get any error message (perhaps for EyeX this is because the service connects ok, but just responds as being in invalid state)
    - you can toggle manual mode multiple times and it always works...

Chopinsky:
The 1st problem seems to be caused by raising 2 roast notification simultaneously -- 1 for manual mode, 1 for error to initialize MyGaze, and only 1 callback to the notification popup are invoked (because the 1st popup thinks it's not closing because the 2nd popup is raised, but it is closing), which mess up the keyboard shield count (in a similar way to the original issue), and hence leaving a shield up even if all popups have been dismissed.

JuliusSweetland:
@kmcnaught Fix is now merged into master and will be included in the next release (I have not prepared this yet)

Issue #409, opened 2017-10-20
Not getting error messages when eye tracker can't connect


kmcnaught:
Version: 2.11.0.0

When I try to connect to a non-existent or disconnected eye tracker, I'm not getting any user-facing error messages. (there are errors in the logs though). I'm sure we used to get user error-messages (red toasts) along the lines of "Uh oh, can't connect to xx"...

On my own PC (Win8.1) I *never* get any of these messages running v2.11.0.0 (have tested EyeX/MyGaze/EyeTribe/Gaze Tracker). On a different (Win10) PC, using EyeX, I get the error messages *after* turning manual mode on and then off again, i.e.
- Launch OptiKey using an eye tracker that isn't connected
- No error message appears, it just silently doesn't work
- Turn on Manual Mode
- Turn off Manual Mode
- An "Uh oh" message immediately appears.

With MyGaze on the 2nd PC I don't get any error message. I haven't tested other eye trackers.

JuliusSweetland:
Fix in 2.12.0 pre_release_2 (https://github.com/OptiKey/OptiKey/releases/tag/v2.12.0_pre-release_2)

PR #410, opened 2017-10-25
Make "Manual Mode" usable if eye tracker's failed to launch

Chopinsky:
This PR is aimed to resolve issue #408, where if the eye tracker can't be activated (e.g. the device is not connected, the device's failed to initialize, etc.), the input service isn't properly resumed in the initialization process, and as a result, users won't be able to switch to "Manual Mode" because the shield count is incorrectly set during the initialization process. 

The root cause is actually the same to issue #409: the eye tracker service is initialized before the `ToastNoticifcationPopup` is loaded; and when the activation failed, we would try to raise a `ToastNotificationPopup` to inform the users, however, since the popup's not available yet at that moment, the popup will not be raised. But the effort to raise the popup have added a shield on top of the keyboard, but the shield is never lowered (the callback to the popup is responsible for lowering the shield, since the popup is never launched, the callback is never invoked either), which will then block the workflow to switch to the manual mode.

The solution is to add rail guards around the error handlers: 1. if the popup hasn't been loaded yet, don't try to raise it and bring up the shield; 2. if someone tries to bring up the popup prematurely in the future, we will fallback on calling the callback directly to clean anything up, such that we won't accidentally block the keyboard because the clean-up callback is not raised. 

Be noted that this is not the full solution to issue #409, it's kind of difficult to make sure the popup is loaded before all the services are initialized -- there're quite entangled code dependencies and assumptions made in the app-launching code, so I would envision a little bit more sophisticated fix to issue #409: (the easier solution) save off all the errors encountered during service initialization, then bring up the toast notification after the main window is fully loaded (before or after the splash popup); (the harder solution) if eye track devices are failed to activate, fallback to the mouse pointer mode (aka the Manual Mode).

### DIFF
--- a/src/JuliusSweetland.OptiKey/App.xaml.cs
+++ b/src/JuliusSweetland.OptiKey/App.xaml.cs
@@ -210,13 +210,13 @@ namespace JuliusSweetland.OptiKey
                 {
                     mainWindowManipulationService.SizeAndPositionInitialised -= sizeAndPositionInitialised; //Ensure this handler only triggers once
                     await ShowSplashScreen(inputService, audioService, mainViewModel);
+                    await mainViewModel.RaisePreloadErrorsToastNotification();
                     await AttemptToStartMaryTTSService(inputService, audioService, mainViewModel);
                     await AlertIfPresageBitnessOrBootstrapOrVersionFailure(presageInstallationProblem, inputService, audioService, mainViewModel);
 
                     inputService.RequestResume(); //Start the input service
 
                     // Raise any preload errors in the toast notification now.
-                    mainViewModel.RaisePreloadErrorsToastNotification();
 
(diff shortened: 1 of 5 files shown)


JuliusSweetland:
@Chopinsky Thanks for this. If you have time would it be possible to look into a fix for #409 as that would solve this issue in a more complete way. This fix, while effective, is a bit of a work around. I'm not sure the user hearing the error tone, but with no indication as to why, would be ideal.

Chopinsky:
I've some rough ideas on how to implement the "easy solution" I mentioned, will work on that.

Chopinsky:
New commit allow us to cache pre-load error messages, and prompt them to end users in the roast notifications right after the splash screen (or if no splash screen, right after launching).

JuliusSweetland:
@Chopinsky Like it, thank you! Merged.

PR #413, opened 2017-11-01
Changes/fixes for dynamic keyboard

kmcnaught:
New functionality:
- you can specify key state overrides, e.g. "magnetic cursor key should be forced to be up when entering this keyboard". Any overridden keys are reset to their original state when leaving the keyboard
- you can request an "output panel" at the top of your dynamic keyboard. In future this might be split further to have (a) autocomplete suggestions and/or (b) scratchpad, but for now you get all or nothing.
- you can choose as a startup keyboard, either:
  - the dynamic keyboard menu, which loads any files it finds in your dynamic keyboards folder; or
  - a specific dynamic keyboard. This would allow you to define your own "menu" keyboard, for example.

Fixes/changes:
- Fixes a Win10 bug which caused keyboards with a height specified to jump around the screen. Now resizing is done on enter/exit rather than on response to OptiKey resizing. 
- Small changes to layout of dynamic menu, and friendlier error handling.


JuliusSweetland:
@kmcnaught Thanks Kirsty!

PR #414, opened 2017-11-01
Force minimum time for toast notifications

kmcnaught:
(feel free to change the constant minimum time)

This improves small toasts, like "Manual mode ..." which disappeared very quickly


JuliusSweetland:
@kmcnaught Could you move the constant into a setting. No need to surface this in the management console, but a setting makes it slightly accessible.

JuliusSweetland:
@kmcnaught Closing this PR so that it can be replaced once the setting is in place.

kmcnaught:
fyi, this is on my todo.

PR #416, opened 2017-11-04
Raise notifications sequentially

Chopinsky:
This PR is to address the new problems found with issue #408. 

Right now there're 2 things wrong with the Roast Notification Popup:
- If 2 notifications are raised simultaneously (e.g. a normal information popup + an error/warning popup), only the 2nd notification will be displayed. This is because only 1 popup form is brought up, but its content are set by the 1st notification request and then overridden by the 2nd popup request. This is slightly annoying because usually it is the informative popup request that's overridden, so we lose 1 popup but are still able to see the error popup.
- The 2nd issue is more serious -- before raising the notification, we usually raise a shield to protect the keyboard from being interact-able when the popup is on, and the shield will be lowered after the popup is gone. However, when 2 notification are raised together, only 1 popup-close-callback will be invoked, causing the shield raised by the 1st notification up forever, and make the keyboard unusable if this takes place. 

Now in this PR, we're going to revise how the popup are raised. We will use the Producer-Consumer pattern, such that all notifications that are raised together will be queued and displayed in sequential orders. And since they'll be raised one after the other, each of them will invoke the post-close-callback and clean up any shields raised by the pre-raise code.

### DIFF
--- a/src/JuliusSweetland.OptiKey/UI/Controls/ToastNotificationPopup.cs
+++ b/src/JuliusSweetland.OptiKey/UI/Controls/ToastNotificationPopup.cs
@@ -21,7 +21,6 @@ namespace JuliusSweetland.OptiKey.UI.Controls
         private Screen screen;
         private ToastNotification toastNotification;
         private Queue<Models.NotificationEventArgs> messageQueue;
-        private int openPopupCounts;
 
         #endregion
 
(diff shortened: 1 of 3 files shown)


JuliusSweetland:
@Chopinsky Nice change. Thank you!

Issue #418, opened 2017-11-18
Prevent crash from incorrect rectangle size (not sure why though?!)


JuliusSweetland:
I have seen this issue a number of times. The error message is essentially "Width and Height should not be negative." and refers to creating a Rect with a negative width or height (I think) at this line:
https://github.com/OptiKey/OptiKey/blob/8cb69d0ec5b0c1d57047613c41675f30b025bda1/src/JuliusSweetland.OptiKey/Services/WindowManipulationService.cs#L1415

I think we just need a safeguard check to discard any bad values and return from that method.

 <Message>A Largura e a Altura não devem ser negativas.</Message>
  <Source>WindowsBase</Source>
  <StackTrace>   em System.Windows.Rect..ctor(Double x, Double y, Double width, Double height)
   em JuliusSweetland.OptiKey.Services.WindowManipulationService.SetAppBarSizeAndPosition(DockEdges dockPosition, Rect sizeAndPosition, Boolean isInitialising, Boolean persist) na C:\Users\Julius\Documents\GitHub\OptiKey\src\JuliusSweetland.OptiKey\Services\WindowManipulationService.cs:linha 1415
   em JuliusSweetland.OptiKey.Services.WindowManipulationService.AppBarPositionChangeCallback(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean&amp; handled) na C:\Users\Julius\Documents\GitHub\OptiKey\src\JuliusSweetland.OptiKey\Services\WindowManipulationService.cs:linha 635
   em System.Windows.Interop.HwndSource.PublicHooksFilterMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean&amp; handled)
   em MS.Win32.HwndWrapper.WndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean&amp; handled)
   em MS.Win32.HwndSubclass.DispatcherCallbackOperation(Object o)
   em System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   em System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)</StackTrace>
  <TargetSite>Void .ctor(Double, Double, Double, Double) @ System.Windows.Rect</TargetSite>
  <Type>System.ArgumentException</Type>
[...]

PR #419, opened 2017-11-18
Adding safeguard when setting window height/width

Chopinsky:
As mentioned in issue #418, there are occasions when calculations of docking window size would go south, and the final width or height may end up with negative values. When trying to create a window of these negative values, OptiKey will crash because of that. 

Based on the code review, this shouldn't happen, though `barData` could be potentially adjusted by PInvoke method, and hence causing incorrect width/height calculations. 

As suggested in the original issue, we will abort resizing/docking if new window's size are negative values.


JuliusSweetland:
@Chopinsky Nice PR, thank you.

Issue #423, opened 2017-11-30
Add alternate mouse scrolling method


JuliusSweetland:
Looking at the centre of the screen stops scrolling, but looking above/below the centre point scrolls up/down.

Horizontal scrolling could also be achieved.

It might make more sense to allow the user to drop the centre point themselves, so that focus can be given to the window chosen (optional). This would also avoid the potentially wrong assumption that the user always wants to scroll the window under the centre point of the screen.

JuliusSweetland:
Looking at the centre of the screen stops scrolling, but looking above/below the centre point scrolls up/down.

Horizontal scrolling could also be achieved.

It might make more sense to allow the user to drop the centre point themselves, so that focus can be given to the window chosen (optional). This would also avoid the potentially wrong assumption that the user always wants to scroll the window under the centre point of the screen.

joshlcampbell:
I will begin looking into this issue.

JuliusSweetland:
Fantastic, thank you. Happy new year.

joshlcampbell:
Hello again, and happy new year to you as well! 

Happily, I'm able to start working on this in earnest now, but I do have a quick question: do you envision this alternate scrolling method (which I'm tentatively labeling "look to scroll" though I'm open to another name) being used in conjunction with the current methods or more as a replacement (chosen via the Management Console)? 

The reason I ask is to determine whether I need to try to make room for new keys on the Mouse keyboard or can re-use the space given to the existing scrolling-related keys.

JuliusSweetland:
Hi Joshua,

I initially thought this would be an alternative method, chosen from the
management console, but the current scroll functionality has 5 keys
(including the scroll amount) on the expanded mouse keyboard, so there
isn't a straightforward way to replace this with a single new scroll
method. If possible so you think the new scroll method could fit somewhere?
I think a management console setting could control which method appears on
the collapsed view though.

joshlcampbell:
I am thinking -- if you're willing to have this alternative method replace the other via settings -- that I might replace those 5 keys with 5 of my own:

1. Vertical Scroll: look above the center point (which I'm tempted to call the "anchor" to avoid confusion with the screen center) to scroll up and look below the "anchor" to scroll down.
2. Horizontal Scroll: likewise, look left of the anchor to scroll left and right of the anchor to scroll right.
3. Free Scroll: a combination of the previous two.
[...]

PR #431, opened 2018-02-07
Add alternative "look to scroll" method of scrolling

joshlcampbell:
This pull request addresses issue #423. 

When enabled from the Other tab of the management console, it replaces the original scrolling-related keys of the Mouse and Web Browsing keyboards with new "look to scroll" keys:

- **Look to Scroll Active:** while locked down, scrolling messages will continually be sent to the inputsimulator based on the currently chosen mode, bounds, and speed. When pressed, if using Window or Custom bounds and not re-using the previously selected window or custom rectangle, you'll need to choose a new window or rectangle.
- **Look to Scroll Mode:** cycles between Vertical, Horizontal, Cross, and Free modes. In Vertical mode, looking above the center of the current scroll bounds scrolls up and looking below it scrolls down. Similarly, in Horizontal mode, looking left of center scrolls left and looking right of center scrolls right. Free mode is a combination of the two; for example, you can look near the top-right corner of the scroll bounds to scroll both up and to the right simultaneously. Cross mode is similar to Free mode except it only allows you to scroll along one axis at a time by looking inside the "cross" shape formed by summing each axis's deadzone; so in order to scroll vertically you have to be looking inside the horizontal deadzone, and _vice versa_. Thus, it is advised that if you want to use Cross mode that you not get rid of either deadzone completely or else you'll have no area in which to look to scroll.
- **Look to Scroll Bounds:** while scrolling is not active, cycles between Screen, Window, and Custom bounds. This determines the rectangular region of the screen in which scrolling can take place. With Screen bounds, the entire screen is used, unless OptiKey is docked in which case only the remaining non-OptiKey portion of the screen is used. With Custom bounds, you choose the rectangular region yourself by placing any two corners of it on-screen. With Window bounds, you choose an open window, and its bounds becomes the scroll bounds. Note that unlike with Screen and Custom, you must be looking at the same window you originally selected in order to scroll. Also, you can't cycle between the 3 types of bounds while scrolling is active. Instead, you'll see the bounds key become locked down. Pressing it in this state will have no effect for Screen bounds, will allow you to select a different window for Window bounds, or will allow you to select a different rectangle for Custom bounds. You'll notice the bounds key remains locked down after scrolling is deactivated. This allows you to re-use the same scroll bounds the next time you start scrolling, without having to re-select a window or custom rectangle. Unlocking it will allow you to resume cycling between the types of bounds and discards the previous scroll bounds.
- **Look to Scroll Speed:** cycles between Slow, Medium, and Fast speed presets. Each speed preset has two components: a base speed and an acceleration. The base speed is the fixed scroll amount in "clicks" per second that gets sent to the inputsimulator when looking any distance above, below, left of, or right of the deadzone, a small rectangular region surrounding the center of the scroll bounds. The acceleration is the additional "clicks" per second added to the base speed for every pixel away from the deadzone that you're looking. The base speeds and accelerations for these presets as well as the size of the deadzone can be configured in the management console.
- **Look to Scroll Increment:** cycles between several increment amounts which are in raw mouse wheel units (120 per "click" of the wheel). All vertical and horizontal scroll amounts sent to the inputsimulator will be multiples of this amount. Any leftover amounts get carried over into subsequent messages. This can be useful when working with applications that only respond to whole "clicks" or specific fractions of a "click" (1/2, 1/3, 1/4, and so on) like some text editors. The list of amounts that are cycled between can be customized from the management console.

All five new keys appear on the expanded Mouse keyboard. The Active and Bounds keys appear when the Mouse keyboard is collapsed. The Web Browsing keyboard contains all of them except the Increment key. The fixation delays for all five are configurable from the management console in a new "Look to Scroll" key group.

To aid in getting used to where to look for the various scroll modes and types of bounds, I've included an overlay window feature. While actively scrolling, it will draw an outline around the current scroll bounds as well as guide lines to show the extent of the deadzone area. The overlay can be toggled on/off and the colors/thicknesses of its lines customized in the management console.

Other settings implemented but not mentioned thus far are:

- Inverting the scrolling direction
- Centering the mouse cursor inside the scroll bounds upon activating look to scroll
- When using Window bounds, activating and bringing the chosen window into focus upon activating look to scroll

As part of my work on this issue, I had to make some minor changes to the inputsimulator in order to support sending partial "clicks" for smoother scrolling. I've updated the binary file (WindowsInput.dll) here with one built from my own fork of your fork of the inputsimulator repository. Knowing how leery I myself would be at accepting some random person's binary file without question, I wouldn't find it amiss if you'd prefer to recompile it yourself and push your own version. To that end, I'll be creating a corresponding pull request in your fork of the inputsimulator shortly. (Here it is: JuliusSweetland/inputsimulator#1)

Please let me know if you find any issues with the code or need clarification as to how part of it works. I must admit: I'm quite curious to know how well it will work with a true independent point source. Up to this point (hah!), I've been using a webcam-based method to control the mouse, but I I'll probably fork (hah again!) over the money for one of the Tobii eye trackers soon. I think it may end up working quite a bit better in fact since you wouldn't have to deal with some of the focus issues with the mouse while scrolling. (Like not being able to scroll a window unless your mouse is over it. I could've sworn in previous versions of Windows it would always send scrolls to whatever window had keyboard focus, but I must be mistaken.)

Thanks!


JuliusSweetland:
@joshlcampbell This looks excellent! I'll be testing it over the next few days. Thank you!

JuliusSweetland:
@joshlcampbell The new svg icons - do we need to add to the Credits.txt file to correctly attribute any artwork that has been used?

JuliusSweetland:
From initial testing it looks amazing. Really, really impressive!

I think free scroll should be the default, maybe with a larger default number of scroll units. What do you think?

There's also an interesting use case where custom area doesn't work until the window is focused. You alluded to this already.

image

Overall fantastic though.

joshlcampbell:
I forgot to mention that the mode, bounds, speed, and increment all carry over between sessions since the current value comes straight from / is written back to Settings.Default while cycling -- similar to how the original scroll amount key works. So the defaults really only effect the first use. I agree, though, that Free scroll should be the default, and I can certainly up the default scroll increment as well. Perhaps 30, as in your screenshot? My thinking with the default of 1 was to make scrolling as smooth/precise as possible out of the box, with the ability to bump it up for apps that can't make use of it or if you prefer a coarser scrolling experience.

For this use case you mentioned, sort of a combination of Custom and Window bounds types: a Sub-Window type? Select two points to define a rectangle -- as in Custom -- but where the two points must be within the same window and where scrolling only happens when that window is being looked at -- as in Window. I think that'd be fairly straight-forward to add.

JuliusSweetland:
Yes the sub window type sounds good. All of these modes are really
powerful, but possibly a bit confusing. Do you think there could be an
alternative scrolling option (keep all of this, but add the following
idea?) where the user selects only the centre point. This point is,
optimally, clicked to focus it, and then scrolling happens as a function of
how far away from the point and in which direction the user is looking.
There would be the usual central dead zone, and scrolling would stop when
the user selects the key again. It's similar to free scroll, but without
having to define the scroll area. Instead you define the central point only
and scrolling accelerates the further you look away from the central point.
Does that make sense? What do you think?

joshlcampbell:
So... implement the functionality the issue was actually asking for before I went and over-complicated the heck out of things? Yes, I can (and probably should) do that. :)

The most expedient way to get there from this point, I think, would be via a new "no bounds" or "around point" bounds type (the new default) plus a setting to send a mouse click when setting the point (defaulted to true). So the cycle would be: "Scroll around Point" -> "Scroll within Screen" -> "Scroll within Window" -> "Scroll within Window Area" -> "Scroll within Custom Area".

I'm not sure if that addresses the "confusing" aspect of all this functionality enough, though. Perhaps some of the it is due to the behavior of the bounds type key? Perhaps we make it a straight-forward cycle like the others and either get rid of the current locked-down behavior or re-introduce it with a setting (that's off by default)?

Or are you picturing an even more streamlined UI for this alternative mode, like just a single key for "Look to Scroll" with everything else (mode, bounds, speed, and increment) customizable via the management console?

JuliusSweetland:
Yes that sounds great! It's brilliant as it is, and I don't think we should
lose any of the functionality you've already made, but this would be a
great addition.

JuliusSweetland:
@joshlcampbell Merged as this is a great PR on its own. The discussed changes can form a new PR if you're happy with that.

joshlcampbell:
Thanks @JuliusSweetland, that sounds good. I hope to get to those changes this weekend, and I'll try to write up a page of documentation (probably based on this PR's description) that you can put on the wiki.

JuliusSweetland:
Fantastic, thanks Josh. After testing I'd like to consider whether this
should become the default scrolling method, or if there should be a quick
way to switch between methods without going to the management console. Do
you have any thoughts on this?

joshlcampbell:
Hmm...  although I think the new method is superior (if you'll forgive the immodesty), as you mentioned before, it's also more complex and possibly confusing. For someone just learning OptiKey, the original scrolling keys seem to fit in much better with how the rest of the keys in the app work, particularly the other "do at point" keys on the Mouse keyboard. You learn how one works, and you've learned how they all work. I like the idea of opting into additional complexity vs. opting out of it, so I would personally argue that the original keys be the default ones seen when launching OptiKey for the first time.

That said, I'm not opposed to adding a way to switch to the "look to scroll" keys on the keyboard itself -- to opt in/out of the complexity as needed -- but I don't immediately see a good place to put this toggle. I'm tempted to hijack part of the "Do at point..." label, as seen below.

keyboard_mouse_numbered

You'd see the "look to scroll" key icon when using the original keys and another icon using the same symbols as the original keys when using the "look to scroll" keys. I'm worried that might be too small a target for users to hit, though.

JuliusSweetland:
I take your point about opting into complexity - I agree. I think maybe we
keep it as it is by default and have your new scrolling functionality as
options on the other tab. I think the small key on the mouse keyboard would
look out of place. I like the balance that (most) keyboards currently have.
It's fairly minor, but I like to think if users have to stare at OptiKey
all day it at least looks good!

PR #436, opened 2018-02-17
Fix invisible whole-screen window being selected when UIAccess enabled

joshlcampbell:
-  Filter out the lock screen, Start screen, and desktop wallpaper manager from window selection.
- Take into account the top-level windows left around by UWP apps even while the app isn't running. Use their CoreWindows in the selection process instead.
- Add a few notes about getting OptiKey running with UIAccess
- Get rid of WS_EX_NOACTIVATE check. This was to fix a problem only occurring during development when running out of Visual Studio and was due to the presence of WPF's debugging overlay window. Rather than possibly excluding legitimate windows that could otherwise be selected just to fix something only a developer would see, I've gotten rid of the check and just added some notes that hopefully a developer will find if they run into the problem.
- Fix potential exception in GetFrontmostWindow due to lazy iteration. hWnds.Any() returned true, the window was immediately closed by the user afterwards, and then it proceeded to try to call hWnds.First().


JuliusSweetland:
Very high quality PR. Merged. Thank you!

PR #465, opened 2018-06-07
Added Simplified Keyboards

samgled:
I've added two simplified keyboards and edited some files to add the "Bindings"? ("AlphaGroup4All" and so on).

But I'm not sure how to make OptiKey properly aware of the keyboards. I don't think it knows SimplifiedAlpha2.xaml is there. I also have had no luck figuring out how to add the button to switch from the first keyboard to the second.

I'm hoping you can complete this for the Japanese translation @JuliusSweetland


JuliusSweetland:
@samgled This looks like the beginning of something good. It's difficult for me to be able to test it fully as I don't understand Japanese, but I've got two pieces of feedback:

1.This symbol doesn't appear to work (unless I'm using it incorrectly): 
image

2.The ability to change between keyboard 1 and 2 you need to set the value of the next key on Alpha 1 to "{x:Static models:KeyValues.Alpha2KeyboardKey}"
The same key on the alpha 2 keyboard would have value "{x:Static models:KeyValues.Alpha1KeyboardKey}"

Also there are a couple of minor fixes to enable simplified keyboards (and the ability to switch between simplified alpha 1 and 2) for Japanese. I've uploaded them on this branch: https://github.com/OptiKey/OptiKey/tree/samgled-master

samgled:
@JuliusSweetland 
Thanks! Those fixes have done the trick.

I've had a quick check and everything seems to be working. There doesn't seem to be any issue with the 小 button you highlighted. It changes the previous character to a smaller version (but it doesn't work with every character). You can try it with the first set of characters (あ -> あ い う え お) if you'd like to see it work.

I'll send a merge request when I get home. There's an issue with the transifex language file I'd also like to include.

Thanks again!

samgled:
Updated the translation.

JuliusSweetland:
@samgled The pull request has not been updated with the new translation? Do you mean the translation on Transifex?

samgled:
I did try updating through github with the pull request but it kept going
to "Compare" and wouldn't let me push...

samgled:
@JuliusSweetland Ok, maybe it was because I had a pull request open. Sorry, still haven't figured out github very well. I closed and tried again.

samgled:
Just emailed zip file too.

PR #470, opened 2018-06-30
CommuniKate for all languages and Presage contraction support

zelmon64:
The CommuniKate keyboard can now be enabled and used by any language. To access it a key has been added to the Output section.

Apostrophe support has been added to Presage along with the database from ACAT which contains contractions.

### DIFF
--- a/src/JuliusSweetland.OptiKey/UI/ViewModels/Management/WordsViewModel.cs
+++ b/src/JuliusSweetland.OptiKey/UI/ViewModels/Management/WordsViewModel.cs
@@ -242,6 +242,53 @@ namespace JuliusSweetland.OptiKey.UI.ViewModels.Management
             set { SetProperty(ref usingCommuniKateKeyboardLayout, useCommuniKateKeyboardLayoutByDefault); }
         }
 
+        private int communiKateSoundVolume;
+        public int CommuniKateSoundVolume
+        {
+            get { return communiKateSoundVolume; }
+            set { SetProperty(ref communiKateSoundVolume, value); }
+        }
+
+        private bool communiKateSpeakSelected;
+        public bool CommuniKateSpeakSelected
+        {
+            get { return communiKateSpeakSelected; }
+            set { SetProperty(ref communiKateSpeakSelected, value); }
+        }
+
+        private int communiKateSpeakSelectedVolume;
+        public int CommuniKateSpeakSelectedVolume
+        {
+            get { return communiKateSpeakSelectedVolume; }
+            set { SetProperty(ref communiKateSpeakSelectedVolume, value); }
+        }
+
+        private int communiKateSpeakSelectedRate;
+        public int CommuniKateSpeakSelectedRate
+        {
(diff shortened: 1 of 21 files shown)


src/JuliusSweetland.OptiKey/UI/ViewModels/Management/WordsViewModel.cs
```diff
@@ -147,7 +147,8 @@ public bool UseDefaultKeyboardLayoutIsVisible
                        || KeyboardAndDictionaryLanguage == Enums.Languages.EnglishUK
```
JuliusSweetland: Could you please change this to instead check UseAlphabeticalKeyboardLayoutIsVisible || UseSimplifiedKeyboardLayoutIsVisible || EnableCommuniKateKeyboardLayoutIsVisible

src/JuliusSweetland.OptiKey/UI/ViewModels/Management/WordsViewModel.cs
```diff
@@ -111,7 +111,7 @@ public Languages KeyboardAndDictionaryLanguage
                 SetProperty(ref this.keyboardAndDictionaryLanguage, value);
                 OnPropertyChanged(() => UseDefaultKeyboardLayoutIsVisible);
                 OnPropertyChanged(() => UseAlphabeticalKeyboardLayoutIsVisible);
-                OnPropertyChanged(() => EnableCommuniKateKeyboardLayoutIsVisible);
+                OnPropertyChanged(() => EnableCommuniKateKeyboardLayout);
```
JuliusSweetland: Shouldn't this be UseCommuniKateKeyboardLayoutIsVisible?

JuliusSweetland:
@zelmon64 Hi William - this looks good, but I'm at work so I can't test for a couple of days. I've added one very minor comment - if you have time to change it that would be great, otherwise I'll change it when I test. Thanks again!

JuliusSweetland:
@zelmon64 Few questions:

1.How does the presage database setting in OptiKey (in the Management Console) work with the presage xml config value, which I think is this one:
<Predictors>
        <DefaultSmoothedNgramPredictor>
            <PREDICTOR>SmoothedNgramPredictor</PREDICTOR>
            <LOGGER>ERROR</LOGGER>
            <DBFILENAME>::PACKAGEDATADIR::/database_en.db</DBFILENAME>

??? Does OptiKey's setting (if set) override the presage setting?

2.Do I need to update this page to change the customisation of the xml file in step 5? https://github.com/OptiKey/OptiKey/wiki/Next-word-prediction

3.Do I need to also change that page to change how the user changes the presage language (bottom of page https://github.com/OptiKey/OptiKey/wiki/Next-word-prediction)

zelmon64:
@JuliusSweetland 

1. The new settings get saved to `C:\Users\user\.presage\presage.xml` and override the default presage settings.
2. Step 5 can be removed. Those settings are now set automatically.
3. Yes, to use a different language database it must now be set from the Management Console.

PR #496, opened 2018-10-01
Service endpoint

puneetpushkar:


JuliusSweetland:
@puneetpushkar This appears to be my code from the service-endpoint branch. At this point I'm not merging this into master as it means that OptiKey requires admin privilege to run.

PR #498, opened 2018-10-02
Plugin extensibility feature

dptucunduva:
* Automatically load plugins when starting.
* New Dynamic Keyboard key type (PluginKey) that calls plugins
* Configurable plugin directory
* Two built-in plugins:
   * HTTP Call Plugin: Calls a HTTP/HTTPS URI, using POST or GET methods
   * External Program Plugin: Calls a windows application
 * New Wiki on how to configure built-in plugins
 * New Wiki on how to create your own plugin


JuliusSweetland:
@dptucunduva Hi Daniel - I'm getting a compilation error on this PR:

"C:\projects\optikey\OptiKey.sln" (default target) (1) ->
"C:\projects\optikey\src\JuliusSweetland.OptiKey.StandardPlugins\JuliusSweetland.OptiKey.StandardPlugins.csproj" (default target) (6) ->
(CoreCompile target) -> 
  Network\HttpCall.cs(61,46): error CS1525: Invalid expression term 'int' [C:\projects\optikey\src\JuliusSweetland.OptiKey.StandardPlugins\JuliusSweetland.OptiKey.StandardPlugins.csproj]
  Network\HttpCall.cs(61,50): error CS1003: Syntax error, ',' expected [C:\projects\optikey\src\JuliusSweetland.OptiKey.StandardPlugins\JuliusSweetland.OptiKey.StandardPlugins.csproj]

Could you please take a look?

FYI I've copied over the wiki images and files so they're ready to go.

dptucunduva:
@JuliusSweetland Hi Julius. It seems to be something related to a new C# notation that is still not supported by AppVeyor compiler version. I changed it to a "traditional" way and it seems to be OK.

BTW, did not notice that OptiKey was using a CI system. Will check build results asap next time.

JuliusSweetland:
@dptucunduva 

Creates path C:\Users\Julius\AppData\Roaming\JuliusSweetland\OptiKey\Plugins\ then attempts to loop through pluggins in a directory which doesn't exist:

"Could not find a part of the path 'C:\\Users\\Julius\\Documents\\GitHub\\OptiKey\\src\\JuliusSweetland.OptiKey\\bin\\Debug\\Resources\\Plugins'."

I presume these would be included in the solution, but I don't have anything:
image

Is this a genuine bug (which I don't think I'll see again as the target directory has now been created so this code will be skipped in future), or something else? Thanks.

JuliusSweetland:
If I build the standard plugin project and add the dll as content to the main project Resources directory then everything is fine, HOWEVER, for runtime performance I would like to embed the release build of the OptiKeyStandardPlugins.dll, but this gives debugging warnings:

image

Any chance you could add build steps to copy over the appropriate debug/release version of the OptiKeyStandardPlugins.dll to the Resources\Plugins directory?

dptucunduva:
> @dptucunduva If I build the standard plugin project and add the dll as content to the main project Resources directory then everything is fine, HOWEVER, for runtime performance I would like to embed the release build of the OptiKeyStandardPlugins.dll, but this gives debugging warnings:
> Any chance you could add build steps to copy over the appropriate debug/release version of the OptiKeyStandardPlugins.dll to the Resources\Plugins directory?

@JuliusSweetland Standard plugins DLL is copied only in the packaging process (batch files and inno scripts). When packaging a release version of OptiKey, release versions of plugins are copied to Resources\Plugins directory. However, as you pointed out, this is not happening when running the application directly in Visual Studio. I was copying manually when I was developing/testing here in my machine.

I will add build steps (just need to figure out how to do it :) ).

JuliusSweetland:
@dptucunduva Fantastic, thanks. It will help people debug the code.

One more for you - when testing the APICALL key in the BigHackerKeyboard I can trace through that an Http GET call should be made, and all the args are applied, but nothing happens when the key fires. Is this something to do with the privilege level that OptiKey runs at, or is there another issue there? Thanks.

dptucunduva:
> @dptucunduva I've encountered a small bug...
> 
> Creates path C:\Users\Julius\AppData\Roaming\JuliusSweetland\OptiKey\Plugins\ then attempts to loop through pluggins in a directory which doesn't exist:
> 
> "Could not find a part of the path 'C:\Users\Julius\Documents\GitHub\OptiKey\src\JuliusSweetland.OptiKey\bin\Debug\Resources\Plugins'."
> Is this a genuine bug (which I don't think I'll see again as the target directory has now been created so this code will be skipped in future), or something else? Thanks.

Yes, it is indeed a bug. I had to delete everything, checkout the code again to reproduce this error. Will work on it now.

dptucunduva:
> One more for you - when testing the APICALL key in the BigHackerKeyboard I can trace through that an Http GET call should be made, and all the args are applied, but nothing happens when the key fires. Is this something to do with the privilege level that OptiKey runs at, or is there another issue there? Thanks.

@JuliusSweetland Did you get an error message in your screen (that big red scarry message box)? If not, I guess that it means that the request was successful and an HTTP code 2XX (200 family) was returned. So, everything went OK.

The idea of this plugin is to interact with other devices/systems - So, the user will know that the request went OK because the target system will do something. If the request timed out (or an 500 error is returned), the user gets the red message box. Example: I am working with this HTTP Call plugin to interact with my "CASA" system (project in my github page). Basically HTTP requests that change the TV channel, adjust the height of the bed, adjust A/C temperature, etc. 

Of course we should add a way for the plugins to interact with OptiKey bi-directionally. But, as this would take another several hours of work, I think we should leave it for a next release. What do you think?

JuliusSweetland:
No error message for me. What would you expect to happen? Does it call the
URL silently?

I'm happy to leave the plugin architecture as one way until we need it for
something.

It looks great by the way - really powerful.

JuliusSweetland:
What I think optikey needs now is an easier way to integrate and use the
dynamic keyboards. It's not particularly obvious at the moment.

dptucunduva:
> No error message for me. What would you expect to happen? Does it call the URL silently? 

Yes, exactly. 

BTW, I've just commited the post-build steps. We got rid of both problems with this: The error when checking for plugins in a directory that doesn't exist and plugin debugging/release.

And also another thing: That change in packaging batch file that i sent you can be ignored. As post-build step is already copying DLL files, you do not need to change your packaging anymore.

JuliusSweetland:
That's great. I'll test tomorrow. Elegant fix for three things in one!

JuliusSweetland:
@dptucunduva I can't see the post build step in this PR? Did you update the pull request?

dptucunduva:
@JuliusSweetland Yes, it is here:

https://github.com/OptiKey/OptiKey/pull/498/commits/94e582edb468b73c94c1bd877f33658853d9855a

> @dptucunduva I can't see the post build step in this PR? Did you update the pull request?

JuliusSweetland:
@dptucunduva Merged. I'm going to merge this into master now. Great PR, thank you!

PR #500, opened 2018-10-03
Add issue template for bugs

Kokan:
Based on the README I created this issue template, so when somebody opens a new issue they already saw what kind of information should be supplied.


JuliusSweetland:
@Kokan Thanks!

PR #511, opened 2018-10-12
Issue #495 : Refresh Icon in Menu and Context Menu

LauraLaureus:
Added Refresh button to menu and context menu. 
Issue #495


LauraLaureus:
I'll need some time, I organize an event for blind people this weekend so until next weekend I'll have no time.

JuliusSweetland:
@LauraLaureus I found some time to merge this - thanks! I ended up removing restart from the menu and leaving it only in the context menu as I wasn't convinced that it is required enough to shrink all the other menu keys by adding a new row. Hope that's acceptable.

Issue #504, opened 2018-10-05
Multi-key selection stays active when changing from normal to simplified keyboard


JuliusSweetland:
In a normal alpha keyboard enable multikey select (lock it down for ease) and then switch to a simplified keyboard - multi key selection will still be enabled and make typing very difficult.

Chopinsky:
Would you think it's sufficient if we simply toggle off the multi-selection lock if switching to a simplified keyboard? Or should it be better to guard the behavior behind a setting?

JuliusSweetland:
It would be sufficient if multi-key selection were released when switching
to a keyboard which doesn't support it

JuliusSweetland:
Included in 2.0.17

PR #514, opened 2018-10-24
Address Issue #504: Toggle Off Multikey Selection When Using Simplified Keyboard

Chopinsky:
As stated in issue #504 , the multikey selection feature will remain active if it's toggled in the alpha keyboard and then switching to the simplified keyboard. This behavior won't make sense because it's not supported in the simplified keyboard.

This PR will update this behavior: 
1. When multikey selection is on when switching to the simplified keyboard, its state will be reset (i.e. turned off), such that it's de facto toggled off when using the simplified keyboard. This patch will stand because there's no multikey selection on/off key on the simplified keyboard.
2. When switching back to the alpha keyboard, the multikey selection key will remain off regardless of its previous state (the state before switching to the simplified keyboard). Though the "remembered" key state won't be affected if user close the keyboard.


JuliusSweetland:
@Chopinsky Looks good, thanks!

Issue #520, opened 2018-11-05
Ukrainian localization


Larkailya:
Hey everyone. I with my friend @darinakulish want to make the Ukrainian localization for OptiKey. We are pleased that there is such a project and people care about people with disabilities. We would like to make the support of the Ukrainian language for this application, so that people in Ukraine can use this application.

We have sent a request for access to the Ukrainian language on transifex. My name on transifex **illialarka**, and friend's name **darinakul**.

We look forward to your permission, @JuliusSweetland.

JuliusSweetland:
Granted on Transifex. In addition to doing the translation we also need
customised keyboards for Ukrainian and to also possibly make any little
changes to handle Ukrainian specific behaviours that are needed. There's a
localisation document here:
https://github.com/OptiKey/OptiKey/blob/master/docs/Localize.md

Larkailya:
Hi, @JuliusSweetland !
We have finished translating into transifex everything that applies to the Ukrainian language.

> After completing the translation in Transifex, you should notify us as soon as possible so the translation you made can be checked and have its `.resx` file generated as you will need this in order to implement the new locale inside OptiKey. Having acquired your language `.resx` file it's time to start adding the new Locale inside OptiKey.

Can we continue and start integrating the language into the application?

JuliusSweetland:
Absolutely. You may be able to download the Ukrainian resx file yourself
and continue. If not let me know what I'll download and share it.

JuliusSweetland:
@Larkailya Released! v2.20.0

Thank you very much! Any more contributions from you would be very well received - your changes were very good.

PR #521, opened 2018-11-06
Issue #520 UA localization

Larkailya:
Hi everyone!
You have implemented the Ukrainian localization. We would be glad to see this locale in the project.


JuliusSweetland:
@Larkailya Excellent! Thank you. I will start reviewing this and release this as soon as I have completed some testing. If I have any changes or questions I will post them here. You should get an automatic email notifying you if I do this.

JuliusSweetland:
@Larkailya When testing I've noticed that the accent key (below) does not restrict which keys can be combined with the accent:
image

In Russian, for example, when you select that key it limits the letters on which the accent can be applied to these:
image

Should Ukrainian be the same? If so could you please add the logic? It lives in KeyEnabledStates.cs - here is the Russian equivalent: https://github.com/OptiKey/OptiKey/blob/c15f414ad744e455bda9b1198f7a7b265bb0750b/src/JuliusSweetland.OptiKey/Models/KeyEnabledStates.cs#L669

Once happy could you please update this PR, or close it and submit a new one?

Apart from that one thing I am happy to create a pre-release for you to test once you've made this change and then we can get this out ASAP. Thanks very much.

Larkailya:
Well, sorry, I missed it.

JuliusSweetland:
@Larkailya That looks better. Next thing I've found - I don't think the combination of the accent to the letters is working correctly. If you try typing the accent key and then the "a" key it will combine as expected, but I believe a different "a" is being used as the value of the key, resulting in the accent + the "a" key not combining into 1 character. You can see this if you debug this line: https://github.com/OptiKey/OptiKey/blob/master/src/JuliusSweetland.OptiKey/Services/KeyboardOutputService.cs#L496

I would expect the accent to combine (using Unicode composition) with each letter to form 1 character (the letter with an accent). What is currently happening is the accent + the letter is resulting in a combined character that is actually 2 characters long. This results in odd behaviour with the auto spaces - keep typing and a space will be inserted between the accented letter and the next letter - this is wrong.

I've tried changing the latin letters over to the ones from my keyboard (a,e,o) and it starts to work correctly, but I can't do this with the other Ukrainian letters. Could you take a look.

I would also like to check that you mean to include the double dotted i in the letters which can have an accent added to them.

Larkailya:
@JuliusSweetland Well, After debugging, I noticed that this bug is also reproducible on the Russian keyboard layout. In other languages that do not use Cyrillic characters, there is no such problem.
photo_2018-11-10_17-58-26

I could be wrong. But during debugging, I noticed that there is a boolean variable suppressNextAutoSpace which takes the value of truth if this condition is met

https://github.com/OptiKey/OptiKey/blob/1ed3aadca0060a700e7b133c36a294de2f967b51/src/JuliusSweetland.OptiKey/Services/KeyboardOutputService.cs#L541

In a language without Cyrillic characters, this condition is true, because the length of variable `newText.Length` is equal one character, but with Cyrillic, this condition is false (is equal two characters)

photo_2018-11-10_18-21-31

So the value of a boolean variable is not true.

Well, after a little research on this problem, I can offer this solution
```csharp
else if(new StringInfo(lastProcessedText).LengthInTextElements == 1
         && new StringInfo(newText).LengthInTextElements == 1
         && !lastProcessedTextWasSuggestion
         && !(keyStateService.KeyDownStates[KeyValues.MultiKeySelectionIsOnKey].Value.IsDownOrLockedDown() && char.IsLetter(newText.First())))
```
This solves the problem of adding space in languages using Cyrillic characters along with an accent.

I am unsure if this is the right solution. Ready for discussion.

JuliusSweetland:
@Larkailya That's a good fix. I would be interested to know if the underlying problem with the characters not combining could be fixed as well. I suspect there are multiple instances of the same letters in the Unicode page - some which combine with diacritic marks and some which don't. English is the same in that respect.

Issue #463, opened 2018-05-23
bi-color keyboard


Sheeva007:

Sheeva007:
@JuliusSweet beige lighter for Alpabetical keys and darker for function keys
Witch can render your keyboard clearer and more aesthetique keybord.

Dominique.

JuliusSweetland:
@PopeSpaceous Hi. Rather than using the SharedSizeGroup (which is an attached property from the Grid class and is only used with sizing), you can instead bind to the Key's HasSymbol property with something like this:

`Binding="{Binding Path=HasSymbol, RelativeSource={RelativeSource AncestorType={x:Type controls:Key}}}" `

I would make the change in the KeyBorderStyle (in a new theme that is designed to be two tone) and add new style setters to set the Background based on whether the key has a symbol or not. I think the disabled key logic should override the two tone behaviour, so define your logic above the existing disabled key logic.

Does that make sense?

PR #527, opened 2018-12-10
Fix issue #463. Added Two_Tone Keyboard. 

PopeSpaceous:
Fixes issue #463 
Added custom colour to to Colours/Android.xaml file. 
If you have a better colour suggestion I can change it. The suggested colour beige didn't look as nice with the white font. 

Suggested changes in the future:
Currently the suggested text is the same colour, as is the buttons for different languages. 

Let me know if there are any changes you would like me to make.


JuliusSweetland:
@PopeSpaceous Thanks for this. Could you make this new theme available in VisualsViewModel.cs Search for Android_Dark and you'll see where it's used. Also you'll need to add to Resources.resx with a string description of the theme.

JuliusSweetland:
@PopeSpaceous Merged. Thank you very much!

Issue #458, opened 2018-05-12
New function key to copy all of scratchpad


JuliusSweetland:
This would allow users to draft emails (for example) in the scratchpad, and then copy/paste them into an email client.

illialarka:
Happy New Year, @JuliusSweetland ! 🎁

PR #529, opened 2018-12-15
#458 Copy all text from scratchpad

Larkailya:
Hi,

I have implemented copying functionality. It is related with #458 issue.

I am looking forward your comment, @JuliusSweetland


JuliusSweetland:
@Larkailya Thank you. I've added a setting to control whether this key is turned on/off. It's controlled from the Visual settings tab.

Larkailya:
@JuliusSweetland , This is good decision. Thank you!

PR #535, opened 2019-01-28
Update ConversationAlpha1.xaml

JorgeMattos:
Adjust (controls: Output) Full all width.
The current one is an empty column.

spanish conversion


JuliusSweetland:
@JorgeMattos Thanks for finding and fixing this problem. Merged!

PR #543, opened 2019-02-26
Allows saving of special characters for key string value settings

### LINKED ISSUE #541: Adding a Setting for a string key with a special character (e.g. TAB) fails to save properly
# How OptiKey was being used

A setting was need to change the fixation time for the TAB key.

# What the user expected to happen

Adding the TAB key setting the settings file should stay valid after the settings is saved by the program, but because the XML saving uses writer.WriteString() instead of writer.WriteCData() the special characters are not saved.

# What the user actually experienced

Custom setting is lost.

# Version
 - Windows 10
 - OptiKey version v2.24.0 (current Master from GitHub)

# Steps to reproduce the issue

1. Add the following item to the KeySelectionTriggerFixationCompleteTimesByKeyValues dictionary in the settings file (note that there is a TAB character inside the CDATA):
```
                <item>
                    <key>
                        <keyValue>
                            <str>
                                <![CDATA[	]]>
                            </str>
                        </keyValue>
                    </key>
                    <value>
                        <ticks>PT2.75S</ticks>
                    </value>
                </item>
```
2. Open OptiKey, then close.
3. Examining the updated settings file shows that the CDATA is gone.

# Any other information that helps to describe and/or reproduce the problem

Problem seems resolvable by changing SerializableDictionaryOfTimeSpanByKeyValues.cs:102 from "writer.WriteString(key.String);" to "writer.WriteCData(key.String);".

wmacchi:
This request fixes #541.

### DIFF
--- a/src/JuliusSweetland.OptiKey/Models/SerializableDictionaryOfTimeSpanByKeyValues.cs
+++ b/src/JuliusSweetland.OptiKey/Models/SerializableDictionaryOfTimeSpanByKeyValues.cs
@@ -99,7 +99,7 @@ namespace JuliusSweetland.OptiKey.Models
                 if (key.String != null)
                 {
                     writer.WriteStartElement("str");
-                    writer.WriteString(key.String);
+                    writer.WriteCData(key.String);
                     writer.WriteEndElement();
                 }
                 writer.WriteEndElement();


JuliusSweetland:
@wmacchi Great stuff. Tested and merged. Thank you. I've also added the TAB key to the management console so you can set the fixation time in there instead of mucking about in xml.

wmacchi:
Fantastic, thank you Julius!

Issue #552, opened 2019-05-11
Dump out settings in log file (full xml if in Debug mode).


kmcnaught:
(useful for debugging).

PR #567, opened 2019-05-13
Log the user's settings.

kmcnaught:
For issue #552 

Can you please review whether:
`ConfigurationManager.OpenExeConfiguration(ConfigurationUserLevel.PerUserRoaming);`
is safe to use to get the users' settings file, and won't somehow be different for some systems/installations?

You might also have a nicer solution for reading/logging the file incrementally without the log4net preamble on every line - I couldn't find an easy way to temporarily disable the logging conversionPattern.

### DIFF
--- a/src/JuliusSweetland.OptiKey/App.xaml.cs
+++ b/src/JuliusSweetland.OptiKey/App.xaml.cs
@@ -278,7 +278,7 @@ namespace JuliusSweetland.OptiKey
 
                 foreach (SettingsProperty property in Settings.Default.Properties)
                 {
-                    Log.InfoFormat("  {0}: {1}", property.Name, Settings.Default[property.Name].ToString());
+                    Log.InfoFormat("  {0}: {1}", property.Name, Settings.Default[property.Name]);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


kmcnaught:
[OptiKey - NORMAL.log](https://github.com/OptiKey/OptiKey/files/3173652/OptiKey.-.NORMAL.log)
[OptiKey - DEBUG.log](https://github.com/OptiKey/OptiKey/files/3173653/OptiKey.-.DEBUG.log)

Here's an example of the logs that come out.

JuliusSweetland:
@kmcnaught I believe "ConfigurationManager.OpenExeConfiguration(ConfigurationUserLevel.PerUserRoaming);" is correct. If it is working then it looks correct from the documentation.

This line: "Settings.Default[property.Name].ToString()" - any chance that can throw a NullReferenceException if the property is null?

kmcnaught:
@JuliusSweetland If `property` is null, it would indeed throw a `System.NullReferenceException`. I'm not sure how you'd get in that state within the foreach, but it doesn't hurt to catch the exception if you prefer? Also technically I suppose `Settings.Default.Properties` _could_ be null, but I imagine that would cause many other problems in OptiKey...

JuliusSweetland:
Although I can't think of any it would be perfectly valid for a nullable
setting type to be null. Could you please add a null check before calling
to string, or use strong interpolation with the property itself. If it's
null then nothing will be logged. Thanks

PR #572, opened 2019-07-17
Finnish language support

lsaranto:
I had an accident with the Danish Conversation keyboard. Please verify that it's still ok. Sorry.

We write language names in lower case. That's why the first 'suomi' in 'suomi 
(Suomi)' is in lower case. Or should it be changed?


JuliusSweetland:
@lsaranto I'll check through this tonight if I can. From a quick check online it looks good. The Danish keyboard needs repairing, but that's easy enough to do. Thanks! I'll let you know once I've got it working.

JuliusSweetland:
@lsaranto I have merged in your changes (and fixed up a few minor things). Can you please pull the latest version of my Finnish branch and test it to make sure it behaves as you expect? If you're happy I'll create a pre-release installer for you to try. Everything looks good from a technical perspective, so thank you! Another language, which brings Optikey to 25!

Issue #581, opened 2019-08-15
Incorrect expand/collapse dock icons on mouse keyboard when docked to bottom on screen


annakirkpatrick:
This is a really minor issue, but hopefully will also be an easy fix.

Because of recent physical workspace set up changes, I've been experimenting with docking Optikey to the bottom of my screen (rather than the top, as has been my preference in the past).

With Optikey set up this way (docked to the bottom), the expand dock and collapse dock icons on the mouse keyboard appear to be drawn for a keyboard docked to the left side of the monitor. I know that when docked to the top, Optikey correctly displays icons reflecting the idea of expanding/collapsing from this location.

I will include some screenshots of the you can see the icons I'm referring to with the keyboard docked to the top and bottom of the screen.

![Docked to bottom](https://user-images.githubusercontent.com/33188271/63119924-3dc35f80-bf6f-11e9-87e2-954d42298a53.png)

![Docked to top](https://user-images.githubusercontent.com/33188271/63120007-5df31e80-bf6f-11e9-9a0d-86bc9aeca504.png)

If the correct icon already exists, this should be an easy fix, and I should be able to make the change myself. If not, then that's probably outside of my skill set.

kmcnaught:
Hi Anna, 

It looks like the symbols are the same, but with an orientation applied in this code:  
[src/JuliusSweetland.OptiKey.Core/UI/ValueConverters/MinimiseAndDockPositionToSymbolOrientation.cs](https://github.com/OptiKey/OptiKey/blob/master/src/JuliusSweetland.OptiKey.Core/UI/ValueConverters/MinimiseAndDockPositionToSymbolOrientation.cs)

You can see where the converter is used in the xaml here, e.g. line 680 (and elsewhere):
[src/JuliusSweetland.OptiKey.Core/UI/Views/Keyboards/Common/Mouse.xaml](https://github.com/OptiKey/OptiKey/blob/master/src/JuliusSweetland.OptiKey.Core/UI/Views/Keyboards/Common/Mouse.xaml)

I assume there's a bug in the MinimiseAndDockPositionToSymbolOrientation logic - do you want to investigate yourself?

JuliusSweetland:
@annakirkpatrick You could add logging to the converter to see which values are passed in:

https://github.com/OptiKey/OptiKey/blob/master/src/JuliusSweetland.OptiKey.Core/UI/ValueConverters/MinimiseAndDockPositionToSymbolOrientation.cs

[...]

PR #585, opened 2019-09-06
Symbol orientation fix for #581

kmcnaught:
The converter previously used was intended for minimising optikey, i.e. a combination of current dock position and minimised position (which may be different). The mouse keyboard's contract/expand functionality is only related to the dock position. In addition there was a bug in the wrong enums being passed into the converter. 

I've made a new converter just for dock position, which solves issue #581


JuliusSweetland:
@kmcnaught Should the symbol not flip depending on whether the mouse keyboard is minimised or not? I.e when it is minimised the symbol should show that we will expand away from the edge, but when maximimised that we will collapse towards the edge?

kmcnaught:
@JuliusSweetland that's what I initially assumed, but it turns out you've already encapsulated that in having an "ExpandDockIcon" and a "CollapseDockIcon" for the two different states (collapsed and full). With this PR the symbol does correctly show expanding away from the docked edge and collapsing toward it.

JuliusSweetland:
@kmcnaught Ok great, thanks.

PR #594, opened 2019-09-11
Magnifier Modifications

AdamRoden:
Modifications to give the user the option of having the magnifier show centered on the screen or around the cursor.


JuliusSweetland:
@AdamRoden Tested. All good. Merged. Thank you very much!

PR #597, opened 2019-09-30
Updade Alpha1.xaml Portuguese

JorgeMattos:
Magnifier already has in the mouse bar.
(,) and (.) Keep the Optikey pattern beside the space.
(-) very useful in Portuguese, indent.
(') Not used in Portuguese.
MULTI_KEY_SELECTION_SPLIT_WITH_NEWLINE There is no patient who uses here in Brazil, I gave unchecked in the console.


JuliusSweetland:
@JorgeMattos I have copied your changes to move , and ? and replace ' with -, but I don't think the other changes are a good idea. I would like to keep ALL languages consistent in their inclusion of various keys such as the Multi Key Selection key and the Web Browsing key. Although you may not know of any Multi Key Selection users, there may be users out there. I would prefer to move toward a model of users customising their own dynamic keyboards. I hope you understand.

PR #604, opened 2019-10-09
Enable color overrides on xml keyboards

AdamRoden:


JuliusSweetland:
@AdamRoden Merged, thank you. I've also added a few lines here: https://github.com/OptiKey/OptiKey/wiki/Creating-and-Using-Dynamic-Keyboards

PR #610, opened 2019-10-30
WindowManipulation improvements

AdamRoden:


JuliusSweetland:
@AdamRoden  Hi Adam,

The new PR looks good from a quick read, and I will test it tonight/tomorrow, but there are a few questions I have around how things work. Are you available to chat or call tomorrow? My questions:

1. IWindowManipulationService.OverrideKeyboard - does this need to be public? I can't see anything in this new PR that suggests that it is used outside of the WindowManipulationService?

2. The new win32 MoveWindow method - why is this used? This would be a departure from how windows are previously moved and sized, so I'm interested in what it achieves. Does this prevent flicker or something similar? This appears (from eyeballing the PR) to only be used when restoring OptiKey from a minimised state so I'm interested in why this is the only (?) use case?

3. Rather than "Override" and "RollbackOverride" when talking about window states, how about we refer to them as "TemporaryState"? This feels easier to understand for me as a temporary state would clearly not be persisted and would need to be rolled back.

4. Naming query again, but should this be called an override?
newDynKeyboard.OverrideKeyboardLayout(persistNewState, overrideWindowState, overridePosition, overrideDockSize, overrideWidth, overrideHeight, horizontalOffset, verticalOffset);

I almost feel like these are required on the constructor. Same with the key overrides. I can't see if these methods are called from anywhere else, which would require them to be methods, but I feel like the dynamic keyboard should be constructed with all of these things, assuming that they are then all immutable from that point onward?

Great PR. Looking forward to speaking to you about it.

Thanks,
Julius

AdamRoden:
@JuliusSweetland 

> IWindowManipulationService.OverrideKeyboard - does this need to be public?

No need for it to be public. In earlier iterations I had exposed it to use elsewhere, but that's not needed now.

> The new win32 MoveWindow method - why is this used? 

It prevents a form of flicker. Previously, with floating windows, there were situations where I observed the window repaint after the new position was set, again after the new height was set, and again after the width was set.

> Rather than "Override" and "RollbackOverride" when talking about window states, how about we refer to them as "TemporaryState"? This feels easier to understand for me as a temporary state would clearly not be persisted and would need to be rolled back.

I don't like Override either. What do you think about "DynamicPosition", "DynamicSize", or "DynamicState"?

> Naming query again, but should this be called an override?
newDynKeyboard.OverrideKeyboardLayout(persistNewState, overrideWindowState, overridePosition, overrideDockSize, overrideWidth, overrideHeight, horizontalOffset, verticalOffset);
I almost feel like these are required on the constructor. Same with the key overrides.

I'm not sure I follow what you mean exactly, but I also think it's a bit awkward the way it is coded now.

JuliusSweetland:
> No need for it to be public. In earlier iterations I had exposed it to use elsewhere, but that's not needed now.
Great, would you please revert to a private member var? Thx.

> It prevents a form of flicker. Previously, with floating windows, there were situations where I observed the window repaint after the new position was set, again after the new height was set, and again after the width was set.
Interesting. Good work finding that solution. Could you please double check that the key map is successfully rebuilt after restoring (just check that any key correctly recognises when it has focus). Could you also please add a comment as to why you are moving using the Win32 method so I, or anyone coming to that code in the future, know the reason?

> I don't like Override either. What do you think about "DynamicPosition", "DynamicSize", or "DynamicState"?
I'm not sure Dynamic is appropriate either as I don't believe there is a link between the fact that we are using a dynamic keyboard and the fact that we are temporarily overriding the window states. It could be that in the future a key triggers functionality to temporarily change the window state until another event occurs. In that made up example dynamic keyboards may not be being used at all. For me the main feature of the new state is that it is temporary. This conveys that it will be reversed, as well as the fact that it will not be persisted. Open to other suggestions though...

> I'm not sure I follow what you mean exactly, but I also think it's a bit awkward the way it is coded now.
I just mean that these two methods don't appear necessary:
DynamicKeyboard.OverrideKeyboardLayout
DynamicKeyboard.SetKeyOverrides
If they are never called anywhere else should we just add the parameters to the constructor so it is clear that a dynamic keyboard expects keyboard layout and key override info when it is created? Would this impact anywhere else?

PR #614, opened 2019-11-04
WIP DynamicKeyboards

JuliusSweetland:


src/JuliusSweetland.OptiKey.Core/Models/XmlKeyboardModels/XmlMultiStepKey.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) 2019 OPTIKEY LTD (UK company number 11854839) - All Rights Reserved
+using JuliusSweetland.OptiKey.Enums;
+using System.Collections.Generic;
+using System.Xml.Serialization;
+
+namespace JuliusSweetland.OptiKey.Models
+{
+    public class XmlMultiStepKey : XmlKey
```
AdamRoden: In order to free up the name "DynamicKey" for my new code, I renamed the old version to "MultiStepKey". It would be ideal to deprecate "MultiStepKey" when the new more advanced "DynamicKey" is released.

src/JuliusSweetland.OptiKey.Core/Models/XmlKeyboardModels/XmlKeys.cs
```diff
@@ -23,15 +23,15 @@ public List<XmlPluginKey> PluginKeys
         { get; set; }
 
         [XmlElement(ElementName = "DynamicKey")]
-        public List<XmlDynamicKey> DynamicKeys
+        public List<XmlMultiStepKey> MultiStepKeys
```
AdamRoden: In order to free up the name "DynamicKey" for my new code, I renamed the old version to "MultiStepKey". It would be ideal to deprecate "MultiStepKey" when the new more advanced "DynamicKey" is released.

src/JuliusSweetland.OptiKey.Core/Models/XmlKeyboardModels/XmlDynamicKey.cs
```diff
@@ -1,24 +1,22 @@
-﻿// Copyright (c) 2019 OPTIKEY LTD (UK company number 11854839) - All Rights Reserved
-using JuliusSweetland.OptiKey.Enums;
+﻿// Copyright (c) 2019 OPTIKEY LTD (UK company number 11854839) - AllRights Reserved
 using System.Collections.Generic;
 using System.Xml.Serialization;
 
 namespace JuliusSweetland.OptiKey.Models
 {
-    public class XmlDynamicKey : XmlKey
```
AdamRoden: In order to free up the name "DynamicKey" for my new code, I renamed the old version to "MultiStepKey". It would be ideal to deprecate "MultiStepKey" when the new more advanced "DynamicKey" is released.

src/JuliusSweetland.OptiKey.Core/Models/XmlKeyboardModels/XmlSuggestionRow.xaml
```diff
@@ -0,0 +1,103 @@
+﻿<!--
+Copyright (c) 2019 OPTIKEY LTD (UK company number 11854839) - All Rights Reserved
+-->
+<UserControl x:Class="JuliusSweetland.OptiKey.Models.XmlSuggestionRow"
```
AdamRoden: I have not been able to apply background overrides to suggestion keys and I still haven't figured out why.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2372,46 +2372,152 @@ private void SelectVoice(string voice)
...
-                foreach (var vStep in command.Split(stringSeparators, StringSplitOptions.None).ToList())
+                Log.InfoFormat("StepList called with singleKeyValue [{0}]", singleKeyValue.String);
+
+                List<string> commandList = singleKeyValue.String.Split('<').ToList();
+                if (commandList.Count > 0)
                 {
-                    Log.DebugFormat("Performing StepList step: {0}.", vStep);
-                    if (vStep.StartsWith("Action>") &&
-                        (Enum.TryParse<FunctionKeys>(vStep.Substring(7), out FunctionKeys result)))
+                    //old but works:
```
AdamRoden: Here is the synchronous processing of a key. If you want a key to press repetitively until released this is where you would get stuck and where we need a new thread.

The 'else' branch is where I go mess with timers and threads.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2372,46 +2372,152 @@ private void SelectVoice(string voice)
...
+                        keyValue = new ChangeKeyboardKeyValue(commandList.First().Substring(vIndex), vReturn);
+                    }
+                    KeySelectionResult(keyValue, multiKeySelection);
+                }
+                else if (commandList.First().StartsWith("KeyDown>"))
+                {
+                    keyValue = new KeyValue(commandList.First().Substring(8));
+                    int delayMs = 0;
+                    keyboardOutputService.ProcessSingleKeyPress(keyValue.String, KeyPressKeyValue.KeyPressType.Press, delayMs);
+                    if (keyStateService.KeyDownStates[singleKeyValue].Value != KeyDownStates.Down)
```
AdamRoden: I've found that the keyStateService can be out of sync from what Windows recognizes as the virtual state of a key.

src/JuliusSweetland.OptiKey.Core/Models/XmlKeyboardModels/XmlSuggestionRow.xaml
JuliusSweetland: Ok, we can come back to this

src/JuliusSweetland.OptiKey.Core/Models/XmlKeyboardModels/XmlMultiStepKey.cs
JuliusSweetland: Ok that sounds good to me.

JuliusSweetland:
@AdamRoden Try this: this code uses a timer which runs on a background thread, but I then use the main application dispatcher to run the handling code back on the UI thread. I don't know if this is going to work as we want, but it seems to solve the immediate issue of getting it to work:

`        //public static void Main()
        public void TimerStepList(Commands cmd)
        {
            //send the looping key commands to the TimerList
            TimerList.Add(cmd);
            var aTimer = new System.Threading.Timer(new TimerCallback(OnTimedEvent), null, 0, 100);
            //We'd want to store this timer against the key being held down so we can dispose of the timer when it is released
        }
        // Specify what you want to happen when the Elapsed event is raised.
        private void OnTimedEvent(object source)
        {
            //check for any new commands
            if (TimerList.Any())
            {
                Application.Current.Dispatcher.Invoke(() =>
                {
                    //handle the commands
                    StepListCommands(TimerList.First().SingleKeyValue, TimerList.First().MultiKeySelection, TimerList.First().CommandList);
                    //remove the commands
                    //TimerList.RemoveAt(0);
                });
            }            
        }`

AdamRoden:
It works, but it doesn't do what we want. The UI gets locked the same as if it was all one thread. I achieved similar results when I tried this alternate approach.

        private void Start(object info)
        {
            Commands cmd = (Commands)info;
            StepListCommands(cmd.SingleKeyValue, cmd.MultiKeySelection, cmd.CommandList);
        }
        private void TimerStepList(Commands cmd)
        {
            ParameterizedThreadStart start = new ParameterizedThreadStart(Start);
            Thread xy = new Thread(start) { IsBackground = true };
            xy.Start(cmd);
        }

JuliusSweetland:
@AdamRoden What are you doing on the UI thread when it fires that could be locking the UI thread? My understanding of the code is that the timer will trigger every 100ms and dispatch back to the UI thread to fire some code. If that code is long running then the UI thread will be locked up. Are there pauses or waits in the handling of that key?

AdamRoden:
@JuliusSweetland 
After firing my long running key, I would expect the keyboard to follow my gaze and to be able to activate another key while the long running key is processing. But what I get is an unresponsive keyboard until the long running key is done.

JuliusSweetland:
@AdamRoden Can you point me towards how you achieve the pauses? I can see the Wait actions, but I couldn't see how you actually pause the thread.

AdamRoden:
@JuliusSweetland 
> @AdamRoden Can you point me towards how you achieve the pauses? I can see the Wait actions, but I couldn't see how you actually pause the thread.

You mean the Wait/Pause/Sleep:
>                else if (commandList.First().StartsWith("Wait>"))
>                {
>                   System.Threading.Thread.Sleep(int.Parse(commandList.First().Substring(5)));
>                }

JuliusSweetland:
Ok that'll be the issue. If we're on the UI thread and we Thread.Sleep then
the whole thread is suspended and googling will respond. I'll look at how
to achieve the same delays using non blocking methods.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2372,46 +2372,152 @@ private void SelectVoice(string voice)
...
+        private void StepListCommands(KeyValue singleKeyValue, List<string> multiKeySelection, List<string> commandList)
+        {
+            Log.InfoFormat("StepListCommands called with a list of {0} commands", commandList.Count);
+
+            KeyValue keyValue;
+            while (commandList.Count > 0)
+            {
+                Log.InfoFormat("Performing StepList step: {0}.", commandList.First());
+
+                if (commandList.First().StartsWith("Loop>"))
```
AdamRoden: Fixed some code:
>if (commandList.First().StartsWith("Loop>"))
                {
                    var count = Int32.Parse(commandList.First().Substring(5));
                    //if this loop has a nested loop then set the end of the loopList to that of the last end loop command
                    var vEnd = (commandList.FindIndex(1, x => x.StartsWith("Loop>")) > 0 
                        && commandList.FindIndex(1, x => x.StartsWith("Loop>")) < commandList.IndexOf("/Loop>"))
                        ? commandList.LastIndexOf("/Loop>") - 1
                        : commandList.IndexOf("/Loop>") - 1;                    
                    for (int i = 0; i < count; i++)
                    {
                        var loopList = commandList.GetRange(1, vEnd);
                        StepListCommands(singleKeyValue, multiKeySelection, loopList);
                    }
                    commandList.RemoveRange(0, vEnd);
                }

JuliusSweetland:
@AdamRoden What happens if you comment out the sleeps completely? However short they will be noticeable.

I'm thinking there are two ways to get around blocking the UI thread. If we assume that the main problem is the pausing (Thread.Sleep), and that the key handling code could be called from either the UI thread, or from a timer executing the key handling code on a threadpool thread, then we have two options:

1. Move the code which dispatches back to the UI thread so it only occurs for actions other than thread.sleep. This should be achievable, BUT does not help us if we are handling a key press on the UI thread which includes waits. This only solves the problem if we start handling from a timer event on a threadpool thread.

2.Make the whole method chain async so we can use a combination of:

a) Dispatching back to the UI thread for handling of a key press. This invoking would also be async. See the (second suggested solution) in the main answer here: https://stackoverflow.com/questions/23442543/using-async-await-with-dispatcher-begininvoke

b) Use await Task.Delay (also shown in that stackoverflow answer) instead of Thread.Sleep.

Solution 2 I think is the correct one as it allows us to use Task.Delay. If you await that you don't block the current thread, instead async/await works by building a state machine around the awaited code and using interrupts to return control back to the code when the Task completes. 

N.B. If we make the key handling awaitable then the first time you await something the thread is released to continue doing what it was doing before. I think this is fine here as the entry point is an event handler in the input service. What would happen is that the UI thread would be released to continue processing messages, and when the Task.Delay completes the UI thread would continue processing from after the Task.Delay. This asynchronous behaviour should give the impression of multiple things happening concurrently even though everything (with the exception of keys which kick off timer runs) happens on the UI thread.

Does that help?

AdamRoden:
@JuliusSweetland I'll attempt option 2.

AdamRoden:
@JuliusSweetland  I tried a few variations of async and have still not had any luck.

>        App.Current.Dispatcher.Invoke(async () => StepList(singleKeyValue, multiKeySelection));
>        ...

        public async Task StepList(KeyValue singleKeyValue, List<string> multiKeySelection)
        {
            await Task.Delay(100);
            if (!string.IsNullOrEmpty(singleKeyValue.String))
            {
                Log.InfoFormat("StepList called with singleKeyValue [{0}]", singleKeyValue.String);

                List<string> commandList = singleKeyValue.String.Split('<').ToList();
                if (commandList.Count > 0)
                {
                    //old but works:
                    if (false)
                        StepListCommands(singleKeyValue, multiKeySelection, commandList);
                    //new but not ready:
                    else
                    {
                        Commands cmd = new Commands
                        {
                            SingleKeyValue = singleKeyValue,
                            MultiKeySelection = multiKeySelection,
                            CommandList = commandList
                        };
                        TimerList.Add(cmd);
                        Application.Current.Dispatcher.InvokeAsync<Task<int>>(TimerStepList);
                    }
                }
            }
        }

        public async Task<int> TimerStepList()
        {
            await Task.Delay(100);
            //send the looping key commands to the TimerList
            StepListCommands(TimerList.First().SingleKeyValue, TimerList.First().MultiKeySelection, TimerList.First().CommandList);
            //remove the commands
            TimerList.RemoveAt(0);
            return 42;
        }

JuliusSweetland:
@AdamRoden I'm guessing the TimerStepList is not pausing as expected - this is probably because the call to Application.Current.Dispatcher.InvokeAsync<Task<int>>(TimerStepList); is not awaited.

Another approach would be to go all the way back up the call chain to here and make this an async (void) event handler: https://github.com/OptiKey/OptiKey/blob/18ab4c5c5552dafbbca52afa613c24dc57f3ecb1/src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs#L118

If you do that you can make each method down to StepList an "async Task" method and then you wouldn't have to make the below async:

`
App.Current.Dispatcher.Invoke(async () => StepList(singleKeyValue, multiKeySelection));
`

Instead you'd be in an async method already so you could just await a call to:

`
Application.Current.Dispatcher.InvokeAsync<Task<int>>(TimerStepList);
`

Looking at the code again I also don't think you want to be dispatching for the call to TimerStepList, but instead for the call in OnTimedEvent to StepListCommands. The idea is that from the top of the callstack (InputService) we are on the UI thread. This only changes in TimerStepList when a timer is created and the OnTimedEvent method is called from a ThreadPool thread. As this called back to StepListCommands (which eventually might do stuff where we need to be back on the UI thread) then we need to dispatch from OnTimedEvent back to the UI thread. Does that make sense?

Can we clarify whether you are using Timers to simulate a looping keypress, Task.Delays to allow a key to achieve a pause, both, or are you using Task.Delays to simulate a repeating key by delaying within a loop (until some condition is satisfied)? If a key can loop then we need to be clear about how the loop is completed. A timer makes this easy as we can associated the timer with the down key value and dispose the timer when the keyvalue is released. However, a timer requires us to dispatch back to the main thread.

Let me know what you are trying to achieve in terms of looping and pauses and we can work it out.

N.B. I am suggesting code based only on reading. If it doesn't work can you push your latest version to your repo and replace the existing PR. I will then pick it up and try to work it out.

AdamRoden:
@JuliusSweetland 
By starting with async here:
>inputServiceSelectionResultHandler = async (o, tuple) =>

and working down level by level, I was able to get an unstable version of multi-threading running. I was able to have a key loop running on one thread, and to simultaneously type letters on another thread. However, after a couple keystrokes on thread 2, thread 1 would stop without completing. And more troubling is the persisted window state gets lost, which I take as an indication that my threading screwed something up.
I'll pick it up again tomorrow.

JuliusSweetland:
Progress! I would add logging all through the key handler code, including
when the handler finishes. Then repeat the test and check the logs. You
will see after the timestamp a number in square brackets - this is the
thread number. 1 is typically the main UI thread:

[image: image.png]

This will allow us to figure out what is happening on which thread and what
is happening. If a thread is not completing there is a chance an exception
is being thrown, although I would expect the unhandled exception handler
defined in App.Xaml.cs to catch and log this and for Optikey to crash.

Let's look into the logs when you have them...

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2372,47 +2380,117 @@ private void SelectVoice(string voice)
...
+        public void StepList(KeyValue singleKeyValue, List<string> multiKeySelection)
+        {
+            if (!string.IsNullOrEmpty(singleKeyValue.String))
+            {
+                Log.InfoFormat("StepList called with singleKeyValue [{0}]", singleKeyValue.String);
+
+                List<string> commandList = singleKeyValue.String.Split('<').ToList();
+                if (commandList.Any())
+                {
+                    var cTask = Task.Run(() => StepListCommands(singleKeyValue, multiKeySelection, commandList));
```
AdamRoden: @JuliusSweetland I got it working so that I can start the threaded task at this point.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2372,47 +2380,117 @@ private void SelectVoice(string voice)
...
+                }
+            }
+        }
+
+        private async Task<int> StepListCommands(KeyValue singleKeyValue, List<string> multiKeySelection, List<string> commandList)
```
AdamRoden: @JuliusSweetland I made this method into an async task.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2372,47 +2380,117 @@ private void SelectVoice(string voice)
...
         {
-            if (!string.IsNullOrEmpty(command))
+            Log.InfoFormat("StepListCommands called with a list of {0} commands", commandList.Count);
+            if (commandList.Contains("Wait>"))
             {
-                KeyValue singleKeyValue;
-                string[] stringSeparators = new string[] { "<" };
-                foreach (var vStep in command.Split(stringSeparators, StringSplitOptions.None).ToList())
+                //if there is a wait command then make this task async by using await
+                await Task.Delay(10);
```
AdamRoden: @JuliusSweetland I only break into async if there is a "Wait" command associated with the key. Worth noting, regardless of this being sync or async, it is always on it's own thread and I haven't traced things down yet, but it seems there is interference with callbacks.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
JuliusSweetland: @AdamRoden If you do this you are spinning up a new threadpool thread to run the StepListCommands method on. You will then have to dispatch back to the UI thread to perform any action on a key. I would think it better to keep everything on the UI thread, i.e. make the whole method chain async and then correctly await each call down the chain.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
JuliusSweetland: @AdamRoden The problem is that nothing is awaiting that returned task:

Task.Run(() => StepListCommands(singleKeyValue, multiKeySelection, commandList));

^^^ that doesn't await an async method, it merely runs it on a threadpool thread.

JuliusSweetland: You COULD do a Task.Run and then call .Wait on the StepListCommands() call, i.e. StepListCommands(parameters).Wait(), but I still think that running the step list on a background thread is not necessary. Starting a new thread "costs" about 1MB of memory (not 2 as I previously said) and you will later have to dispatch back to the UI thread anyway.

I'll email you my version of the class which is async/await all the way down to the StepList method. Can you compare and see if you can carry on the async/await pattern into the delays etc?

JuliusSweetland:
@AdamRoden

Issue #599, opened 2019-10-01
Holding down a key in a Dynamic Keyboard 


tqphan:
I would like to create a dynamic keyboard where the user can hold down a key. Basically its behavior would be exactly how the modifier keys (shift, alt, ctrl) work. The reason for this is to play games. Many games require keys to be held down. Would this be possible without modifying the source code?

AdamRoden:
@tqphan I don't know of any way to do what you are asking without modifying the code. However, the modifications could get tied in with any efforts toward #595. I'd like to have more control regarding key-down and key-up behavior, and your request makes it more likely something will be done.

JuliusSweetland:
@tqphan @AdamRoden This might be possible by adding a new publicly settable properties for keys which can be pressed down and keys which can be locked down:
https://github.com/OptiKey/OptiKey/blob/master//src/JuliusSweetland.OptiKey.Core/Models/KeyValues.cs#L412:38

A dynamic keyboard could add keys to these properties when it is loaded, as long as exiting the keyboard handles the transition to another keyboard by checking the destination keyboard for keys which can be pressed and locked down and ensuring that the properties are updated and releasing keys which can no longer be held down.

AdamRoden:
@tqphan @JuliusSweetland I'm thinking about how it would make sense to switch a key into locked down mode and have two ideas for you to consider.
a. If a key has locking enabled, then it can be locked down by staring at it for a specified amount of time.
b. We add a new key, that when pressed tells OptiKey to press down the next key, but do not release it. 

In either case the key would get released by changing keyboard or activating it again.

I think option 'a' is the fancy way and option 'b' is the easy to understand and implement way.

tqphan:
@AdamRoden @JuliusSweetland  Games often requires holding down keys and maybe at the same time spamming keys. For example, holding down 'W' to move forward while spamming 'E' to swing an axe. The drawback for option B is that you'll have to constantly pressing the 'locking enabled' key to switch between the holding and spamming behavior.

JuliusSweetland:
@AdamRoden @tqphan How about a LOCK key which itself can be locked down. The behaviour would be:
1.If LOCK is pressed down then the next key pressed is locked down and LOCK is then released.
2.If LOCK is locked down then multiple other keys can be locked down until LOCK is released manually.

A similar idea could be implemented for a RAPID or TURBO key which presses and releases a key, or set of keys every x milliseconds (based on a setting).

Thoughts?

[...]

PR #615, opened 2019-11-07
Adam roden master

JuliusSweetland:


JuliusSweetland:
@AdamRoden This is the new PR, FYI.

JuliusSweetland:
@AdamRoden This is only temporary so you can get the code back into your fork of the project. Can you check, test, etc and see if it's good. I can get the long running key to run and press other keys at the same time.

I mentioned in my email, but I don't understand the point of keyboardOutputService.ProcessSingleKeyPress and it duplicates code. Can we discuss this with a view to removing this?

AdamRoden:
@JuliusSweetland 
keyboardOutputService.ProcessSingleKeyPress was an unimplemented stub I found that seemed intended for what I needed to do. The difference from existing code is that you explicitly tell it what key state to send the input simulator. I thought this would be preferential for special key up and key down actions and I was experimenting and didn't want to break core code, so I branched. What you commented out is needed to handle new KeyUp, KeyDown, and KeyToggle actions that I was using in my DemoMove keyboard. That doesn't mean it's set in stone, it's just an explanation of what I did.

JuliusSweetland:
Understood. My concern is that you could send nonsense signals to Windows,
e.g. pressing down a key multiple times without releasing, which would not
be interpreted correctly by other applications. Could we remove this for
now?

AdamRoden:
@JuliusSweetland The key up/down actions were for issue #599. Can we be more specific about what to remove/change and how it relates to #599?

JuliusSweetland:
> 
> 
> @JuliusSweetland The key up/down actions were for issue #599. Can we be more specific about what to remove/change and how it relates to #599?

Well there is already handling for <KeyDown> and <KeyUp> which call keyStateService.KeyDownStates[commandKeyValue].Value = KeyDownStates.Down; and keyStateService.KeyDownStates[commandKeyValue].Value = KeyDownStates.Up;

That achieves exactly the same thing.

I would keep that handling and remove the keyboardOutputService.ProcessSingleKeyPress method and (commented out) calls to it.

The outstanding issue is then the fact that you would need two keys: one to press a key down and one to release it, which doesn't make much sense as 2 keys as the actions should be mutually exclusive. Instead I think a dynamic key should have a progression of states which are defined in the xml. Something like MultiStateDynamicKey, which cycles through the states. A typical example might be a key initially in a state where pressing it triggers a KeyDown. The key then changes to the next state where the background is changed (or similar) and the next press triggers a KeyUp.

We would have to think about how KeyDown is enhanced so you can also specify that when the key is down the action (or list of actions) is repeated every x milliseconds, which would give us the repeated keystroke behaviour you want. If this were achieved using a timer (as previously discussed) then we could store the timer in a dictionary keyed by the keyvalue and set a listener on that key state so that when the key is released (by the next key press, or whatever event releases the key) then we can automatically dispose of the timer and stop the repeating keystroke behaviour.

AdamRoden:
@JuliusSweetland That makes sense. I'll can update code and test more today. I pushed up my code (not how you requested yet), but fully visible again.
My idea for a repeated keystroke was to use the (loop, count, keydown, and wait) commands I built into the new XmlKeyboard, and have a count of zero be an indication to loop indefinitely. If that seems confusing, another new key action type.

JuliusSweetland:
> 
> 
> @JuliusSweetland That makes sense. I'll can update code and test more today. I pushed up my code (not how you requested yet), but fully visible again.
> My idea for a repeated keystroke was to use the (loop, count, keydown, and wait) commands I built into the new XmlKeyboard, and have a count of zero be an indication to loop indefinitely. If that seems confusing, another new key action type.

I was going to reply and say that an infinite loop sounds dangerous, as I couldn't think of a nice way to stop the loop, but actually what we could do is:
a)Validate that there is only 1 infinite loop within a key, or a key's current state if you implement the concept of multi state keys. This doesn't have to be the top level loop as it might be useful to be able to do some optional set up actions (including loops) then have a main infinite loop, and some optional exit actions (including loops).

b)When a dynamic key is pressed the first thing we should do is check if the KeyValue is currently infinitely looping and stop it. This gives us the workflow that to start a looping key you select it, and then select it again to stop it. I think we could do this by maintaining (in the KeyStateService) a HashSet of looping keys by KeyValue. When we start an infinite loop then it goes in there and the loop only continues while it exists in there - something like:

while(keyStateService.LoopingKeyValues.Contains(keyValue))
{
//do looping stuff
}

When that key is selected again the first thing it would do is while(keyStateService.LoopingKeyValues.Remove(keyValue) to cause the previous loop to end.

c)If a key (or the current key state) only contains one Action or Text then that should be the generated Key's value. This way the key will correctly respond to key down/up state changes by animating and changing the background colour. In this way the key could (hopefully) blink as it is automatically pressed down/up in a loop, or have a coloured background if it is being held down.

I am not currently sure how key states would work in terms of changing the actual Key used in the keyboard as the handling of the input selection and the visual Key class are not associated with each other, and shouldn't be (visual and logical separation in WPF is a core tenet). We'll have to think about this. I think if we can correctly identify that a dynamic key only has one Action or Text value and then initially create the Key with that value then it should correctly represent key up/down states via the theme.

What do you think?

JuliusSweetland:
@AdamRoden Regarding the multi-state dynamic key - this concept could (I think) only be used to change the behaviour of the next selection of the key, with the obvious usecase being state1=KeyDown, state2=KeyUp. The visual styling of the key would be the same, unless we can think of an MVVM compatible way to achieve this.

AdamRoden:
@JuliusSweetland My head is spinning. I'm back to thinking I had things right when I was using keyboardOutputService.ProcessSingleKeyPress. I can cut down duplicate code, but I'm not seeing how key toggling would work cleanly without passing in a parameter to specify whether to toggle the key, or press and release. 
Everything else you laid out is very much in line with what I had envisioned.

JuliusSweetland:
If a dynamic key has text or a function key which maps to a key (e.g.
enter) then the current behaviour is for OptiKey to press and release the
key. If the key can be pressed or locked down you get that behaviour
instead, eg ctrl.

How about we remove SingleKeyPress, key down, and key up (unless we need
them?). That just leaves infinite looping which we can achieve by defining
an infinite loop and then storing the key value in a hash set which
contains currently looping keys, as per my previous message.

Does that cover all the use cases you had in mind?

AdamRoden:
@JuliusSweetland I concur. The default behavior for Dynamic keys is PressAndRelease and the existing Ctrl, Alt, Shift, Win logic works the same way as with traditional keys. The addition of 3 new behaviors: KeyUp, KeyDown, and KeyToggle is not needed; all we need is one new behavior KeyToggle used in conjunction with KeyStateService. 

I thought I had KeyIsDownColorOverride working on DynamicKeys, but something got messed up. It works on the Dynamic Sleep key, but not arrows. I'll have to revisit that.

Yes, we need to have something to track KeyValues that are looping and with that, we also need a ReleaseAll action.

JuliusSweetland:
What will be the use case for KeyToggle?

AdamRoden:
By KeyToggle I mean perform the next selection action: if up then press, if down then release.

JuliusSweetland:
Ah understood, yes that is useful.

JuliusSweetland:
Just to be clear though, toggling a key to down is not what should action
an infinite loop to fire up, down, up, down, etc. That would be achieved in
a loop which contains KeyDown,Wait,KeyUp,Wait. Key toggle is so you can
hold down keys.

Is that your understanding?

AdamRoden:
All this back and forth and I bet you could have finished the code by now. If you want to write code, and have me test it, I'm fine with that. Otherwise, all the chatter trying to clarify things might be doing just as much muddying of the water. I think we are 90% on the same page, just saying things different ways.

So, from where we are at do you want to send some new code my way, or wait on me to send some code to you?

JuliusSweetland:
@AdamRoden I'm very happy for you to write it - the innovation is all yours. I think we are now on the same page. We will end up with dynamic keys which can produce:

Standard text/functionkey keys which simulate a key down/key up
Keys which can action the pressing and holding down of a key with text or action
Keys which can trigger a sequence of events
Keys which can trigger one or more events, but with an infinite loop which will be terminated when the key is pressed again

Plus all the extra ability to override background colour/opacity and selection colour/opacity

Does that sound correct?

AdamRoden:
That sounds correct. 
A couple comments for the record so I don't forget again... 
I went back to some of my old code and old research. 
First, with a physical keyboard and key repeats when you hold down a key, it only sends the KeyDown message repeatedly. There is only one KeyUp message at the end when you release the key. When I initially set out, it was to model that behavior. 
Second, I did have KeyDown color overrides working, but it is dependent on sending the full KeyValue to the keystateservice. (Not just the command from from the Dynamic key.) I think we might be able to use the same/similar idea for looping keys.
Third, maybe we need to send both the full KeyValue, and the individual virtual key codes to the keystateservice.

JuliusSweetland:
Ok that all sounds good. We'd need to check if the anything actually
happens (including whether the selection animation works) if a key state is
repeatedly set to down. I suspect after the first key down nothing will
happen as the state isn't actually changing.

I understand that a physical keyboards sends a key down as a series of only
key down signals, but I think it would be equivalent to send key up, then
key down signals. I can't think of an example where holding down a key
differs from pressing and releasing a key over and over. Unless you know of
one I suggest we simulate repeated key down/up and look to simulate a
sequence of key downs is the need for it comes to light later.

Agree?

PR #616, opened 2019-11-10
Key state and looping files for review

AdamRoden:
@JuliusSweetland I rebased against v2.32.0 and made some tweaks to my code to get color overrides and key up/down states behaving to my expectations. (If we change it going forward that's fine, but for now it is working in a way that makes sense to me.) 
Otherwise, looping and long running keys should be unchanged from what you last saw, and I need to pull in your most recent code additions for error handling

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
@@ -493,7 +493,7 @@ namespace JuliusSweetland.OptiKey.Services
 
         public async Task ProcessSingleKeyPress(string inKey, KeyPressKeyValue.KeyPressType type)
         {
-            Log.InfoFormat("Processing input value {0} with key press type {1}", inKey, type);
+            Log.InfoFormat("ProcessSingleKeyPress called for key [{0}] press type [{1}]", inKey, type);
 
             FunctionKeys funKey;
             var virtualKeyCode = (Enum.TryParse(inKey, out funKey)) ? funKey.ToVirtualKeyCode() : null;
--- a/src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Text;
+using System.Threading.Tasks;
 using System.Windows.Forms;
 using WindowsInput.Native;
 using JuliusSweetland.OptiKey.Enums;
@@ -66,7 +67,7 @@ namespace JuliusSweetland.OptiKey.Services
         #endregion
 
         #region Properties
-        
+
         public string Text
         {
             get { return text; }
@@ -142,7 +143,7 @@ namespace JuliusSweetland.OptiKey.Services
                         {
                             backOneCount = Text.Length; //Coallesce backCount if somehow the Text length is less
                         }
-                        
+
                         if (backOneCount == 1)
                         {
                             var inProgressWord = Text.InProgressWord(Text.Length);
@@ -477,7 +478,7 @@ namespace JuliusSweetland.OptiKey.Services
 
             var capturedTextAfterComposition = CombineStringWithActiveDeadKeys(capturedText);
             ProcessText(capturedTextAfterComposition, true);
-            
+
             lastProcessedTextWasSuggestion = false;
 
             //Special handling for simplified keyboards
@@ -490,12 +491,62 @@ namespace JuliusSweetland.OptiKey.Services
             }
         }
 
-        public void ProcessSingleKeyPress(string key, KeyPressKeyValue.KeyPressType type, int delayMs = 0)
+        public async Task ProcessSingleKeyPress(string inKey, KeyPressKeyValue.KeyPressType type)
         {
-            // TODO: This is a stub for future use
-            throw new NotImplementedException();            
+            Log.InfoFormat("Processing input value {0} with key press type {1}", inKey, type);
+
+            FunctionKeys funKey;
+            var virtualKeyCode = (Enum.TryParse(inKey, out funKey)) ? funKey.ToVirtualKeyCode() : null;
+
+            if (virtualKeyCode != null)
+            {
+                if (type == KeyPressKeyValue.KeyPressType.Press)
+                    publishService.KeyDown(virtualKeyCode.Value);
+                else if (type == KeyPressKeyValue.KeyPressType.Release)
+                    publishService.KeyUp(virtualKeyCode.Value);
+                else
+                    publishService.KeyDownUp(virtualKeyCode.Value);
+            }
+            else
+            {
+                //Get keyboard layout of currently focused window
+                IntPtr hWnd = PInvoke.GetForegroundWindow();
+                int lpdwProcessId;
+                int winThreadProcId = PInvoke.GetWindowThreadProcessId(hWnd, out lpdwProcessId);
+                IntPtr keyboardLayout = PInvoke.GetKeyboardLayout(winThreadProcId);
+
+                //Convert this into a culture string for logging
+                string keyboardCulture = "Unknown";
+                var installedInputLanguages = InputLanguage.InstalledInputLanguages;
+                for (int i = 0; i < installedInputLanguages.Count; i++)
+                {
+                    if (keyboardLayout == installedInputLanguages[i].Handle)
+                    {
+                        keyboardCulture = installedInputLanguages[i].Culture.DisplayName;
+                        break;
+                    }
+                }
+                foreach (var chaKey in inKey)
+                {
+                    var vkKeyScan = PInvoke.VkKeyScanEx(chaKey, keyboardLayout);
+                    var vkCode = vkKeyScan & 0xff;
+                    if (vkKeyScan != -1)
+                    {
+                        if (type == KeyPressKeyValue.KeyPressType.Press)
+                            publishService.KeyDown((VirtualKeyCode)vkCode);
+                        else if (type == KeyPressKeyValue.KeyPressType.Release)
+                            publishService.KeyUp((VirtualKeyCode)vkCode);
+                        else
+                            publishService.KeyDownUp((VirtualKeyCode)vkCode);
+                    }
+                    else
+                    {
+                        Log.InfoFormat("No virtual key code found for '{0}'", chaKey);
+                    }
+                }
+            }
         }
-        
+
         private string CombineStringWithActiveDeadKeys(string input)
         {
             Log.InfoFormat("Combing dead keys (diacritics) on '{0}'", input);
@@ -548,11 +599,11 @@ namespace JuliusSweetland.OptiKey.Services
                 Log.Debug("Suppressing auto space before this capture as the last text change was null or white space.");
                 suppressNextAutoSpace = true;
             }
-			else if(!Settings.Default.KeyboardAndDictionaryLanguage.SupportsAutoSpace()) //Language does not support auto space
-			{
-				Log.DebugFormat("Suppressing auto space before this capture as the KeyboardAndDictionaryLanguage {0} does not support auto space.", Settings.Default.KeyboardAndDictionaryLanguage);
+            else if(!Settings.Default.KeyboardAndDictionaryLanguage.SupportsAutoSpace()) //Language does not support auto space
+            {
+                Log.DebugFormat("Suppressing auto space before this capture as the KeyboardAndDictionaryLanguage {0} does not support auto space.", Settings.Default.KeyboardAndDictionaryLanguage);
                 suppressNextAutoSpace = true;
-			}
+            }
             else if (new StringInfo(lastProcessedText).LengthInTextElements == 1
                      && new StringInfo(newText).LengthInTextElements == 1
                      && !lastProcessedTextWasSuggestion
@@ -640,7 +691,7 @@ namespace JuliusSweetland.OptiKey.Services
                         }
 
                         //Remove (from the end of the string) the part of the in-progress word which will be changed by composition
-                        Text = Text.Substring(0, Text.Length - countOfCharactersToRemove); 
+                        Text = Text.Substring(0, Text.Length - countOfCharactersToRemove);
 
                         //Remove changed in-progress word suffix string from external applications by outputting backspaces - the new suffix will be published seperately
                         for (var backCount = 0; backCount < countOfCharactersToRemove; backCount++) //Don't include newTextProcessed as it does not exist on Text yet
--- a/src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
+++ b/src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
@@ -160,14 +160,15 @@ namespace JuliusSweetland.OptiKey.UI.ViewModels
                                 b
                             */
 
-                            //if the key is in a running state and gets pressed, then stop it and change the state to up
+                            //if the key is in a running state and gets pressed, then stop it
                             if (keyStateService.KeyRunningStates[singleKeyValue].Value)
                             {
+                                Log.InfoFormat("StepList key triggerred while in a running state. Ending key: {0}", singleKeyValue.String);
                                 keyStateService.KeyRunningStates[singleKeyValue].Value = false;
-                                keyStateService.KeyDownStates[singleKeyValue].Value = KeyDownStates.Up;
                             }
                             else
                             {
+                                Log.InfoFormat("Starting StepList key: {0}", singleKeyValue.String);
                                 await StepList(singleKeyValue, multiKeySelection);
                             }
                         }
@@ -2430,51 +2431,78 @@ namespace JuliusSweetland.OptiKey.UI.ViewModels
         {
             if (!string.IsNullOrEmpty(singleKeyValue.String))
             {
-                Log.InfoFormat("StepList called with singleKeyValue [{0}]", singleKeyValue.String);
+                Log.InfoFormat("StepList called with singleKeyValue: {0}", singleKeyValue.String);
 
-                List<string> commandList = singleKeyValue.String.Split('<').ToList();
+                List<string> commandList = singleKeyValue.String.Substring(11).Split('<').ToList();
                 if (commandList.Any())
                 {
+                    var keyDownList = new List<KeyValue>();
+                    keyStateService.KeyRunningStates[singleKeyValue].Value = true;
+                    keyStateService.KeyDownStates[singleKeyValue].Value = KeyDownStates.LockedDown;
+
+                    //populate keyDownList with the keys having potential to be pressed and not released
+                    foreach (var downKey in commandList.FindAll(x => (x.StartsWith("KeyToggle>") || x.StartsWith("KeyDown>"))).Distinct())
+                    {
+                        var keyUpCandidate = new KeyValue(downKey.Substring(downKey.IndexOf(">") + 1));
+                        if (keyStateService.KeyDownStates[keyUpCandidate].Value == KeyDownStates.Up)
+                            keyDownList.Add(keyUpCandidate);
+                    }
+
                     await StepListCommands(singleKeyValue, multiKeySelection, commandList, 0);
+
+                    //if the StepList was stopped by an external process and there are any keys
+                    //that were pressed and not released then execute KeyUp processing
+                    if (!keyStateService.KeyRunningStates[singleKeyValue].Value && keyDownList != null && keyDownList.Any())
+                    {
+                        foreach (var keyUpCandidate in keyDownList)
+                        {
+                            if (keyStateService.KeyDownStates[keyUpCandidate].Value != KeyDownStates.Up)
+                            {
+                                Log.InfoFormat("StepList canceled. Sending key up on [{0}] key", keyUpCandidate.String);
+                                await keyboardOutputService.ProcessSingleKeyPress(keyUpCandidate.String, KeyPressKeyValue.KeyPressType.Release);
+                                keyStateService.KeyDownStates[keyUpCandidate].Value = KeyDownStates.Up;
+                            }
+                        }
+                    }
+                    else
+                    {
+                        keyStateService.KeyRunningStates[singleKeyValue].Value = false;
+
+                        //if the StepList left any keys down then return without changing the StepList key to Up
+                        if (keyDownList != null && keyDownList.Any())
+                        {
+                            foreach (var keyUpCandidate in keyDownList)
+                            {
+                                if (keyStateService.KeyDownStates[keyUpCandidate].Value != KeyDownStates.Up)
+                                {
+                                    Log.InfoFormat("StepList {0} finished without changing state to Up because key [{1}] is down", singleKeyValue.String, keyUpCandidate.String);
+                                    return;
+                                }
+                            }
+                        }
+                    }
+                    //if no keys are down then change the StepList key to Up
+                    keyStateService.KeyDownStates[singleKeyValue].Value = KeyDownStates.Up;
                 }
             }
         }
 
         private async Task StepListCommands(KeyValue singleKeyValue, List<string> multiKeySelection, List<string> commandList, int nestLevel)
         {
-            Log.InfoFormat("StepListCommands called with a list of {0} commands", commandList.Count);
-            var keyDownList = new List<string>();
-
-            if (nestLevel == 0)
-            {
-                keyStateService.KeyRunningStates[singleKeyValue].Value = true;
-                keyStateService.KeyDownStates[singleKeyValue].Value = KeyDownStates.Down;
-
-                keyDownList.AddRange(commandList.FindAll(x => x.StartsWith("KeyToggle>")));// || x.StartsWith("KeyDown>")););
-            }
+            Log.InfoFormat("StepListCommands called with command count: {0}, nest level: {1}", commandList.Count, nestLevel);
+            
             while (commandList.Any())
             {
-                //if a key stop has been triggerred, then stop processing and return
-                if (!keyStateService.KeyRunningStates[singleKeyValue].Value)
+                //if an external process has ordered this key to stop then return
+                if (!keyStateService.KeyRunningStates[singleKeyValue].Value) 
                 {
-                    //if at the top level execute KeyUp processing
-                    if (nestLevel == 0 && keyDownList != null && keyDownList.Any())
-                    {
-                        foreach (var downKey in keyDownList)
-                        {
-                            var keyUpCandidate = new KeyValue(downKey.Substring(downKey.IndexOf(">")));
-                            if (keyStateService.KeyDownStates[keyUpCandidate].Value != KeyDownStates.Up)
-                                keyStateService.KeyDownStates[keyUpCandidate].Value = KeyDownStates.Up;
-                        }
-                    }
                     return;
                 }
-                Log.InfoFormat("Performing StepList step: {0}.", commandList.First());
 
                 if (commandList.First().StartsWith("Loop>"))
                 {
                     var loopCount = Int32.Parse(commandList.First().Substring(5));
-                    Log.InfoFormat("StepListCommand looping {0} times", loopCount);
+                    Log.InfoFormat("StepListCommand: Looping {0} times", loopCount);
 
                     //determine if there is a loop nested in this one or if this is a solo loop
                     //for a loop with additional nested loops, find the final end loop (/Loop>) command
@@ -2500,7 +2528,7 @@ namespace JuliusSweetland.OptiKey.UI.ViewModels
                     if (commandList.First().StartsWith("Action>") &&
                         (Enum.TryParse<FunctionKeys>(commandList.First().Substring(7), out FunctionKeys actionFunctionKey)))
                     {
-                        Log.InfoFormat("StepListCommand triggering function key {0} press", actionFunctionKey);
+                        Log.InfoFormat("StepListCommand: Press function key: {0}", actionFunctionKey);
                         commandKeyValue = new KeyValue(actionFunctionKey);
                         KeySelectionResult(commandKeyValue, multiKeySelection);
                     }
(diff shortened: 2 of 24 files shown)


JuliusSweetland:
Sounds great. I'll review and get back to you. FYI I've got a fairly busy
week so it might be a few days before I can respond.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -1128,10 +1128,6 @@ private void HandleFunctionKeySelectionResult(KeyValue singleKeyValue)
                     SelectLanguage(Languages.HebrewIsrael);
                     break;
 
-                case FunctionKeys.HungarianHungary:
```
AdamRoden: Oops. Don't worry, I put it back in.

src/JuliusSweetland.OptiKey.Core/Models/KeyValues.cs
```diff
@@ -427,9 +427,16 @@ public static List<KeyValue> KeysWhichCanBePressedDown
...
                     MultiKeySelectionIsOnKey
-                };
+                    };
+
+
+                var defaultList = "abcdefghijklmnopqrstuvwxyz"
+                    .ToCharArray()
+                    .Select(c => new KeyValue(c.ToString(CultureInfo.InvariantCulture)))
+                    .ToList();
+                keys = keys.Concat(defaultList).ToList();
```
AdamRoden: I'm not certain I need this. I may remove later.

JuliusSweetland:
I thought this was a merge issue. I removed these from the keys which can
be pressed and locked down properties. I don't think these are correct.

src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
```diff
@@ -490,12 +491,20 @@ public void ProcessSingleKeyText(string capturedText)
...
+            Log.InfoFormat("Processing input value {0} with key press type {1}", inKey, type);
+
+            FunctionKeys funKey;
+            Char chaKey;
+            if (Enum.TryParse(inKey, out funKey))
+                PublishKeyPress(funKey, type);
+            else if (char.TryParse(inKey, out chaKey))
+                PublishKeyPress(chaKey, type);
+            else
+                Log.InfoFormat("No virtual key code found for '{0}'", inKey);
```
AdamRoden: I'm using this method so I can pass the KeyPressType. I stripped it down from my previous version and redirect most of the handling to PublishKeyPress.

src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
```diff
@@ -1033,20 +1042,27 @@ private void StoreSuggestions(List<string> suggestions)
...
                     Log.InfoFormat("Publishing function key '{0}' => as virtual key code {1}", functionKey, virtualKeyCode);
-                    publishService.KeyDownUp(virtualKeyCode.Value);
+                    if (type == KeyPressKeyValue.KeyPressType.Press)
+                        publishService.KeyDown(virtualKeyCode.Value);
+                    else if (type == KeyPressKeyValue.KeyPressType.Release)
+                        publishService.KeyUp(virtualKeyCode.Value);
+                    else if (type == KeyPressKeyValue.KeyPressType.Toggle)
+                        publishService.KeyToggle(virtualKeyCode.Value);
+                    else
+                        publishService.KeyDownUp(virtualKeyCode.Value);
```
AdamRoden: I expanded the capabilities of PublishKeyPress.

src/JuliusSweetland.OptiKey.Core/Models/KeyValues.cs
AdamRoden: They are definitely not needed and are being removed.

JuliusSweetland:
Thanks

AdamRoden:
@JuliusSweetland 
I'm working toward this concept you laid out and copying the notes to this PR.

When a dynamic key is pressed the first thing we should do is check if the KeyValue is currently infinitely looping and stop it. This gives us the workflow that to start a looping key you select it, and then select it again to stop it. I think we could do this by maintaining (in the KeyStateService) a hashSet of looping keys by KeyValue. When we start an infinite loop then it goes in there and the loop only continues while it exists in there - something like:

while(keyStateService.LoopingKeyValues.Contains(keyValue))
{
//do looping stuff
}

When that key is selected again the first thing it would do is while(keyStateService.LoopingKeyValues.Remove(keyValue) to cause the previous loop to end.

If a key (or the current key state) only contains one Action or Text then that should be the generated Key's value. If we can correctly identify that a dynamic key only has one Action or Text value and then initially create the Key with that value then it should correctly represent key up/down states via the theme.

src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
```diff
@@ -1119,7 +1142,14 @@ private void PublishKeyPress(char character)
                         releaseAlt = true;
                     }
 
-                    publishService.KeyDownUp((VirtualKeyCode)vkCode);
+                    if (type == KeyPressKeyValue.KeyPressType.Press)
```
JuliusSweetland: Looks good, but it is worth considering what happens around the modifier keys. The way this logic works is a query is made to the operating system to figure out which key strokes achieve the character we want to output, based on the user's keyboard layout. This might be that we need to simulate the ALT key being down to be able to fire the correct keystrokes to output the character. This is fine if our action is to press the key down and up as the sequence becomes:
ALT DOWN
KEY DOWN
KEY UP
ALT UP

But what happens if this keypress is part of a dynamic keyboard and the key is only toggled. That would result in:

Dynamic key press 1:
ALT DOWN
KEY DOWN
ALT UP

Dynamic key press 2:
ALT DOWN
KEY UP
ALT UP

Is that going to cause unexpected behaviour? I don't know how the OS, or other apps, might interpret the release of the ALT key - would they think the key press sequence is complete, or might they wait for the release of the actual key...I don't know. Just flagging as a potential pain point. Might be worth adding a few notes in a comment here so we can consider this again if an issue does arise in future.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -128,6 +124,15 @@ private void SetupInputServiceEventHandlers()
                 if (SelectionMode == SelectionModes.Key
                     && (singleKeyValue != null || (multiKeySelection != null && multiKeySelection.Any())))
                 {
+                    if (singleKeyValue != null && singleKeyValue.FunctionKey != null && singleKeyValue.FunctionKey.Value == FunctionKeys.StepList)
+                    {
+                        await StepList(singleKeyValue, multiKeySelection);
+
+                        //do stuff on another thread
```
JuliusSweetland: Minor cleanup - could you remove these commented out lines pls.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
JuliusSweetland: Could you pls update the PR with this reinstated? It is quite hard to keep track of what is/is not in a logical change if it spans multiple PRs. Thanks.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2376,45 +2372,131 @@ private void SelectVoice(string voice)
...
                     }
-                    else if (vStep.StartsWith("Wait>"))
+                    else if (commandList.First().StartsWith("KeyDown>"))
                     {
-                        System.Threading.Thread.Sleep(int.Parse(vStep.Substring(5)));
+                        commandKeyValue = new KeyValue(commandList.First().Substring(8));
+                        Log.InfoFormat("StepListCommand: key down on {0} key", commandKeyValue.String);
+                        await keyboardOutputService.ProcessSingleKeyPress(commandKeyValue.String, KeyPressKeyValue.KeyPressType.Press);
+                        //singleKeyValue represents the visual item
+                        keyStateService.KeyDownStates[singleKeyValue].Value = KeyDownStates.Down;
```
JuliusSweetland: Is there any danger of the singleKeyValue being null? Can you pls give an example of what the singleKeyValue might be compared to the commandKeyValue so I can understand? Could you also pls add these examples to the comments here?

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2376,45 +2372,131 @@ private void SelectVoice(string voice)
...
+                        KeySelectionResult(commandKeyValue, multiKeySelection);
                     }
-                    else if (vStep.StartsWith("Text>"))
+                    else if (commandList.First().StartsWith("KeyToggle>"))
                     {
-                        keyboardOutputService.ProcessSingleKeyText(vStep.Substring(5));
+                        commandKeyValue = new KeyValue(commandList.First().Substring(10));
+                        Log.InfoFormat("StepListCommand: key toggle on {0} key", commandKeyValue.String);
+
+                        if (keyStateService.KeyDownStates[singleKeyValue].Value != KeyDownStates.Up)
```
JuliusSweetland: Could this be simplified to use the ProcessSingleKeyPress call with KeyPressType.Toggle?

src/JuliusSweetland.OptiKey.Core/app.config
```diff
@@ -3,1424 +3,2363 @@
 Copyright (c) 2019 OPTIKEY LTD (UK company number 11854839) - All Rights Reserved
 -->
```
JuliusSweetland: Unless you have added/removed any application settings it's safest to remove these files changes from the PR completely.

src/JuliusSweetland.OptiKey.Core/Models/KeyValues.cs
```diff
@@ -470,8 +477,38 @@ public static List<KeyValue> KeysWhichCanBeLockedDown
...
+                    F10Key,
+                    F11Key,
+                    F12Key,
+                    HomeKey,
+                    InsertKey,
+                    PgDnKey,
+                    PgUpKey
                 };
 
+                var defaultList = "abcdefghijklmnopqrstuvwxyz"
```
JuliusSweetland: Another one to remove?

src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
```diff
@@ -1055,7 +1071,14 @@ private void PublishKeyPress(char character)
...
+                        publishService.KeyDown(virtualKeyCode.Value);
+                    else if (type == KeyPressKeyValue.KeyPressType.Release)
+                        publishService.KeyUp(virtualKeyCode.Value);
+                    else if (type == KeyPressKeyValue.KeyPressType.Toggle)
+                        publishService.KeyToggle(virtualKeyCode.Value);
+                    else
+                        publishService.KeyDownUp(virtualKeyCode.Value);
                     return;
                 }
```
JuliusSweetland: If Settings.Default.PublishVirtualKeyCodesForCharacters == false then toggling/pressing/releasing keys will not work correctly (except for these keys: https://sourcegraph.com/github.com/OptiKey/OptiKey@f5b394472cdff2fe1e020dde1b6ee011e15fd8b9/-/blob/src/JuliusSweetland.OptiKey.Core/Extensions/FunctionKeysExtensions.cs#L16:39). You might want to think about what the correct behaviour is here - if we receive a request to KeyDown/KeyDown/Toggle a character or key and this setting is false then we will ignore that and send the text to the OS in a similar way as if a key down&up had been completed. Either we could add extra checks here and only allow the sending of text if this setting is false AND we are performing a PressAndRelease, otherwise ALWAYS attempt to lookup the virtual key code, or we live with this setting overriding all key behaviour to output text. I would lean towards only sending text if we are performing a PressAndRelease, but log the fact that we are ignoring this setting for Press, Release, and Toggle so at least the log is clear that the settings dictate we send text, but that it makes no sense in this situation.

JuliusSweetland:
@AdamRoden I've added some comments. Thanks.

src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
AdamRoden: I hear you. I've been internally debating what makes the most sense for KeyDown/Up/Toggle in regard to obeying other rules and interacting with modifiers. I'm leaning toward KeyDown/Up/Toggle being considered special use case not necessarily expected to conform to other rules.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
AdamRoden: It is done.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
AdamRoden: Perhaps. Though, sometimes I just want a universal all keys up when dealing with troublesome applications or remote computers.

src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
AdamRoden: As noted in another comment. I think we'll make KeyDown/Up/Toggle special.

src/JuliusSweetland.OptiKey.Core/Models/KeyValues.cs
AdamRoden: It is done.

src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
```diff
@@ -490,12 +491,66 @@ public void ProcessSingleKeyText(string capturedText)
...
+                            publishService.KeyToggle((VirtualKeyCode)vkCode);
+                        else
+                            publishService.KeyDownUp((VirtualKeyCode)vkCode);
+                    }
+                    else
+                    {
+                        Log.InfoFormat("No virtual key code found for '{0}'", chaKey);
+                    }
+                }
+            }
```
AdamRoden: I was really unsatisfied with trying to force KeyDown/Up/Toggle logic into the existing methods, so I isolated them here. These unique actions have special needs and I'm betting some changes yet to be identified.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2376,46 +2395,137 @@ private void SelectVoice(string voice)
...
-                KeyValue singleKeyValue;
-                string[] stringSeparators = new string[] { "<" };
-                foreach (var vStep in command.Split(stringSeparators, StringSplitOptions.None).ToList())
+                //if the key has been stopped, then end any ongoing processing and return.
+                //might need to implement additional keystate clean up here as well
+                if (!keyStateService.KeyRunningStates[singleKeyValue].Value)
+                {
+                    keyStateService.KeyDownStates[singleKeyValue].Value = KeyDownStates.Up;
+                    return;
+                }
```
AdamRoden: This is how I end a running key prematurely.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -115,24 +111,52 @@ private void SetupInputServiceEventHandlers()
...
+                            {
+                                keyStateService.KeyRunningStates[singleKeyValue].Value = false;
+                                keyStateService.KeyDownStates[singleKeyValue].Value = KeyDownStates.Up;
+                            }
+                            else
+                            {
+                                keyStateService.KeyRunningStates[singleKeyValue].Value = true;
+                                keyStateService.KeyDownStates[singleKeyValue].Value = KeyDownStates.Down;
+                                await StepList(singleKeyValue, multiKeySelection);
+                            }
```
AdamRoden: This is the first way to end a running key. Simply activate it again.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2376,46 +2395,137 @@ private void SelectVoice(string voice)
...
+                if (commandList.Any())
+                {
+                    commandList.RemoveAt(0);
+                }
+                //if we are done with all commands and at the top level then the key states can be changed
+                if (!commandList.Any() && nestLevel == 0)
+                {
+                    keyStateService.KeyRunningStates[singleKeyValue].Value = false;
+                    keyStateService.KeyDownStates[singleKeyValue].Value = KeyDownStates.Up;
+                }
```
AdamRoden: This is the second way to end a running key. It finishes all it's tasks on it's own.

A third way to end a running key: ending it from some other key, is not yet built.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
AdamRoden: Done.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
AdamRoden: SingleKeyValue is the full array of action/value commands associated with a Dynamic key and looks like this:
`<StepList><Loop>10<KeyDown>2<Wait>1000<KeyUp>2</Loop>`

CommandKeyValue is what you get when you pull out one of the values and would look like this:
`2`

src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
JuliusSweetland: @AdamRoden I think that's fair, but it may be worth adding extra WARN logging in situations where a modifier key(s) would need to be pressed to achieve the virtual key code - we can continue doing that, but log that the behaviour may be different from what is expected as the modifier will be auto released after the KeyUp/Down/Toggle event has been fired.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
JuliusSweetland: @AdamRoden Understood. Would you please add a comment to that effect early in the method so people coming to the code fresh can understand easily?

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
JuliusSweetland: @AdamRoden I'm not saying we should remove any of KeyUp/KeyDown/Toggle functionality, but instead I was querying if the check on the KeyDownStates[singleKeyValue] needed to be done as there is a later check on the key state in the PublishService. However, I believe this is a different check as it is an OS level check to see what needs to be sent in the event to achieve the correct output, rather than a check on the key state within OptiKey. I think you can ignore my comment.

JuliusSweetland:
@AdamRoden Latest version is looking really good. The logic is quite complicated though, so any comments you could add would be helpful. Thanks!

src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
```diff
@@ -490,12 +491,62 @@ public void ProcessSingleKeyText(string capturedText)
...
+
+            FunctionKeys funKey;
+            var virtualKeyCode = (Enum.TryParse(inKey, out funKey)) ? funKey.ToVirtualKeyCode() : null;
+
+            if (virtualKeyCode != null)
+            {
+                if (type == KeyPressKeyValue.KeyPressType.Press)
+                    publishService.KeyDown(virtualKeyCode.Value);
+                else if (type == KeyPressKeyValue.KeyPressType.Release)
+                    publishService.KeyUp(virtualKeyCode.Value);
```
AdamRoden: I eliminated KeyPressType.Toggle and made it so a Toggle action checks the KeyStateService and sends either Press or Release to the KeyboardOutputService.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -115,24 +111,83 @@ private void SetupInputServiceEventHandlers()
...
+                                Text>a
+                                Loop>3
+                                Text>b
+                                /Loop>
+
+                            commandKeyValues:
+                                a
+                                b
+                                b
+                                b
```
AdamRoden: A big comment and example.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2376,44 +2426,175 @@ private void SelectVoice(string voice)
...
+                    {
+                        foreach (var downKey in keyDownList)
+                        {
+                            var keyUpCandidate = new KeyValue(downKey.Substring(downKey.IndexOf(">")));
+                            if (keyStateService.KeyDownStates[keyUpCandidate].Value != KeyDownStates.Up)
+                                keyStateService.KeyDownStates[keyUpCandidate].Value = KeyDownStates.Up;
+                        }
+                    }
+                    return;
+                }
```
AdamRoden: The start of new logic to handle long running keys and resetting of key states. It's not done yet.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2376,44 +2426,175 @@ private void SelectVoice(string voice)
...
                         else
                         {
-                            ChangeKeyboardKeyValue keyValue = new ChangeKeyboardKeyValue(vStep.Substring(vIndex), vReturn);
-                            ProcessChangeKeyboardKeyValue(keyValue);
+                            commandKeyValue = new ChangeKeyboardKeyValue(commandList.First().Substring(vIndex), vReturn);
+                            Log.InfoFormat("StepListCommand changing keyboard to string {0}. Replace previous keyboard={1}", commandKeyValue.String, vReturn);
                         }
+                        KeySelectionResult(commandKeyValue, multiKeySelection);
+                    }
+                    else if (commandList.First().StartsWith("KeyToggle>"))
```
AdamRoden: I no longer send Toggle to KeyboardOutputService. I make the determination here whether to send Press or Release.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2376,44 +2426,175 @@ private void SelectVoice(string voice)
...
+                                allKeysUp = false;
+                                break;
+                            }
+                        }
+                        if (allKeysUp)
+                            keyStateService.KeyDownStates[singleKeyValue].Value = KeyDownStates.Up;
+                    }
+                    else
+                    {
+                        keyStateService.KeyDownStates[singleKeyValue].Value = KeyDownStates.Up;
```
AdamRoden: This is some more unfinished code that will eventually manage key states when a key is complete.

AdamRoden:
@JuliusSweetland 
I added more comments in my code, made some changes around KeyToggle functionality, and continued work toward cleaning up key states when a key is finished or manually terminated. It's not all working properly yet, but it is coming together.

JuliusSweetland:
It looks great. Thanks for that.

AdamRoden:
@JuliusSweetland 
I finished the code for long running keys. (We need a better term like: Async, Background, Command, List, Macro, Recursive, Task)
I tried to be thorough in my logging and commenting.
The code is now ready for some critical review and testing.

src/JuliusSweetland.OptiKey.Core/UI/Views/Keyboards/Common/DynamicKeyboard.xaml.cs
```diff
@@ -326,13 +331,488 @@ private void SetupErrorLayout(string heading, string content)
...
+                    if (dynamicItem.Col > -1)
+                    {
+                        //if column exists then block all preceding columns in the row
+                        if (openGrid.ElementAt(row).Exists(x => x == dynamicItem.Col))
+                            openGrid.ElementAt(row).RemoveAll(x => x < dynamicItem.Col);
+                        //else block the whole row and check another 
+                        else
+                        {
+                            openGrid.ElementAt(row).RemoveRange(0, openGrid.ElementAt(row).Count);
+                            goto findAvailableRow;
```
JuliusSweetland: Although there's technically nothing wrong with a goto statement, flow control using goto statements is not really correct. Would it be possible to refactor this logic into a method instead?

JuliusSweetland:
@AdamRoden Great PR. Only one comment regarding the use of goto. I'll conduct testing and comment on how that goes next.

src/JuliusSweetland.OptiKey.Core/UI/Views/Keyboards/Common/DynamicKeyboard.xaml.cs
AdamRoden: LOL! You purist. No infinite loops. No goto statements.
You want a refactor? I'll give you a refactor.

JuliusSweetland:
@AdamRoden One possible bug. I started Optikey docked to the top edge with a height of around 40/50%. I switched to the Demo_Gamepad.xml dynamic keyboard and then right clicked on Optikey and selected QUIT. When I restarted Optikey it started docked along the bottom edge suggesting that the default window state had been changed. This may be by design, but the keyboard doesn't appear to specify PersistNewState so I *think* it's a bug. What do you think?

JuliusSweetland:
@AdamRoden Another small thing - by changing the MainWindow Background and BorderBrush to Transparent the standard keyboards have a Transparent border instead of a grey border. You get more correct behaviour if you only specify the Background of the window as Transparent, but this results in the dynamic keyboards having a window border around them, even when transparent (e.g. the Demo_Gamepad.xml).

I think the Window should have nothing specific in it as this prevents the theme from overriding the styling. Dynamic keyboards should override the MainWindow's border and background style. I'm making a few small changes so I'll make this change as I pull this PR into my master branch...

JuliusSweetland:
@AdamRoden By the way I'm impressed with your code. There's some really impressive design and detail in there.

JuliusSweetland:
> 
> 
> @AdamRoden Another small thing - by changing the MainWindow Background and BorderBrush to Transparent the standard keyboards have a Transparent border instead of a grey border. You get more correct behaviour if you only specify the Background of the window as Transparent, but this results in the dynamic keyboards having a window border around them, even when transparent (e.g. the Demo_Gamepad.xml).
> 
> I think the Window should have nothing specific in it as this prevents the theme from overriding the styling. Dynamic keyboards should override the MainWindow's border and background style. I'm making a few small changes so I'll make this change as I pull this PR into my master branch...

@AdamRoden I've added this functionality, but I haven't completed my testing of the core functionality, so I'll finish that in the morning and then merge this into my master branch. You can then rebase from my master and add the goto refactoring...

JuliusSweetland:
@AdamRoden Actually I had time to test - all looking great. I've merged this PR. Could you submit a new PR for the goto refactoring.

Great PR - really powerful stuff in there! If you have any documentation for the new features (including keyboard extract examples) that would be ideal. Again, great stuff!

AdamRoden:
> @AdamRoden One possible bug. I started Optikey docked to the top edge with a height of around 40/50%. I switched to the Demo_Gamepad.xml dynamic keyboard and then right clicked on Optikey and selected QUIT. When I restarted Optikey it started docked along the bottom edge suggesting that the default window state had been changed. This may be by design, but the keyboard doesn't appear to specify PersistNewState so I _think_ it's a bug. What do you think?

@JuliusSweetland Yes. that would be a bug. The way the code works now, the default values only get restored by navigating to a keyboard. We can do a band aid fix for cases like the one you found, but fixing the bug 100% could turn into a significant effort. I'll take a more detailed look later.

JuliusSweetland:
A completely valid fix might be to attach an application exit handler to
restore previous state in the same-ish way as the existing change
keyboard/restore previous dimensions logic. There are a few examples of
hooking into the application exit event in the code, you just have to make
sure we only hook into it once so we only restore from the last dynamic
keyboard if it's not persistent. We can discuss if you're not sure where
this should live.

AdamRoden:
@JuliusSweetland Here's my To Do list:

1. Provide additional sample keyboards and documentation of new features
2. Fix bug in WindowManipulationService
3. Create a master reset key that will stop all running keys and change all key states to Up.
4. Look into transparency overrides and test your patch.
5. Update the other two themes to include new overrides
6-10. ???

JuliusSweetland:
Perfect, thanks.

alexandre-mbm:
> I finished the code for long running keys. (We need a better term like: Async, Background, Command, List, Macro, Recursive, Task)

Could you write a tutorial about how to use the new feature to create "macros"?

JuliusSweetland:
Here you go:
https://github.com/OptiKey/OptiKey/wiki/Creating-and-Using-Dynamic-Keyboards

PR #619, opened 2019-11-19
Refactoring and bug fixing

AdamRoden:


JuliusSweetland:
@AdamRoden Thanks Adam. Is the change to include temp state in the App getters/setters for the windows manipulation service a way to prevent temporary dynamic keyboard state being persisted incorrectly if the user uses the context menu to Qui optikey? Or does there need to be another change to catch the Application.Exit event and restore the last persisted state?

AdamRoden:
I got wrapped up in the idea of ensuring the temp state doesn't become the persisted state if the application is terminated or restarted and neglected to ensure the context menu's 'Back' button behaves properly. It probably need a call to WindowManipulationService.

JuliusSweetland:
Possibly. Does back from the context menu not behave in the same way as a
normal Back key?

The two scenarios to cover would be going back and quitting the application.

JuliusSweetland:
Are you happy to address these?

AdamRoden:
@JuliusSweetland I looked at the code and retested and the back action works properly and via the same code. I don't need to change code further.

JuliusSweetland:
Great news. How about if the user quits while in a dynamic keyboard which
has resized OptiKey in a non-persistant/temporary way. Will OptiKey
incorrectly be in the same state when it starts again?

AdamRoden:
Quitting is fine as well. The way I coded it only the persisted state values ever get permanently stored in Settings. The temp state values are only saved as variables.

JuliusSweetland:
@AdamRoden Great. I will test and aim to create a full release and update to the wiki later tonight then! Great work as always! I have invited to be a team member of the Optikey Github org if this interests you. It allows me to give you greater permissions on the repo going forward, and is a symbolic thank you for what you have contributed so far.

JuliusSweetland:
@AdamRoden Merged. One more thing - could you help me update the bundled dynamic keyboards and let me know what note I should include on this release (and on the wiki page) so users know how to upgrade their existing dynamic keyboards? I know Keys -> Content, but what else has changed? All keys are now DynamicKey?

JuliusSweetland:
@AdamRoden FYI If you have time to make the changes directly you'll need to update from my master branch first. You should see the Resources\Dynamic Keyboards\Samples directory with your keyboards in it.

AdamRoden:
I wonder how hard it would be to write a keyboard translation method that outputs the active keyboard to a string using the new structure.

JuliusSweetland:
It would probably be relatively straightforward, but when would it be used?
I think auto translating old style keyboards into a new style keyboard and
then overriding the user's keyboard would be quite imposing for the end
users, especially if something went wrong, or even formatting or comments
were lost. What did you have in mind?

AdamRoden:
I ran into an issue with plugin keys and rather than kicking the can down
the road, it would be nice to handle it now. To work within a DynamicKey, I
need to wrap the plugin, method, arguments and so on. So, I move PluginKey
inside of DynamicKey, and do some renaming and normalization: XmlElements
PluginKey becomes Plugin, Plugin becomes Name, Arguments gets eliminated.
Sample image below.

The thing is, this small change is just a domino that impacts code, plugin
setup, and wiki documentation. Coding to the new structure is basically
done if we go with it, but the other stuff would need to be updated as well.

[image: image.png]

Adam Roden

JuliusSweetland:
We have breaking changes that are pending release so let's make all these
changes as well and release together. So I understand fully would these
changes break any existing dynamic keyboards which use plugins?

AdamRoden:
With this one as with previous, I can do it in a way where old keyboard
plugins still work. You simply end up with two similar sets of code to
maintain.

Also, I think I found a bug related to the way you did transparency
overrides where it carries forward rather than being a single keyboard
override. Before I dig, do you know why it would do that?

Adam Roden

JuliusSweetland:
Backward compatible would be best if you don't mind. If the old code could
be split out, maybe into methods that are named in such as way as to be
obviously catering to old/obsolete content?

No idea on the transparency. That should only live at key level and all
keys are replaced when the keyboard changes...

PR #621, opened 2019-11-22
Add missing resource

kmcnaught:
I've been merging master (42fa9786) into my own feature branch, and it looks like the resource ERROR_HANDLING_INPUT_SERVICE_SELECTION_RESULT is missing from Resources.resx. It's present in Resources.Designer.cs only.

This adds it to Resources.resx based on the comment saying what it should be in Resources.Designer.cs

@AdamRoden for comment/approval (his commit d511b1786 introduced the string)


AdamRoden:
@kmcnaught It looks like you are right. I'm not sure how that was missed til now, but I'm glad you caught it.

JuliusSweetland:
@kmcnaught Great catch, thanks.

PR #624, opened 2019-11-25
Separate apps + installers

kmcnaught:
To build installers:

- Open OptikeyDeployment.sln in VS
- Build entire solution in Release
- Open all four AIP projects from e.g. <OptikeyRoot>/installers/OptikeyMouse.aip
- Build all in AdvancedInstaller
- Setup files will be built in <OptikeyRoot>/installers/SetupFiles/

Insert code-signing steps along the way as appropriate...

# Possible issue with AdvancedInstaller
Since some dialogs are shared between all installers, you might need to tell AdvancedInstaller to look locally for these files. Go to Settings -> Repository Manager -> top level Settings -> Change repository path. Set it to <OptikeyRoot>/installers

# Ignoring generate code changes
When you build the deployment solution you'll get some changes in git that you'll want to ignore with: 
`git update-index --skip-worktree src/JuliusSweetland.OptiKey.InstallerActions/InstallerStrings.cs`

# Tests
There are a few tests in the Deployment solution. It would be worth hooking them up to CI at some point. I took them out of the main unit test project, since I wanted to keep all the extra deployment-related projects in the separate solution.


kmcnaught:
Sorry, made a mistake and this wasn't an up to date version of my feature branch! Have opened a new PR with the correct commits...

PR #626, opened 2019-11-26
Expands the list of keys that are do not work in a steplist

AdamRoden:


JuliusSweetland:
@AdamRoden Merged. Can you please check this new release and I will release publically once you give it the thumbs up: https://github.com/OptiKey/OptiKey/releases/download/v2.32.0_v4/OptiKeySetup-2.32.0.exe

AdamRoden:
@JuliusSweetland It works great! 
One note: using the transparency trick on the main window does allow mouse clicks to flow through to whatever window is beneath OptiKey, but unfortunately the same cannot be said for the 'Look to Scroll' action. It works with a normal sized keyboard, but it won't work on a full screen keyboard i.e. my FullQwerty sample keyboard.

JuliusSweetland:
@AdamRoden Understood. I want to add functionality for a separate mouse toolbar soon so the fullscreen transparency hack will have a workaround soon.

PR #631, opened 2020-01-24
Resizing

kmcnaught:
Optikey apps can now be re-sized by hand, by dragging the edges/corners.

Notes:
- Can also change from floating/docked in management console, as well as dock position
- Management console is also now resizable, which helps you get out of awkward situations where it doesn't sit well with a floating Optikey.
- Some tidying up of formatting in visuals view

### DIFF
--- a/src/JuliusSweetland.OptiKey.Chat/App.xaml.cs
+++ b/src/JuliusSweetland.OptiKey.Chat/App.xaml.cs
@@ -199,6 +199,12 @@ namespace JuliusSweetland.OptiKey.Chat
                 {
                     mainWindowManipulationService.SizeAndPositionInitialised += sizeAndPositionInitialised;
                 }
+
+                Current.Exit += (o, args) =>
+                {
+                    mainWindowManipulationService.PersistSizeAndPosition();
+                    Settings.Default.Save();
+                };
             }
             catch (Exception ex)
             {
--- a/src/JuliusSweetland.OptiKey.Chat/Properties/Settings.cs
+++ b/src/JuliusSweetland.OptiKey.Chat/Properties/Settings.cs
@@ -1,5 +1,7 @@
 ﻿// Copyright (c) 2019 OPTIKEY LTD (UK company number 11854839) - All Rights Reserved
 
+using JuliusSweetland.OptiKey.Enums;
+
 namespace JuliusSweetland.OptiKey.Chat.Properties
 {
 
@@ -10,6 +12,11 @@ namespace JuliusSweetland.OptiKey.Chat.Properties
         {
             Settings defaultInstance = ((Settings)(global::System.Configuration.ApplicationSettingsBase.Synchronized(new Settings())));
             InitialiseWithDerivedSettings(defaultInstance);
+        }
+
+        public override AppType GetApp()
+        {
+            return AppType.Chat;
         }
 
         /*
--- a/src/JuliusSweetland.OptiKey.Core/Enums/AppType.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Enums/AppType.cs
@@ -6,6 +6,7 @@ namespace JuliusSweetland.OptiKey.Enums
         Pro,
         Mouse,
         Chat,
-        Symbol
+        Symbol,
+        Tests
     }
 }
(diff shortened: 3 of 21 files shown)


src/JuliusSweetland.OptiKey.Chat/App.xaml.cs
```diff
@@ -199,6 +199,12 @@ private void App_OnStartup(object sender, StartupEventArgs e)
                 {
                     mainWindowManipulationService.SizeAndPositionInitialised += sizeAndPositionInitialised;
                 }
+
+                Current.Exit += (o, args) =>
+                {
+                    mainWindowManipulationService.PersistSizeAndPosition();
```
JuliusSweetland: @kmcnaught Do you know if there is the potential for this to incorrectly persist a temporary window state, e.g. in a dynamic keyboard which has temporarily resized optikey?

JuliusSweetland:
@kmcnaught I'm going to merge this as it looks great, but there's a question I addressed to you regarding potentially persisting a temporary state. Let me know when you have a second. Thanks very much!

JuliusSweetland:
@kmcnaught Hi Kirsty. I've been making some other changes and testing this PR today - it looks like there's a bug that I cannot resolve. I've had to revert these commits for the moment. Essentially if you use Symbol or Chat (I presume, I've only testing Symbol), then open the Management Studio, then click OK to close it, the fullscreen Symbol/Chat main window is incorrectly set to docked. The change to the VisualViewModel Apply method appears to be the culprit, but when I try debugging and setting both the MainWindowState and MainWindowPreviousState to Maximised (on startup) you get the same behaviour, so for now I've removed the commits.

Secondly, I've also noticed that when Symbol starts up it first appears to register an AppBar (as Visual Studio is resized, sometimes permanently), before opening in fullscreen. I tried to look into what is going on, but I actually can't see how fullscreen (Maximised) is achieved for Symbol or Chat - can you point me in the right direction?

Thirdly, should Symbol and Chat be overriding settings to set their MainWindowState and MainWindowPreviousState to Maximised for correctness?

Lastly, your commits to allow manual resizing of Optikey added some settings to the Visual tab of the Management Console, but these only appeared to allow the user to set up docked positioning. Should these be:
a)App specific, i.e. not available in Chat and Symbol?
b)Also have options available for Maximised?

Thanks!

kmcnaught:
Apologies, I missed your earlier question completely. I'll give you some responses tomorrow.

kmcnaught:
In approximately reverse order, since those are easier...

- The Visuals tab only shows window options for Mouse and Pro (i.e. there's a visibility=hidden binding for `MaximisedApp`).

- Would it ever make sense for Mouse/Pro to be maximised? If so, we should add that as an option, but I can't think of an obvious use case. 

- Yes, it does make sense for Chat/Symbol to override everything to maximised in Settings. Having added this just now in my original branch, I don't get the issue you described with an appbar being registered before maximising (but I did see this before overriding the settings). 

  ( See https://github.com/kmcnaught/OptiKey/tree/resizing )

- Do Chat/Symbol ever get minimised? It doesn't look like it to me. If they do have that functionality, then the above fix will need more thought. 

- The above fix also seems to prevent the behaviour you describe when leaving settings, as far as I can tell - can you test it too please?

The default maximising logic seems to be in MainViewModel::InitialiseKeyboard, where there's a special case for 'Conversation'-ey keyboards. This maximises them on initialisation, so it makes sense that sometimes the window could have been docked until the keyboard loads. 

Finally, on your first question - I suspect there could indeed be bad interactions with persisting state from a dynamic keyboard - good spot. I don't think I'd realised dynamic keyboards could change window state when I was writing that. I'll have to have a play to see.

JuliusSweetland:
@kmcnaught 

> Would it ever make sense for Mouse/Pro to be maximised? If so, we should add that as an option, but I can't think of an obvious use case.
Pro can be maximised when you select 'Conversation' from the menu. This is equivalent to the Chat app.

> Do Chat/Symbol ever get minimised? It doesn't look like it to me. If they do have that functionality, then the above fix will need more thought.
No I don't think they should ever be minimised.

> Finally, on your first question - I suspect there could indeed be bad interactions with persisting state from a dynamic keyboard - good spot. I don't think I'd realised dynamic keyboards could change window state when I was writing that. I'll have to have a play to see.
I would like to wait until this is in place before I test the whole thing. No rush at all. Are you happy for me to wait for this and then I can test and go with a new PR? Seriously no rush...

kmcnaught:
Sure, no problem,

PR #641, opened 2020-02-05
Addresses issue 635: Floating dynamic keyboards don't always remove a…

AdamRoden:
…ppbar

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Services/WindowManipulationService.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Services/WindowManipulationService.cs
@@ -524,14 +524,17 @@ namespace JuliusSweetland.OptiKey.Services
                         ? screenBoundsInDp.Right - newWidth + horizontalOffset
                         : screenBoundsInDp.Width / 2d - newWidth / 2d + horizontalOffset;
                 }
+                if (oldWindowState == WindowStates.Docked)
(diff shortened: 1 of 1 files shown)


src/JuliusSweetland.OptiKey.Core/Services/WindowManipulationService.cs
```diff
@@ -524,14 +524,17 @@ public void OverridePersistedState(bool inPersistNewState, string inWindowState,
...
                 savePersistedState(inPersistNewState);
                 saveFloatingSizeAndPosition(new Rect(newLeft, newTop, newWidth, newHeight));
             }
             else
+            {
+                if (oldWindowState == WindowStates.Docked)
+                    UnRegisterAppBar();
                 savePersistedState(inPersistNewState);
-
-            if (oldWindowState == WindowStates.Docked && newWindowState != WindowStates.Docked)
```
JuliusSweetland: @AdamRoden Question: the old logic checks that the new state is not also docked before unregistering the app bar. Is this definitely no longer needed? I.e. is it possible that Optikey changes from docked to docked, but the app bar is unregistered incorrectly?

Thanks!

JuliusSweetland: @AdamRoden Ignore that. I've just checked the code in full. All good.

PR #649, opened 2020-03-02
Adds KeysWhichCanBeLockedDown to dynamic keys KeyFamily

AdamRoden:


src/JuliusSweetland.OptiKey.Core/Models/KeyEnabledStates.cs
```diff
@@ -80,7 +83,8 @@ public class KeyEnabledStates : BindableBase
 
                 //Key is not Sleep, but we are sleeping
                 if (keyStateService.KeyDownStates[KeyValues.SleepKey].Value.IsDownOrLockedDown()
-                    && keyValue != KeyValues.SleepKey)
+                    && keyValue != KeyValues.SleepKey
+                    && !keyFamily.Exists(x => x.Item1 == keyValue && x.Item2 == KeyValues.SleepKey))
```
JuliusSweetland: @AdamRoden What does this do? Only from the changed files on GitHub I'm struggling to understand the meaning if this line. Thanks

AdamRoden: @JuliusSweetland KeyFamily is the collection of parent KeyValues (Item1) and the KeyValues of child commands that could be locked down when the parent key is triggered (Item2). My scenario is a key that changes keyboards and executes Sleep, and then another keyboard that starts off sleeping and needs to wake up and changes to another keyboard.

Basically, any key with a subcommand of Sleep should not be disabled by Sleep.

JuliusSweetland:
Got it. Could you please add that as a comment on the line and bundle it
into the same PR and I'll merge it in?

AdamRoden:
@JuliusSweetland Comment added.

PR #650, opened 2020-03-03
Feature/issue 613 translation key

SamuelSnowball:
Please go to my branch and play around with this! 

Notes:
1. The additional Yandex requirements must be implemented as a part of this feature! (on git about page/website etc) read more here: https://tech.yandex.com/translate/doc/dg/concepts/design-requirements-docpage/
2. The default target translation language is english, and will result in a (handled) exception if you try to convert a language to itself. 
3. The translated text is automatically copied to the clipboard (without the License text). 

Comments:
-Does the loading of settings work for you? Or do you have to re-enable translation every time? If this is the case it isn't so bad, but putting in the API key would be.
-Do we need something to let users know text has been copied? is mentioning it in the console enough?
-Changing of voice will be as an add-on to this feature
-The translation key hasnt been added to the default sample dynamic keyboards

Functionality added:
-Added translation key which can be enabled through management console, it has options for:
	-changing target translation language, default is set to english.
	-changing the API key
-when translation key enabled and pressed, it gets passed scratchpad text
-some special characters in the input text are replaced, it doesnt properly convert all special characters, like some of the currency ones, they turn into ?s
-request is sent to the Yandex Translation API, the API auto detects input language
-scratchpad text is translated and replaced
-success/failure sound and notification is shown (if there was any exception)
-exceptions logged if there are any
-translated text is set in the scratchpad, along with the License text
-translated text (without License text) is automatically copied to clipboard

(This is my first contribution so I'm sure there will be lots of feedback, happy to discuss anything and update code where needed!)


SamuelSnowball:
I have just realized, the "powered by yandex" text is spoken as well as the translated text - will update this

src/JuliusSweetland.OptiKey.Core/Services/Translation/TranslationService.cs
```diff
@@ -0,0 +1,105 @@
...
+        /*
+         * Returns the translation status, translated text and exception content
+         */
+        public async Task<Response> translate(string text)
+        {
+            Response response;
+            response.status = ""; response.translatedText = ""; response.exceptionMessage = "";
+
+            if (!string.IsNullOrEmpty(Settings.Default.SeparateTranslationApiKey)) {
+                this.apiKeyToUse = Settings.Default.SeparateTranslationApiKey;
```
JuliusSweetland: OverrideTranslationApiKey?

src/JuliusSweetland.OptiKey.Core/Services/Translation/TranslationService.cs
```diff
@@ -0,0 +1,105 @@
...
+            public string exceptionMessage;
+        }
+        
+        /*
+         * Returns the translation status, translated text and exception content
+         */
+        public async Task<Response> translate(string text)
+        {
+            Response response;
+            response.status = ""; response.translatedText = ""; response.exceptionMessage = "";
```
JuliusSweetland: Does this work? From just looking at the code the response variable would be uninitialised so setting properties on it should cause a NullReferenceException, no?

Either way could this line be split over 3 lines.

src/JuliusSweetland.OptiKey.Core/Services/Translation/TranslationService.cs
```diff
@@ -0,0 +1,105 @@
...
+        public const string LICENSE_TEXT = "(Powered by Yandex.Translate)";
+
+        public TranslationService(HttpClient httpClient)
+        {
+            this.client = httpClient;
+        }
+
+        public struct Response
+        {
+            public string status;
```
JuliusSweetland: Could these be recased to be Title Cased please?

Also I wouldn't mind a constructor - structs don't NEED to be newed up, but if you don't call the constructor (and set the properties via the constructor) then you need to set ALL of the properties before the struct can be used. I just prefer the unambiguous approach of using a ctor.

src/JuliusSweetland.OptiKey.Core/Services/Translation/TranslationService.cs
```diff
@@ -0,0 +1,105 @@
...
+            public string translatedText;
+            public string exceptionMessage;
+        }
+        
+        /*
+         * Returns the translation status, translated text and exception content
+         */
+        public async Task<Response> translate(string text)
```
JuliusSweetland: Title case on the method name please

src/JuliusSweetland.OptiKey.Core/Services/Translation/TranslationService.cs
```diff
@@ -0,0 +1,105 @@
...
+                using (var provider = CodeDomProvider.CreateProvider("CSharp"))
+                {
+                    provider.GenerateCodeFromExpression(new CodePrimitiveExpression(input), writer, null);
+                    return writer.ToString();
+                }
+            }
+        }
+    }
+
+    class YandexResponse
```
JuliusSweetland: Could you explicitly mark this as private. I like explicit code. Thx

src/JuliusSweetland.OptiKey.Core/Services/Translation/TranslationService.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        public int code { get; set; }
```
JuliusSweetland: Title case for properties please.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2279,15 +2280,66 @@ private void HandleFunctionKeySelectionResult(KeyValue singleKeyValue)
                     break;
 
                 case FunctionKeys.Speak:
+                    string textToSpeak = "";
+                    if (!string.IsNullOrEmpty(keyboardOutputService.Text))
+                    {
+                        if (keyboardOutputService.Text.Contains(TranslationService.LICENSE_TEXT))
```
JuliusSweetland: Could you please remove the license text concept entirely? I know there's a requirement (https://tech.yandex.com/translate/doc/dg/concepts/design-requirements-docpage/), but I think the UX is wrong. I will contact Yandex before releasing the code to discuss with them. It may be sufficient to give them a shout somewhere else.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2279,15 +2280,66 @@ private void HandleFunctionKeySelectionResult(KeyValue singleKeyValue)
...
+                        string textFromScratchpad = KeyboardOutputService.Text;
+
+                        if (!string.IsNullOrEmpty(textFromScratchpad))
+                        { 
+                            TranslationService.Response response = await translationService.translate(textFromScratchpad);
+                            if (response.status == "Error")
+                            {
+                                string errorToShow;
+
+                                if (response.exceptionMessage == "") {
```
JuliusSweetland: Could you please push the curly bracket onto the next line. Extremely minor I know...

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2279,15 +2280,66 @@ private void HandleFunctionKeySelectionResult(KeyValue singleKeyValue)
...
+                                    errorToShow = response.exceptionMessage;
+                                }
+                                else {
```
JuliusSweetland: Push { onto the next line pls

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2279,15 +2280,66 @@ private void HandleFunctionKeySelectionResult(KeyValue singleKeyValue)
...
+                                    errorToShow = "Translation response was empty! is target language set?";
+                                }
+
+                                Log.Error("Error/exception during translation: " + errorToShow);
```
JuliusSweetland: Could you use string interpolation, i.e. $"Error or exception during translation: {errorToShow}";

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.ServiceEventHandlers.cs
```diff
@@ -2279,15 +2280,66 @@ private void HandleFunctionKeySelectionResult(KeyValue singleKeyValue)
...
+                                RaiseToastNotification(Resources.ERROR_DURING_TRANSLATION, errorToShow, NotificationTypes.Error, () =>
+                                {
+                                     inputService.RequestResume();
+                                });
+                            }
+                            else
+                            {
+                                keyboardOutputService.ProcessFunctionKey(FunctionKeys.ClearScratchpad);
+                                // Must include Powered by Yandex.Translate as license requirement
+                                keyboardOutputService.Text = response.translatedText + " " + TranslationService.LICENSE_TEXT;
```
JuliusSweetland: As mentioned let's drop that and I'll figure out a work around!

JuliusSweetland:
@SamuelSnowball Good PR from the look of things! I've added some minor comments. If you're happy to accommodate them I'll perform testing when the PR is updated. Thanks Sam!

SamuelSnowball:
> @SamuelSnowball Good PR from the look of things! I've added some minor comments. If you're happy to accommodate them I'll perform testing when the PR is updated. Thanks Sam!

Will address the comments, should be done on monday not around much this weekend

JuliusSweetland:
No rush. I'm pretty snowed as well!

SamuelSnowball:
@JuliusSweetland 
I think that is everything addressed

JuliusSweetland:
@SamuelSnowball I have made a handful of very minor changes, but everything looks good! Merged!

Thank you very much. What are you thinking about auto changing the language after translation and then resetting it when the text is cleared? I think it should be do-able.

SamuelSnowball:
> @SamuelSnowball I have made a handful of very minor changes, but everything looks good! Merged!
> 
> Thank you very much. What are you thinking about auto changing the language after translation and then resetting it when the text is cleared? I think it should be do-able.

Did yandex respond? and pretty much, I tried to quickly add it onto this but requires a bit more effort. There just needs to be a mapping between the translated text and the voice - ideally they would match with the same country code. I need to install different language packs and create a solution for all of them, I will raise as new issue.

JuliusSweetland:
I've contacted them so let's see what they say.

I had two suggestions:

1. The Voice for the translated text should be configurable by the user. We
could use logic to suggest a default voice when the translation language
changes, but it might be easier to leave it as an optional thing (new
setting: auto change voices to speak translated text) and if set have that
voice set when the translated text is received, and unset when the
scratchpad text is next null or an empty string (as the user might just
backspace to delete it, rather than select the clear key).

2. Split out a new setting for whether the translated text is placed on the
clipboard automatically. This gives the user a little more control.

SamuelSnowball:
@JuliusSweetland Opened as new issue

SamuelSnowball:
> @JuliusSweetland Opened as new issue

https://github.com/OptiKey/OptiKey/issues/657

kmcnaught:
@SamuelSnowball hi there, we've found that the Yandex service has recently been returning an error code - see #849. What are the limits on the API key you were using? Should we get a new one ourselves, or be encouraging users to get their own? If you're able to check your accounts and find out what the status is, that would be great, thank you

SamuelSnowball:
@JuliusSweetland @kmcnaught 

Hi both,

TLDR: 
I can't find the credentials, why I didn't share the credentials I have no idea!
I think it'll be best to create a new account and key here: https://translate.yandex.com/developers/keys

Their website says API keys shouldn't expire, however it also says "Starting May 27, 2020, free API keys aren't issued." https://yandex.com/dev/translate/doc/dg/concepts/about.html
"Individuals can register (https://translate.yandex.com/developers/account) and receive a grant for $75 (5 million characters) to test the Yandex.Translate API after linking a bank card (no charges will be made)."
Source: https://translate.yandex.com/developers/keys

As an aside / random observation I believe the endpoint we are using is v1.5, at least that's specified within the URL we use: "https://translate.yandex.net/api/v1.5/tr.json/translate?" (https://github.com/OptiKey/OptiKey/blob/master/src/JuliusSweetland.OptiKey.Core/Services/Translation/TranslationService.cs)
Slightly odd, as V5 is out, and has been for 7 years.. their docs also say versions 1-4 have been disabled, so not entirely sure what's going on there.

Issue #667, opened 2020-04-17
Dynamic key virtual keycodes incorrect for characters requiring modifiers


kmcnaught:
I was trying to make a dynamic keyboard which would press "<" and ">" for a game. I found it gave me "," and "." instead. 

I've tracked it down to the method `KeyboardOutputService.ProcessSingleKeyPress` which calls `PInvoke.VkKeyScanEx` to get a virtual key code, but doesn't then do anything with the modifier information returned. 

I'll open a PR with a fix.

@AdamRoden @JuliusSweetland , I *think* the desired behaviour is for dynamic keyboards to respect the global Optikey modifier state, i.e. if I lock the Shift button down in Optikey, it *should* affect my dynamic keys, and turn a "," into a "<". Do you agree? For symbols this could be a bit unexpected, but it general it seems reasonable that modifier keys should work with dynamic keyboards.

JuliusSweetland:
I agree that if the output character requires modifiers that they should be
pressed (or released) as necessary. To be clear I don't think modifiers
pressed outside of a keyboard should affect the current keyboard, unless
those keys exist on the current keyboard and are pressed. I don't think
that's what you are suggesting though.

Am I correct in thinking that your PR will just ensure that the required
modifier keys are down to be able to output the requested character and
then released afterwards (if not locked down already)?

kmcnaught:
Sorry I actually realised I'd misunderstood something in the existing modifier logic. 

>  just ensure that the required
modifier keys are down to be able to output the requested character and
then released afterwards (if not locked down already)?

Yes, this is what it should be doing.

JuliusSweetland:
Cool. I'm surprised it isn't doing that. From memory the modifiers should
be checked and executed as required. There's some logic in the Keyboard
output service

kmcnaught:
Yeah, but at some point I think I half-copied your correct solution and missed some of the logic out...

kmcnaught:
to be precise on how modifiers work now on a dynamic keyboard: 
1) if your character requires a modifier, and the modifier *isn't* held down within optikey, it holds it down temporarily while executing the requested key [press/release/prese-and-release] and then releases the modifier
2) if your character requires a modifier, and the modifier is already held down, it just does the keypress and leaves the modifier on (which gives the correct result). 
[...]

PR #668, opened 2020-04-17
Fix virtual key codes #667

kmcnaught:
Fix for bug #667, use the modifiers returned by vkKeyScan when processing key presses requested from a dynamic keyboard.

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
@@ -517,16 +517,11 @@ namespace JuliusSweetland.OptiKey.Services
 
                 foreach (var chaKey in inKey)
                 {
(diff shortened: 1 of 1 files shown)


JuliusSweetland:
@kmcnaught Looks good, but do you think the SingleKeyPress method should fail to output if the virtual key code cannot be found? The other method falls back to using a different windows API to type the text:

```
Log.InfoFormat("No virtual key code found for '{0}' so publishing as text (OS keyboard layout:{1})",
character.ToPrintableString(), keyboardCulture);
publishService.TypeText(character.ToString());
```

Admittedly I've not done a deep dive to refamiliarise myself with the code, but it feels like the code to output a single key should be essentially identical between the two methods, including moving the code which works out the keyboard culture into the else clause just before publishService.TypeText is called. That means we can de-dupe the code quite a bit.

What do you think?

kmcnaught:
Fair point, I've extracted more of the common code, and now the logging (including culture info) and the fallback option are both shared

PR #669, opened 2020-04-25
Allow any virtual keycode enum for dynamic keypress

kmcnaught:
Currently you can specify function keys in a dynamic keyboard with things like:
`<KeyDown>Home</KeyDown>`

However, this is limited to whatever has been exposed via the FunctionKeys enum, which specifies an associated virtual key code. Today, I wanted a key to press "Return" (as opposed to typing "\n" which the alpha keyboard does) and this was not exposed as a FunctionKey since Optikey doesn't use it elsewhere. The current behaviour is that you just get "Return" typed out letter by letter. 

It would be nice to be able to specify any virtual key from `WindowsInput.Native.VirtualKeyCode`. This PR changes the `ProcessSingleKeyPress` so it first searches for a match in Optikey's `FunctionKeys` enum, and then in `WindowsInput.Native.VirtualKeyCode` before falling back to typing characters. 

I don't think there's anything controversial about this, but it would be worth @AdamRoden confirming.

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Services/KeyboardOutputService.cs
@@ -506,15 +506,28 @@ namespace JuliusSweetland.OptiKey.Services
                     publishService.KeyUp(virtualKeyCode.Value);
                 else
                     publishService.KeyDownUp(virtualKeyCode.Value);
+                return;
             }
-            else
+
+            // If not an Optikey FunctionKey, check if it matches a virtual key name
+            VirtualKeyCode vkCode;
+            if (Enum.TryParse(inKey, true, out vkCode)) // ignore case
             {
-                // Actual key presses
-                foreach (var chaKey in inKey)
-                {
-                    this.PressKey(chaKey, type);
-                }
+                if (type == KeyPressKeyValue.KeyPressType.Press)
+                    publishService.KeyDown(vkCode);
+                else if (type == KeyPressKeyValue.KeyPressType.Release)
+                    publishService.KeyUp(vkCode);
+                else
+                    publishService.KeyDownUp(vkCode);
+                return;
             }
+            
+            // Otherwise a vanilla key press
+            foreach (var chaKey in inKey)
+            {
+                this.PressKey(chaKey, type);
+            }
+        
         }
 
         private string CombineStringWithActiveDeadKeys(string input)


AdamRoden:
@kmcnaught I agree with this idea. How are you defining virtual key codes in the xml file to make this work? (0x01, 0x02, etc.)?

kmcnaught:
It's using WindowsInput.Native.VirtualKeyCode enums from InputSimulatorPlus.1.0.7, i.e. anything from here: https://github.com/TChatzigiannakis/InputSimulatorPlus/blob/master/WindowsInput/Native/VirtualKeyCode.cs

For instance for my current case I'm using 
```
<KeyDown>Return</KeyDown>
<Wait>100</Wait>
<KeyUp>Return</KeyUp>
```
I've set the enum parsing to be case-insensitive (the enums are actually all capitalised) but I've now realised that makes it inconsistent with FunctionKeys, which get parsed in a case-sensitive way. I'd be comfortable with both being made case-insensitive if that's okay with you? That is the most user-friendly option imo.

AdamRoden:
The one thing I strongly suggest is that that rather than 
`<KeyDown>Return</KeyDown>`
you use 
`<KeyDown>{Return}</KeyDown>`
Also, make it case-insensitive.

kmcnaught:
@AdamRoden, I'm not sure about this - currently anything that is an Optikey function key is already used like so:
`<KeyDown>Home</KeyDown>`

It would be inconsistent if FunctionKeys are referred to differently to native VirtualKeyCodes, so we'd want to also change how FunctionKeys are parsed which would break existing dynamic keyboards. 

Are there any scenarios where you would envisage someone using multiple characters within a single Key Up/Down event? Although it does currently default to typing out characters one by one if it can't parse any special keys, I can't imagine you would intentionally use this instead of `<Text>` . 

i.e. you would do:
```
<KeyDown>w</KeyDown>  // automatically derives correct key code for character "w"
<KeyDown>Home</KeyDown> // converts to Optikey FunctionKeys enum -> key code
<KeyDown>Return</KeyDown> // converts to InputSimulatorPlus VirtualKeyCodes enum -> key code
<Text>Hello, world!</Text> // types out character by character
```
but I wouldn't expect anyone to do:
`<KeyDown>Hello, world!</KeyDown>`

Unless I'm missing something, I think this means the syntax is already pretty unambiguous, so I'm not convinced that adding a special character to signify a special key is helpful, and it gives users an extra rule to learn. 

I can however envisage a _potential_ further fallback of letting a user specify an exact key code directly, e.g. `0x02` or `2` in which case you might need to introduce special syntax. But I'm not sure whether this is really required.

AdamRoden:
@kmcnaught You are correct. \<KeyUp> and \<KeyDown> are totally fine. I was thinking more along the lines of being able to embed the new keys in \<Text>, which cannot be done with Action keys so my thought is irrelevant. You may proceed.

JuliusSweetland:
@kmcnaught @AdamRoden Merged and new pre-releases created here: https://github.com/OptiKey/OptiKey/releases/tag/v3.0.5_pre-release3

It would be greatly appreciated if you could test these and let me know if you are happy for them to go out? Thank you.

PR #673, opened 2020-05-05
Fix #672 ReturnToHere parsing.

### LINKED ISSUE #672: Dynamic Keyboards "BackReturnsHere" attribute is wrong
Currently with a `ChangeKeyboard` action, the behaviour is incorrect:

`default`: returns to this keyboard (correct)
`BackReturnsHere="true"`: does _not_ return to this keyboard (wrong)
`BackReturnsHere="false"`: [Edit typo!] _does_ return to this keyboard (wrong)

I'll raise a PR. 

Here's an example set of keyboards to demo.
[test_keyboards.zip](https://github.com/OptiKey/OptiKey/files/4580753/test_keyboards.zip)

kmcnaught:
The default was wrong, *and* the parsed values got flipped, so it
seemed that the default was correct :)

@AdamRoden, I don't know if any built-in example keyboards need amending?

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Models/XmlKeyboardModels/XmlDynamicKey.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Models/XmlKeyboardModels/XmlDynamicKey.cs
@@ -43,7 +43,7 @@ namespace JuliusSweetland.OptiKey.Models
 
         [XmlIgnore]
         public bool BackReturnsHere
-        { get; set; }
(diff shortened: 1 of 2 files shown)


AdamRoden:
You are right. I will need to update some examples.

JuliusSweetland:
@AdamRoden I will merge this change now and wait for the example keyboard updates to follow.

Thanks @kmcnaught & @AdamRoden

PR #675, opened 2020-05-11
Fix bug in dynamic key locking logic.

AdamRoden:


JuliusSweetland:
@AdamRoden Good stuff. Is this everything that was outstanding? I.e. should I create a new pre-release tomorrow pending a release as soon as everyone is happy?

AdamRoden:
Yes. Now would be a good time to do a release. I've been very pleased with
everything.

Adam Roden

JuliusSweetland:
@AdamRoden Cool. I've had a bug report come through today so I'll see if that needs to be fixed and then prepare something asap.

Thanks, as always.

JuliusSweetland:
@AdamRoden Oh, and there were also updates to dynamic keyboards required, or are they no longer needed?

AdamRoden:
@JuliusSweetland I will prepare an updated set of sample keyboards for inclusion in the release.

PR #678, opened 2020-06-02
Resizing

kmcnaught:
As per https://github.com/OptiKey/OptiKey/pull/631, with more testing and handling of corner cases, including dynamic keyboards. 

## What I've implemented:

 -  If an app is floating or docked (but _not_ minimised or maximised) it has a tiny grab handle in the bottom right, and all four edges are grabbable
 -  You can resize by dragging these edges.
 -  You can also move by left-click dragging on the main keyboard (as opposed to the edges)
 -  These changes get saved to settings whenever you close optikey or go to a different window state (e.g. minimise/maximise/dynamic)
 -  If the keyboard is docked, once you let go of the mouse the new shape will be coerced appropriately to still be docked on the same edge

	 -  there is still an unfortunate visual jump as it reinstates the appbar (same as you see when resizing through the optikey size & position keyboard)
 -  I've disabled all 'clever' OS stuff like double-clicking to maximise, snapping to edges and snapping to fill half the screen.
 -  Any resizing you do gets persisted via the settings, except with dynamic keyboards, see below. 
 -  The management window is also resizable, and does allow double-click to resize which can be helpful when the title bar gets hidden behind a floating optikey.
 -  Management Window has options for window state, dock position, etc

	 -  These are hidden in fullscreen apps like Chat
	 -  These are disabled if Optikey is in a temporary window state, e.g.

		 -  currently minimised
		 -  currently maximised
		 -  currently in a dynamic keyboard with temporary position overrides (see below)
	 -  If you open the management console in a 'resizable' state and then change to one of the above states, any changes you made to the layout in the settings will be silently ignored

## Behaviour with dynamic keyboards:
 -  The existing behaviour was :

	 -  when any dynamic keyboard was loading, it marked the window state is being overridden (`OptiKeyApp.persistedState = false`), which already deals with not propagating size/position changes.
	 -  the exception was/is when a keyboard defines `<PersistNewState>true</PersistNewState>` in which case the keyboard's window state becomes Optikey's new state
 -  I've modified this so that any dynamic keyboards which don't actually override any position properties (such as windowState, position, dockSize, width, height, ...) don't flag `OptiKeyApp.persistedState = false`
 -  If a dynamic keyboard overrides any size/position properties, then you *can* still resize with grab handles, or drag to move, but this is only temporary and gets reverted when you reload / go back to a different keyboard

	 -  It's possible you might argue that we should just turn off manual resizing completely for these dynamic keyboards. This would likely be easy to do, but I can imagine it might sometimes be useful (for instance you might want to drag it out the way of something temporarily). Opinions welcome. 
 -  If a dynamic keyboard *does not* override any size/position properties, you can resize as per usual and this persists, since it's inheriting Optikey's existing window state
 -  While you're in a dynamic keyboard with overridden properties (or rather, when OptiKeyApp.persistState == false) the options for window state, size and position in the Management Console are visible but disabled (since they don't apply to the current state of Optikey). 

## Open questions:
 -  I've included "drag to reposition" functionality, where you can left click anywhere on optikey and move it around. I think this is helpful, but I'm not sure if it might be undesirable in some circumstances. The only thing I can think of is if someone is using a touch screen with imprecise motor control. What do you think? Should this be disabled, or optionally disabled?

## Demo
Currently uploading here: https://www.loom.com/share/aa0bdb75d7754e1c87be024089553584


kmcnaught:
FYI I'm still investigating an issue where the manual mode toasts are somehow getting interferred with when rapidly clicked, which gives negative `SuspendRequestCount`s and repeated toast behaviour

kmcnaught:
Okay, it's back in a good place now. There were quite a few subtleties around mouse focus and popups in the end, so I'd appreciate some aggressive testing.

JuliusSweetland:
@kmcnaught This looks really good. The video tour is really useful for features like this. I'll give it a proper test ASAP and go from there.

One thing: what do you think about the Size & Position button that's on the Optikey Mouse keyboard? Do you think this is still needed? If so I've just been informed that it is missing from the portrait version of the expanded mouse keyboard, and I also noticed that an empty key might be useful below the Size & Position key as there's currently a gap with no key and it's noticeable as the key borders don't line up perfectly. Minor though...

kmcnaught:
I'm not sure - obviously the drag-handling isn't eye-gaze friendly, so this resizing is more for either a support person setting things up, or for a user who has limited mouse use for occasional tasks. The size & position keyboard gives full eye gaze access to all of this, so we probably wouldn't want to lose it for Optikey Pro at least. I'm more ambivalent about Optikey Mouse. I know that doesn't answer your question, I'm leaving the call up to you!

JuliusSweetland:
@kmcnaught Great PR Kirsty - thank you!

JuliusSweetland:
@kmcnaught New pre-release of 3.0.6 available here if you want to try it out: https://github.com/OptiKey/OptiKey/releases/tag/v3.0.6_pre-release_2

Issue #679, opened 2020-06-02
Crash upon launch with key/mouse trigger source


kmcnaught:
# How OptiKey was being used
- Version 3.0.5
- deleted all settings to get default
- Launch OptikeyPro
- Open management console, change key selection source to key (Insert) _or_ mouse button (left mouse)
- Close and allow restart
- Optikey crashes, see log

[OptiKeyPro.log](https://github.com/OptiKey/OptiKey/files/4718613/OptiKeyPro.log)

This is also present on origin/master, unsurprisingly.

kmcnaught:
there's a quick null conditional fix in #680 if you're happy with it

PR #680, opened 2020-06-02
Fix possible null pointer #679

kmcnaught:
In particular, overrideTimesByKey was null when a trigger other than
dwell fixation is being used


kmcnaught:
Second commit (714d4aa) fixes #681, but I propose this needs a bit more consideration :-) Either a thorough search of other places that assume `overrideTimesByKey` is not-null, or initialise an empty dictionary for other trigger sources.

PR #684, opened 2020-06-18
Resizing improvements

kmcnaught:
See commit logs. Couple of small fixes + expose floating size/position in management console.

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Properties/Resources.Designer.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Properties/Resources.Designer.cs
@@ -2051,6 +2051,15 @@ namespace JuliusSweetland.OptiKey.Properties {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Floating position (left, top, width, height).
+        /// </summary>
+        public static string FLOATING_POSITION_LABEL {
+            get {
+                return ResourceManager.GetString("FLOATING_POSITION_LABEL", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to Font.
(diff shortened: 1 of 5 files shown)


JuliusSweetland:
@kmcnaught Nice PR - thanks. I'll prepare a pre-release and send you a link.

JuliusSweetland:
@kmcnaught One small thing, the styling is a bit odd on the new settings. This looks great:
image

but this is a little off...
image

kmcnaught:
Good spot, I've added a "0 decimal place" formatting to that field now in a separate PR. I'd also add spaces after the commas if I knew how, but I'm finding the C# string format documentation very opaque...

JuliusSweetland:
@kmcnaught Thanks Kirsty. That great. Do you know why the text box is slightly smaller and not aligned with the other controls?

JuliusSweetland:
@kmcnaught Oh and you could achieve the formatting with spaces, or whatever, using a value converted. You'd just need to be carefult that the conversion back to a Rect works as expected.

kmcnaught:
Yeah, I didn't really want to do the conversion myself, because then I'm responsible for any corner cases - it's currently just the built-in C# string <--> list conversion, and I'm not sure how to modify that, except it seems to correctly propagate any single-number formatting you give it.

The left-alignment just comes from `TextBox` styling - all the other fields around it are actually `NumericUpDown`s. Not sure where that styling is coming in from though, it would be nice if it aligned.

JuliusSweetland:
Got it thanks. I'll look into the styling tomorrow.

JuliusSweetland:
@kmcnaught Got it. It was just a stray Margin="5" where it shouldn't have been

PR #695, opened 2020-07-19
Crash restart

kmcnaught:
Tentative proposal for now, wouldn't mind another pair of eyes on the logic. 

2 new settings:
- `AttemptRestartUponCrash`, defaults true, might be exposed to the user in management console at some point (but I haven't added that yet)
- `CleanShutdown`, used internally by app. Set to true any time a normal shutdown/restart is requested by a user (e.g. quit buttons, quit through contextmenu, restart through contextmenu). Set to false when crash occurs, prior to the app either exiting or attempting a restart.

End result should be that by default, the app automatically restarts one time upon a crash. If it crashes again in the startup process, it won't attempt another restart. If it crashes again later during app usage (e.g. the user tries the same things again and triggers the same crash) it also won't attempt another restart. Once a full app cycle (launch -> clean shutdown) has occurred, this behaviour gets reset. 

This avoids any subtlety in "when has the app completely loaded correctly", but loses some usability in that a user could inadvertently trigger the same behaviour again, much later in their session, and then be locked out of Optikey if they don't have another way to relaunch with gaze.

If this approach is deemed acceptable, then please sanity check my logic and let me know if there are other "clean shutdown" paths that I've missed.

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Properties/Settings.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Properties/Settings.cs
@@ -2773,6 +2773,37 @@ namespace JuliusSweetland.OptiKey.Properties {
             }
         }
 
+        [global::System.Configuration.UserScopedSettingAttribute()]
+        [global::System.Diagnostics.DebuggerNonUserCodeAttribute()]
+        [global::System.Configuration.DefaultSettingValueAttribute("false")]
+        [global::System.Configuration.SettingsManageabilityAttribute(global::System.Configuration.SettingsManageability.Roaming)]
+        public virtual bool CleanShutdown
+        {
+            get
+            {
+                return ((bool)(this["CleanShutdown"]));
+            }
+            set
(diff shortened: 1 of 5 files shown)


JuliusSweetland:
@kmcnaught I think the logic is sound, but would you not need to call Settings.Default.Save(); each time after setting Settings.Default.CleanShutdown = true; ???

kmcnaught:
I'm pretty sure (but need to verify) that all clean shutdown / restart paths will trigger the shutdown handler that saves settings:

```
            //Attach shutdown handler
            Current.Exit += (o, args) =>
            {
                Log.Info("PERSISTING USER SETTINGS AND SHUTTING DOWN.");
                Settings.Default.Save();
            };
```

The unclean shutdown in the crash flow doesn't seem to.

JuliusSweetland:
@kmcnaught An excellent point! You're right. I'll merge in the change now.

Issue #697, opened 2020-07-20
Command plugin with arguments


kmcnaught:
This very helpful functionality, as documented in the wiki, allows you to launch a program / open a file.
```
<DynamicKey>
	<Label>Skype</Label>
	<Plugin>
		<Name>ExternalProgram</Name>
		<Method>RUN</Method>
		<Argument>
			<Name>command</Name>
			<Value>"C:\Program Files (x86)\Microsoft\Skype for Desktop\Skype.exe"</Value>
		</Argument>
	</Plugin>
</DynamicKey>
```

From the documentation, I was expecting this to work with commands with arguments also, for example:
```
<Value>"C:\Program Files (x86)\Google\Chrome\Application\chrome.exe" https://www.bbc.com/news</Value>
```

This does not work :( 

Looking at [the documentation for Process.Start](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.start?view=netcore-3.1), passing a single string only _starts a process resource by specifying the name of a document or application file_ and doesn't allow parameter passing, so it's not the same as running an arbitrary command from the command prompt. You can pass commandline arguments by using the `Process.Start(string, string)` overload, so I imagine this could be added to the plugin fairly easily?

kmcnaught:
Thanks for the PR @bigmutty

bigmutty:
I'm glad I could help.

Issue #691, opened 2020-07-15
Possible Bug: Sizing Issues


bigmutty:
# How OptiKey was being used

I created a Dynamic Keyboard and added a Dynamickey to navigate to the ConversationAlpha1Keyboard.

# What the user expected to happen

I expected Optikey to open the ConversationAlpha1Keyboard in fullscreen mode.

# What the user actually experienced

The ConversationAlpha1Keyboard opened in fullscreen mode for a second but then it sized the keyboard to the previous keyboard's size (docked to the left side).

# Version
 - Windows 10
 - OptiKey version 3.0.6.0

[VerticalQWERTYKeyboard.zip](https://github.com/OptiKey/OptiKey/files/4922446/VerticalQWERTYKeyboard.zip)

JuliusSweetland:
@AdamRoden Hi Adam - any idea here?

AdamRoden:
No idea, but I can take a look.

JuliusSweetland:
Thanks very much

AdamRoden:
@JuliusSweetland 
Well, this one is going to be fun. I may be a little off, but the flow goes something like this:
The resizing of the conversation keyboard from full screen to docked happens because when we leave a dynamic keyboard we always reset to the persisted window settings, which also drops a message to Windows for the appbar, which asynchronously responds triggering us to resize our window. 
To properly fix this, I think we need to tie the built-in keyboards to an enum so we know whether or not we need to register a new appbar. 
Julius, you know the built-in keyboards better than me. Is there already something I can use?

JuliusSweetland:
Possibly yes. I'm going to be busy tomorrow, but I'll take a look by the
weekend and reply again

kmcnaught:
when you look at this, can you please also bear in mind [issue 685](https://github.com/OptiKey/OptiKey/issues/685), which smells like it might have a similar cause...

AdamRoden:
@kmcnaught @JuliusSweetland I'll look at 685 as well. I have time tonight and tomorrow, so I hope to get to the bottom of both.
If all I need to worry about on 691 are the two versions of the conversation keyboard, then I think I have a solution.

JuliusSweetland:
@AdamRoden @kmcnaught Pre-release here: https://github.com/OptiKey/OptiKey/releases/tag/v3.0.6_pre-release_3

It would be great if you have a chance to test

bigmutty:
[...]

PR #699, opened 2020-07-25
Added ability to pass parameters along with the command via the External Program Plugin

bigmutty:
This is a fix to Issue #697.

### DIFF
--- a/src/JuliusSweetland.OptiKey.StandardPlugins/Native/ExternalProgram.cs
+++ b/src/JuliusSweetland.OptiKey.StandardPlugins/Native/ExternalProgram.cs
@@ -7,7 +7,8 @@ using System.Diagnostics;
  * It has only one method
  * 
  * 1) RUN
- *    . command: the complete command lina that will be executed by the plugin
+ *    . command: the command that will be executed by the plugin
+ *    . parameters: parameters to be passed along with the command
  * 
  * Please refer to OptiKey wiki for more information on registering and developing extensions.
  */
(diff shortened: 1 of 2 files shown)


src/JuliusSweetland.OptiKey.StandardPlugins/Native/ExternalProgram.cs
```diff
@@ -17,9 +18,12 @@ namespace JuliusSweetland.OptiKey.StandardPlugins
     public class ExternalProgram
     {
         // Simply run it.
-        public void RUN(string command)
+        public void RUN(string command, string parameters = "")
```
JuliusSweetland, line 21: Could we change to:
, string parameters = null

src/JuliusSweetland.OptiKey.StandardPlugins/Native/ExternalProgram.cs
```diff
@@ -17,9 +18,12 @@ namespace JuliusSweetland.OptiKey.StandardPlugins
...
         {
-            Process.Start(command);
+            if (parameters == "")
```
JuliusSweetland, line 23: Could we change to:
if (string.IsNullOrEmpty(parameters))

JuliusSweetland:
@bigmutty Hi Daryl. Thanks for this - could we might 2 minor tweaks and then it's good to go.

bigmutty:
Ok, I made those changes. Thanks for catching them.

JuliusSweetland:
@bigmutty thanks. Could you also send over your dynamic keyboards to recreate #691 and I'll look at both tomorrow evening? Thx

PR #708, opened 2020-08-24
Resizing fixes

kmcnaught:
This is a proposed edit to PR#707

As far as I can tell, the changes to `RestorePersistedState` (both the function and the calling code) do not break anything in the resizing logic, but removing the `PersistSizeAndPosition` calls within `Maximise` and `Minimise` do cause problems - they are there to make sure any pending manual changes to size/position aren't lost when you minimise / re-maximise. 

I've reverted half of @AdamRoden's commit and it still seems to fix the issue with the Minimised size decreasing each time, but without breaking the resizing. If @AdamRoden's happy after testing this, I'm happy with it.


AdamRoden:
I'll test later, but my first impression is that this will break.

kmcnaught:
@AdamRoden I'm afraid I got slightly lost in the thread of issue 691, there were a few different issues mentioned at different points. Which issue has reappeared? Do you see the minimised size shrinking every time you minimise from Daryl's keyboard?

kmcnaught:
Ah, I see the problem now - the minimised size was shrinking once and then staying stable. I'll make another PR with an alternative suggestion

Issue #711, opened 2020-08-30
Mouse magnifier does not center on selection point in v3.0.6.0


AdamRoden:
Upgraded from 3.05 to 3.06 and now the magnify pop-up window is always appearing offset to the right. This makes the feature unusable.

AdamRoden:
I could be wrong, but I suspect this is somehow related to the modifications to allow manual resizing. @kmcnaught do you have any thoughts?

kmcnaught:
Okay, this is a _really weird_ one. @JuliusSweetland for some rubber ducking, in the hope you have a clearer idea of how the MagnifyPopup's layout works. 

This bug was introduced in 61800d423a, where I do nothing more than occasionally turn off the main window's resize grips so I can avoid the logic of "what does the user mean when they resize in this minimised/collapsed state". (note that the addition of this line is irrelevant to the issue:  `    if (getWindowState() != WindowStates.Docked) return;` )

The behaviour is such that if you have not ever minimised or collapsed Optikey this session, the mouse magnification will work correctly. As soon as the resize grips are turned off (by minimising or collapsing), the magnify popup ends up translated to the right. I believe that the translation is equal to the width of the popup (unless it gets curtailed by the edge of the screen). Most importantly, **none of the numbers in MagnifyPopup::SetSizeAndPosition are wrong.** While testing I've hardcoded a source point for consistent behaviour, and I cannot tell any difference at all between the properties of the MagnifyPopup in the case where it appears in the right place, and the case where it's offset. It just seems like changing the parent's ResizeMode has somehow changed the MagnifyPopup's sizing behaviour.

JuliusSweetland:
What happens if the popup code temporarily sets the resize mode before attempting to size and position itself? Does that fix the behaviour?

kmcnaught:
No, that doesn't help, because it's not the current resize mode that's the problem, but rather "has the resize mode ever been set to NoResize before this point"

kmcnaught:
I'm increasingly getting deja vu with the unexplained behaviour in https://github.com/OptiKey/OptiKey/issues/670 . I'm wondering if there's something funky going on with the HorizontalOffset, which (a) causes #670 and (b) is somehow affected by changing the resize mode.

[...]

Issue #670, opened 2020-04-30
Mouse magnification step appears on wrong monitor sometimes


kmcnaught:
Yes, I know we don't _officially_ support multimonitor setups, but we're close enough that I think this bug is worth attempting to fix at some point (happy for it to go on my stack but I might not get round to it for a while). 

Setup is: 
- two monitors, the right hand monitor is primary, and both Tobii eye tracking and Optikey Pro (3.0.5) are running on the right hand primary monitor. 
- Mouse magnification is turned on, with magnification destination size 60% (this might be bigger than default?) 
- I use "left click at point" and the magnified area _sometimes_ appears on the left hand monitor, which I am unable to track gaze on. 

video demo here:
https://www.youtube.com/watch?v=ysc6Yw6IoSU&feature=youtu.be

kmcnaught:
I had a brief look at this last week, so I'm commenting mainly as a message to my future self...

In `MagnifyPopup::SetSizeAndPosition`, there are a bunch of coordinate transforms making use of `Window::GetTransformFromDevice`. I haven't been able to pinpoint where any of these numbers are wrong, but I suspect they are (with my particular monitor setup at least!).

The behaviour I'm observing is that if my first click (to pick the magnified area) is at the left edge of the screen, but in the middle vertically, it gets positioned correctly. If I move up toward the top of the screen, everything is fine until it gets to the point where the magnifypopup ends up butting up against the top of the screen. If I go any higher, the popup gets requested "too high" and then gets auto-pushed onto the left monitor. This suggests that the inputs into this bit of logic are wrong:

```
    if (pointInWpfCoords.Y + (height / 2d) > screenHeight)
        distanceFromTopBoundary = screenHeight + (height / 2d);
    else
        distanceFromTopBoundary = pointInWpfCoords.Y - (height / 2d);
```

It's also worth noting that my monitors are setup so that the left (secondary) monitor is higher resolution than the right (primary) monitor, *and* it's a high-DPI Surface Pro. So either it's a simple issue of somewhere overestimating the screen height and allowing the popup to overflow, or it's a trickier issue with combined high-DPI / normal-DPI screens. 

Next steps:
- Interrogate the logic above again, see if I can find any numbers that are wrong!
- Look into f8505c2 in case there's an interaction there

bigmutty:
I found a temporary workaround. If you enable "Show magnifier window at center of screen," the issue is nonexistent.

JuliusSweetland:
[...]

PR #712, opened 2020-08-31
Use Position=Absolute for magnify popup

kmcnaught:
This seems to fix both #711 and #670 and makes the popup placement much easier to reason about. Having said that, I still don't understand why there was an interaction between `Placement=AbsolutePoint` and `ResizeMode`, so this PR should face some testing/scrutiny. I'm also unsure whether there was some extra reasoning behind using `AbsolutePoint` in the first place. You can read the explanations at [Placement Mode enums](https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.primitives.placementmode?view=netcore-3.1#System_Windows_Controls_Primitives_PlacementMode_Absolute).

I've tested on my Surface Pro with (a) extended desktop, (b) hi-DPI surface screen only, (c) normal-DPI monitor only and (d) surface screen in "tablet mode".

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/UI/Controls/MagnifyPopup.cs
+++ b/src/JuliusSweetland.OptiKey.Core/UI/Controls/MagnifyPopup.cs
@@ -131,21 +131,19 @@ namespace JuliusSweetland.OptiKey.UI.Controls
             MaxWidth = MinWidth = Width = width;
             MaxHeight = MinHeight = Height = height;
 
+            // Screen-centered version (TODO: use Popup Placement="Centre" instead?)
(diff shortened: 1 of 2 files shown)


AdamRoden:
I'll deploy the code and test on my Surface Book.

JuliusSweetland:
@AdamRoden @kmcnaught If you've both tested then I'll release 3.0.7 as a full release for everyone

AdamRoden:
@JuliusSweetland Works great!

JuliusSweetland:
Thanks Adam! Appreciate the speedy testing and feedback.

PR #718, opened 2020-09-09
Don't save window size if not initialised!

### LINKED ISSUE #713: MenuKeyboard action shows a tiny black square in the middle of the screen.
I hope this isn't just me again.

I updated to 3.0.7 and now when I select the MenuKeyboard action on my custom dynamic keyboard, I just get a tiny black square in the middle of the screen. There's no way to recover from this other than restarting OptiKey.

I just checked the wiki but didn't notice anything that might have changed.

Windows 7
Tobii 4C

I've been using floating keyboard setting instead of docked. Maybe that's related.

[FinnishKeyboard.txt](https://github.com/OptiKey/OptiKey/files/5178451/FinnishKeyboard.txt)

[user.config.txt](https://github.com/OptiKey/OptiKey/files/5177763/user.config.txt)

kmcnaught:
Fixes #713, but I haven't tested all the other dynamic / resizing / startup combinations.


JuliusSweetland:
@AdamRoden @kmcnaught Thoughts on these two PRs? They appear to be trying to implement the same sort of idea (exactly the same idea?) in different ways: #717 #718

AdamRoden:
@JuliusSweetland It's the same fix, but in mine, I moved the check a level deeper to be more inclusive in case another method makes the call before initializing.

kmcnaught:
Yeah, what he said^ :)

PR #721, opened 2020-09-10
Replace Kalman filter with custom logic

AdamRoden:
@kmcnaught @JuliusSweetland This is the first step toward replacing the Kalman filter. Small movements are heavily smoothed and large movements are lightly smoothed. The values work great on my Surface Book display, but I don't know if the same will be true of other devices. It would be good to hear from you on how it feels.
If you guys like how this performs, then we can continue forward with this and work on the other changes we've discussed.

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/DataFilters/KalmanFilter.cs
+++ b/src/JuliusSweetland.OptiKey.Core/DataFilters/KalmanFilter.cs
@@ -19,8 +19,9 @@ namespace JuliusSweetland.OptiKey.DataFilters
 
         public double Update(double measurement)
         {
-            Gain = (EstimationConfidence + ProcessNoise) / (EstimationConfidence + ProcessNoise + MeasurementNoise);
-            EstimationConfidence = MeasurementNoise * (EstimationConfidence + ProcessNoise) / (MeasurementNoise + EstimationConfidence + ProcessNoise);
+            //The Kalman filter was not delivering the desired result so it has been replaced with a custom calculation
+            Gain = System.Math.Abs(measurement - EstimatedValue) > 500
+                ? .5 : .001 + System.Math.Abs(measurement - EstimatedValue) / 1000d;
             double result = EstimatedValue + (measurement - EstimatedValue) * Gain;
             EstimatedValue = result;
             return result;


JuliusSweetland:
Which Tobii data stream settings do you recommend testing this with? The
raw stream, or an already smoothed stream?

I'll check it out tonight

kmcnaught:
I'd be much happier to stick with a kalman filter that correctly downweights smoothing of larger movements by having the right process model. This can achieve the same desired outcome of only smoothing small movements, but while staying principled and allowing for future modifications within the KF framework. 

I spent a couple of days looking into this and have a reasonable proposal. I can dig it out and share, I just hadn't yet because I wasn't sure how we'd best tune it and didn't want to jump the gun.

JuliusSweetland:
Is there any reason why we couldn't have both smoothing options available?
Users could try combinations of settings to get the right feel?

kmcnaught:
Not really, except that we're already struggling from a combinatorial explosion of options that can be scary to a user. This one definitely shouldn't be called a Kalman Filter though :)

Since any approach is subject to tuning/customisation, I was finding it really hard to judge any one solution just by feel (though obviously the final test is whatever feels best, when everything is tuned finally). This is why I spent a couple of days getting data and deriving a more appropriate model. This change will certainly be an improvement from what we have now, but I'd prefer if it were compared against a proper KF model seeing as I've already put the work in!

(btw, I'm still 100% in favour of calling any of these options "smoothing" in user-facing places)

JuliusSweetland:
Let us know when your KF changes are ready and we can test and then make a
decision from there

AdamRoden:
@kmcnaught I would definitely like to see what you have, because I was unable to get appropriate behavior using Kalman, no matter the settings.
Regarding my alternative, I've got another version where I created a new class with the ability to specify the radius for a dead zone, the distance over which the effect fades, and the level of smoothing from 0 to 100%.

AdamRoden:
@JuliusSweetland @kmcnaught I spent another hour playing with my code, and I can say without question that what I've come up with is better than anything I've seen from Tobii, Smartbox, or anyone else. My precision and speed are near par with traditional mouse input devices. The Magnetic cursor with a switch is now magnificent! It's like I just added magic sauce to my eye gaze! This is a game-changer.

JuliusSweetland:
Did you close this by mistake?

AdamRoden:
No, I did it on purpose. I'm resetting my code and going to share my new
version.

Adam Roden

PR #723, opened 2020-09-11
Smoothing Concept

AdamRoden:
@JuliusSweetland @kmcnaught 
At this point, this is more of a concept than a finished product; meant to show how we could implement it and for you to be able to try it out and see how well it works.

First, I piggybacked off the data stream processing level setting to also impact my smoothing level. Then, I repurposed the Point Selection's Lock On and Fixation Radius values to be used in my algorithm. Lock On is now the "deadzone" that will freeze the gaze point on your screen and hold it still. Fixation is now the distance over which the gaze point on your screen will accelerate from zero up to the full-speed of your eye movement. By adjusting the values you can get a very slow, steady cursor that you could trace with, or a swift but controlled cursor for rapid cursor movement. I recommend using the magnetic cursor to test; with the settings tuned properly, it is oddly satisfying. 

In this figure, you will see the optimal initial Pointing settings, Point Lock On and Completion times, and Point Lock On and Fixation radius.

image

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Extensions/DoubleExtensions.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Extensions/DoubleExtensions.cs
@@ -14,5 +14,10 @@ namespace JuliusSweetland.OptiKey.Extensions
         {
             return Math.Max(value, lowerLimit);
         }
+
+        public static double Clamp(this double value, double lowerLimit, double upperLimit)
+        {
+            return Math.Min(Math.Max(value, lowerLimit), upperLimit);
+        }
     }
 }
--- a/src/JuliusSweetland.OptiKey.Core/UI/Controls/MagnifyPopup.cs
+++ b/src/JuliusSweetland.OptiKey.Core/UI/Controls/MagnifyPopup.cs
@@ -72,7 +72,7 @@ namespace JuliusSweetland.OptiKey.UI.Controls
 
                         try
                         {
-                            DisplayScaledScreenshot(sourcePoint.Value);
+                            DisplayScaledScreenshot();
                         }
                         catch (System.ComponentModel.Win32Exception ex)
                         {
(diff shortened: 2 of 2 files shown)


kmcnaught:
This sounds great Adam. It sounds like your solution is very similar to the non-linear KF process model I'd derived which also smoothly interpolates between nearby smoothing and faraway no-smoothing but using a mathematical model. I was going to write up my results today but to be honest I don't have the capacity to be racing/competing on this as it seems it will be duplicate effort. Apologies if I didn't explain well enough what I'd achieved. I look forward to trying your solution out when it's complete. 

A couple of things I was planning to look into that you might want to take into account - firstly, the current filter is one dimensional, so it would work better to have a class that operates on 2D points, so it behaves equivalently for diagonal and axis-aligned movements. This shouldn't be hard to implement, you'd just need to combine the 2 calculations with a euclidean distance. Also, I was wondering whether settings for things like "fixation radius" should be defined as %screen rather than pixels. The correct answer would be "degrees of visual angle" but we obviously don't know how this relates to pixels! I'd imagine that screen size in degrees for a "normal, comfortable setup" is more consistent than pixels, which could easily be halved or doubled just by varying screen resolution.

AdamRoden:
I agree with all that Kirsty. I'm still interested to see your model. There
may be a benefit in combining it with my model. The true eureka moment in
my model wasn't so much my formula, rather it was when I stopped trying to
be purely mathematical and incorporated variables for the deadzone (Lock On
Radius), and the damping zone (Fixation Radius). To mark when to completely
freeze and completely release the cursor.

One thing I would still like to add is special logic when your gaze is on a
key. That way you could have a really slow tracking mouse combined with a
snappy keyboard. We could explore things like having the gaze point jump to
the center of keys rather than slowly drifting over them.

AdamRoden:
@kmcnaught @JuliusSweetland I've made additional improvements. I reverted the previous changes to the KalmanFilter class and went ahead with creating a new GazeFilter class. I took Kirsty's advice to use 2D points, and I was able to add awareness of when you are looking at a key so that you can combine a slow mouse simulation without harming your ability to have a snappy keyboard. This is really great stuff! Enjoy!

JuliusSweetland:
@AdamRoden I've just tried the new filter with your screenshot settings and it works well, but to be honest I don't use eye tracking enough to be able to feel the difference. It certainly feels good.

The only issue I had was when using the magnifier I couldn't make small adjustments to my final selection, e.g. if I almost hit where I wanted, but actually I wanted to be slightly to the right, the selection would complete in the original position instead.

src/JuliusSweetland.OptiKey.Core/OptiKeyApp.cs
```diff
@@ -785,7 +785,7 @@ protected static ICalibrationService CreateCalibrationService()
                         Settings.Default.PointSelectionTriggerFixationLockOnTime,
                         Settings.Default.PointSelectionTriggerFixationCompleteTime,
                         Settings.Default.PointSelectionTriggerLockOnRadiusInPixels,
-                        Settings.Default.PointSelectionTriggerFixationRadiusInPixels,
+                        Settings.Default.PointSelectionTriggerLockOnRadiusInPixels,
```
JuliusSweetland, line 788: This doesn't look right - you're passing the lock on radius (px) twice. The signature being called is:

public PointFixationSource(
            TimeSpan lockOnTime,
            TimeSpan timeToCompleteTrigger,
            double lockOnRadius,
            double fixationRadius,
            IPointSource pointSource)

AdamRoden:
> @AdamRoden I've just tried the new filter with your screenshot settings and it works well, but to be honest I don't use eye tracking enough to be able to feel the difference. It certainly feels good.
> 
> The only issue I had was when using the magnifier I couldn't make small adjustments to my final selection, e.g. if I almost hit where I wanted, but actually I wanted to be slightly to the right, the selection would complete in the original position instead.

@JuliusSweetland I am so sorry for having wasted your time with something I broke. Just before publishing my revised model, I reverted code, but there was one line that was needed. I've re-implemented that line and when you test again you will get the correct behavior. You can play with your lock and fixation radius values to what works for you. Use a large fixation radius for a very slow cursor.

JuliusSweetland:
> 
> 
> > @AdamRoden I've just tried the new filter with your screenshot settings and it works well, but to be honest I don't use eye tracking enough to be able to feel the difference. It certainly feels good.
> > The only issue I had was when using the magnifier I couldn't make small adjustments to my final selection, e.g. if I almost hit where I wanted, but actually I wanted to be slightly to the right, the selection would complete in the original position instead.
> 
> @JuliusSweetland I am so sorry for having wasted your time with something I broke. Just before publishing my revised model, I reverted code, but there was one line that was needed. I've re-implemented that line and when you test again you will get the correct behavior. You can play with your lock and fixation radius values to what works for you. Use a large fixation radius for a very slow cursor.

@AdamRoden Don't worry! It wasn't a waste of time at all. I've added a comment concerning your last change. Can you check you meant to change the args passed to the PointFixationSource class ctor and let me know before I retest?

src/JuliusSweetland.OptiKey.Core/OptiKeyApp.cs
AdamRoden: @JuliusSweetland I will implement this differently in my final version, but for now, just know that it is on purpose to get the correct experience.

JuliusSweetland:
@AdamRoden I've just tried it again - feels really good. Nothing to do with your code, but the ET5 sometimes has issues after re-calibrating (via Optikey's menu) where the tracker seems to turn off and needs to be disconnected/reconnected. Disappointing!

AdamRoden:
@JuliusSweetland Regarding the ET5, I have the same calibration issue even when not running Optikey. I can only recalibrate if I have someone nearby to disconnect/reconnect for me. Also, ET5 does not have any of the improvements that TobiiDynavox added in their I-13 and I-16  to track eyes outside.

JuliusSweetland:
Yes it feels like it's actually a step backward from the 4C.

What were the improvements in the I series?

AdamRoden:
> Yes it feels like it's actually a step backward from the 4C. What were the improvements in the I series?

@JuliusSweetland The new I-Series are supposed to be capable of working outside in sunlight, which is simply not possible with most eye trackers.

src/JuliusSweetland.OptiKey.Core/Observables/PointSources/PointServiceSource.cs
```diff
@@ -59,8 +58,8 @@ public IObservable<Timestamped<PointAndKeyValue>> Sequence
                             eh => pointGeneratingService.Point -= eh)
                         .Where(_ => State == RunningStates.Running)
                         .Select(ep => Settings.Default.KalmanFilterEnabled && pointGeneratingService.KalmanFilterSupported
-                            ? new Timestamped<Point>(new Point((int)kalmanFilterX.Update(ep.EventArgs.Value.X), (int)kalmanFilterY.Update(ep.EventArgs.Value.Y)), ep.EventArgs.Timestamp)
-                            : ep.EventArgs
+                            ? new Timestamped<Point>(kalmanFilter.Update(ep.EventArgs.Value, ep.EventArgs.Value.ToKeyValue(PointToKeyValueMap)), ep.EventArgs.Timestamp)
+                            : new Timestamped<Point>(gazeFilter.Update(ep.EventArgs.Value, ep.EventArgs.Value.ToKeyValue(PointToKeyValueMap)), ep.EventArgs.Timestamp)
```
JuliusSweetland, line 62: @AdamRoden This looks like the Kalman filter will be used if enabled, but if not then the GazeFilter will ALWAYS be used. Is that the intention? What do you think about the Management Console setting being something like "Additional smoothing" : Kalman, Adam's Filter, None?

AdamRoden:
@JuliusSweetland @kmcnaught I just added another commit which adds my special sauce to the Kalman Filter so that it operates in accordance with the lock radius, fixation radius, and keyboard snappiness concepts. Now, if you enable the Kalman Filter it will be used, and if you disable it, my new GazeFilter will be used. It is the best of both worlds and opens the door to using what Kirsty built if it's worth it.

src/JuliusSweetland.OptiKey.Core/Observables/PointSources/PointServiceSource.cs
AdamRoden: @JuliusSweetland For the final solution, we'll want to regroup some settings and decide what is best. As Kirsty said, we don't want to overwhelm users unnecessarily. So, we can offer all the options, but we'll need to make a call and put our best foot forward and denote the Default values if users start playing around.

AdamRoden:
I'm closing this PR so I can submit a change to MagnifyPopUp. After that change is merged I will submit a final version of Gaze Smoothing.

PR #726, opened 2020-09-18
Gaze smoothing

AdamRoden:
I decided to both, enhance the Kalman Filter and make Gaze Smoothing adjustable with or without the KF. So, it is all built into the KF class, but the algorithm is slightly different depending on whether or not KF is enabled. 

image

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Extensions/DoubleExtensions.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Extensions/DoubleExtensions.cs
@@ -14,5 +14,10 @@ namespace JuliusSweetland.OptiKey.Extensions
         {
             return Math.Max(value, lowerLimit);
         }
+
+        public static double Clamp(this double value, double lowerLimit, double upperLimit)
+        {
+            return Math.Min(Math.Max(value, lowerLimit), upperLimit);
+        }
     }
 }
--- a/src/JuliusSweetland.OptiKey.Core/GazeFilter.cs
+++ b/src/JuliusSweetland.OptiKey.Core/GazeFilter.cs
@@ -1,0 +1,55 @@
+﻿// Copyright (c) 2020 OPTIKEY LTD (UK company number 11854839) - All Rights Reserved
+using System.Windows;
+using JuliusSweetland.OptiKey.Models;
+using JuliusSweetland.OptiKey.Properties;
+
+namespace JuliusSweetland.OptiKey.DataFilters
+{
+    /// <summary>
+    /// For a continuous stream of points and optional KeyValues the GazeFilter 
+    /// will apply logic intended to smooth the inputs and provide silky smooth 
+    /// positioning in accordance with settings configurable by the user
+    /// </summary>
+    public class GazeFilter
+    {
+        public double DampingLevel { get; set; } = 1f; // from .01 to 1. rate at which the point moves in accordance with the gaze
+        public double FixationRadius { get; set; } = 0f; // the "gravity well" where damping varies
+        public double LockRadius { get; set; } = 0f; // the distance within which the point will not move
+        private Point LastPoint;
+
+        public GazeFilter()
+        {
+            var level = Settings.Default.TobiiEyeXProcessingLevel; 
+            this.DampingLevel = level == Enums.DataStreamProcessingLevels.High
+                ? .1f : level == Enums.DataStreamProcessingLevels.Medium
+                ? .4f : level == Enums.DataStreamProcessingLevels.Low
+                ? .7f : 1;
+            this.FixationRadius = Settings.Default.PointSelectionTriggerFixationRadiusInPixels;
+            this.LockRadius = Settings.Default.PointSelectionTriggerLockOnRadiusInPixels;
+            this.LastPoint = new Point();
+        }
+
+
+        /// <summary>
+        /// If point is on a key then prioritize speed over smoothing.
+        /// Else if inside the lock radius then do not move.
+        /// Else if inside the fixation radius then prioritize smoothing.
+        /// Else if outside the fixation radius then balance speed with smoothing.
+        /// </summary>
+        public Point Update(Point point, KeyValue keyValue = null)
+        {
+
+            Vector delta = point - LastPoint;
+            double multiplier = keyValue != null
+                    ? .5 * (1 + DampingLevel) 
+                : delta.Length > FixationRadius 
+                    ? .2 * (1 + 4 * DampingLevel) 
+                : delta.Length > LockRadius
+                    ? DampingLevel * (System.Math.Pow(delta.Length, 1 + DampingLevel) / System.Math.Pow(FixationRadius, 1 + DampingLevel))
+                : 0;
+            Point result = LastPoint + multiplier * (point - LastPoint);
+            LastPoint = result;
+            return result;
+        }
+    }
+}
--- a/src/JuliusSweetland.OptiKey.Core/JuliusSweetland.OptiKey.Core.csproj
+++ b/src/JuliusSweetland.OptiKey.Core/JuliusSweetland.OptiKey.Core.csproj
@@ -258,6 +258,7 @@
     <Compile Include="Extensions\SvgViewboxAttachedProperties.cs" />
     <Compile Include="Extensions\UiElementExtensions.cs" />
     <Compile Include="Extensions\WindowExtensions.cs" />
+    <Compile Include="GazeFilter.cs" />
     <Compile Include="Models\DynamicKeyboardFolder.cs" />
     <Compile Include="Models\ILastMouseActionStateManager.cs" />
     <Compile Include="Models\KeyboardOutputServiceState.cs" />
--- a/src/JuliusSweetland.OptiKey.Core/Observables/PointSources/PointServiceSource.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Observables/PointSources/PointServiceSource.cs
@@ -19,8 +19,7 @@ namespace JuliusSweetland.OptiKey.Observables.PointSources
         
         private readonly TimeSpan pointTtl;
         private readonly IPointService pointGeneratingService;
-        private readonly KalmanFilter kalmanFilterX;
-        private readonly KalmanFilter kalmanFilterY;
+        private readonly GazeFilter gazeFilter;
 
         private IObservable<Timestamped<PointAndKeyValue>> sequence;
 
@@ -34,10 +33,7 @@ namespace JuliusSweetland.OptiKey.Observables.PointSources
         {
             this.pointTtl = pointTtl;
             this.pointGeneratingService = pointGeneratingService;
-            this.kalmanFilterX = new KalmanFilter(Settings.Default.KalmanFilterInitialValue, Settings.Default.KalmanFilterConfidenceOfInitialValue, 
-                Settings.Default.KalmanFilterProcessNoise, Settings.Default.KalmanFilterMeasurementNoise);
-            this.kalmanFilterY = new KalmanFilter(Settings.Default.KalmanFilterInitialValue, Settings.Default.KalmanFilterConfidenceOfInitialValue, 
-                Settings.Default.KalmanFilterProcessNoise, Settings.Default.KalmanFilterMeasurementNoise);
+            this.gazeFilter = new GazeFilter();
         }
 
         #endregion
@@ -59,7 +55,7 @@ namespace JuliusSweetland.OptiKey.Observables.PointSources
                             eh => pointGeneratingService.Point -= eh)
                         .Where(_ => State == RunningStates.Running)
                         .Select(ep => Settings.Default.KalmanFilterEnabled && pointGeneratingService.KalmanFilterSupported
-                            ? new Timestamped<Point>(new Point((int)kalmanFilterX.Update(ep.EventArgs.Value.X), (int)kalmanFilterY.Update(ep.EventArgs.Value.Y)), ep.EventArgs.Timestamp)
+                            ? new Timestamped<Point>(gazeFilter.Update(ep.EventArgs.Value, ep.EventArgs.Value.ToKeyValue(PointToKeyValueMap)), ep.EventArgs.Timestamp)
                             : ep.EventArgs
                         )
                         .PublishLivePointsOnly(pointTtl)
--- a/src/JuliusSweetland.OptiKey.Core/OptiKeyApp.cs
+++ b/src/JuliusSweetland.OptiKey.Core/OptiKeyApp.cs
@@ -785,7 +785,7 @@ namespace JuliusSweetland.OptiKey
                         Settings.Default.PointSelectionTriggerFixationLockOnTime,
                         Settings.Default.PointSelectionTriggerFixationCompleteTime,
                         Settings.Default.PointSelectionTriggerLockOnRadiusInPixels,
-                        Settings.Default.PointSelectionTriggerFixationRadiusInPixels,
+                        Settings.Default.PointSelectionTriggerLockOnRadiusInPixels,
                         pointSource);
                     break;
 
--- a/src/JuliusSweetland.OptiKey.Core/UI/Controls/MagnifyPopup.cs
+++ b/src/JuliusSweetland.OptiKey.Core/UI/Controls/MagnifyPopup.cs
@@ -2,7 +2,6 @@
 using System;
 using System.Drawing;
 using System.Linq;
-using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using System.Windows;
 using System.Windows.Controls.Primitives;
@@ -72,7 +71,7 @@ namespace JuliusSweetland.OptiKey.UI.Controls
 
                         try
                         {
-                            DisplayScaledScreenshot(sourcePoint.Value);
+                            DisplayScaledScreenshot();
                         }
                         catch (System.ComponentModel.Win32Exception ex)
                         {
@@ -122,36 +121,55 @@ namespace JuliusSweetland.OptiKey.UI.Controls
 
         private void SetSizeAndPosition(Point point)
         {
+            point = new Point(point.X.Clamp(screen.Bounds.Left, screen.Bounds.Right), point.Y.Clamp(screen.Bounds.Top, screen.Bounds.Bottom));
+
+            var centerOnScreen = Settings.Default.MagnifierCenterOnScreen;
+            var magnifySourcePercentage = Settings.Default.MagnifySourcePercentageOfScreen / 100d;
+            var destinationPercentage = Settings.Default.MagnifyDestinationPercentageOfScreen / 100d;
+
+            var captureWidth = magnifySourcePercentage * screen.Bounds.Width;
+            var captureHeight = magnifySourcePercentage * screen.Bounds.Height;
+            var captureX = (point.X - (captureWidth / 2d)).Clamp(screen.Bounds.Left, screen.Bounds.Right - captureWidth);
+            var captureY = (point.Y - (captureHeight / 2d)).Clamp(screen.Bounds.Top, screen.Bounds.Bottom - captureHeight);
+
+            //Calculate source area
+            sourceArea = new Rect(captureX, captureY, captureWidth, captureHeight);
+
+            //Get WPF coords in order to calculate the magnified size and position
+            var pointInWpfCoords = window.GetTransformFromDevice().Transform(new Point(point.X, point.Y));
             var screenTopLeftInWpfCoords = window.GetTransformFromDevice().Transform(new Point(screen.Bounds.Left, screen.Bounds.Top));
             var screenBottomRightInWpfCoords = window.GetTransformFromDevice().Transform(new Point(screen.Bounds.Right, screen.Bounds.Bottom));
 
             var screenWidth = (screenBottomRightInWpfCoords.X - screenTopLeftInWpfCoords.X);
             var screenHeight = (screenBottomRightInWpfCoords.Y - screenTopLeftInWpfCoords.Y);
-
-            var destinationPercentage = Settings.Default.MagnifyDestinationPercentageOfScreen / 100d;
-            var centerOnScreen = Settings.Default.MagnifierCenterOnScreen;
-
             var width = destinationPercentage * screenWidth;
             var height = destinationPercentage * screenHeight;
 
-            MaxWidth = MinWidth = Width = width;
-            MaxHeight = MinHeight = Height = height;
-
-            // Screen-centered version (TODO: use Popup Placement="Centre" instead?)
+            // Screen-centered version
             var distanceFromLeftBoundary = screenTopLeftInWpfCoords.X + ((1d - destinationPercentage) / 2d) * screenWidth;
             var distanceFromTopBoundary = screenTopLeftInWpfCoords.Y + ((1d - destinationPercentage) / 2d) * screenHeight;
-            var pointInWpfCoords = window.GetTransformFromDevice().Transform(new Point(point.X, point.Y));
 
             // Point-centered version 
             if (!centerOnScreen)
             {
-                distanceFromLeftBoundary = (pointInWpfCoords.X - (width / 2d)).CoerceToLowerLimit(0);
-                distanceFromLeftBoundary.CoerceToUpperLimit(screenWidth - width);
+                var widthCropped = Math.Min(pointInWpfCoords.X - screenTopLeftInWpfCoords.X, screenBottomRightInWpfCoords.X - pointInWpfCoords.X)
+                    .CoerceToUpperLimit(width / 2d) + width / 2d;
+                var heightCropped = Math.Min(pointInWpfCoords.Y - screenTopLeftInWpfCoords.Y, screenBottomRightInWpfCoords.Y - pointInWpfCoords.Y)
+                    .CoerceToUpperLimit(height / 2d) + height / 2d;
 
-                distanceFromTopBoundary = (pointInWpfCoords.Y - (height / 2d)).CoerceToLowerLimit(0);
-                distanceFromTopBoundary.CoerceToUpperLimit(screenHeight - height);
+                var minSize = destinationPercentage - .6 * Math.Pow(destinationPercentage, 2);
+                destinationPercentage = (widthCropped / screenWidth).Clamp(minSize, destinationPercentage);
+                destinationPercentage = (heightCropped / screenHeight).Clamp(minSize, destinationPercentage);
+
+                width = destinationPercentage * screenWidth;
+                height = destinationPercentage * screenHeight;
+
+                distanceFromLeftBoundary = (pointInWpfCoords.X - (width / 2d) - screenTopLeftInWpfCoords.X).CoerceToLowerLimit(0);
+                distanceFromTopBoundary = (pointInWpfCoords.Y - (height / 2d) - screenTopLeftInWpfCoords.Y).CoerceToLowerLimit(0);
             }
 
+            MaxWidth = MinWidth = Width = width;
+            MaxHeight = MinHeight = Height = height;
             HorizontalOffset = distanceFromLeftBoundary;
             VerticalOffset = distanceFromTopBoundary;
 
@@ -162,39 +180,14 @@ namespace JuliusSweetland.OptiKey.UI.Controls
 
         #region Capture & Display Scaled Screenshot
 
-        private void DisplayScaledScreenshot(Point point)
+        private void DisplayScaledScreenshot()
         {
-            var bitmap = CaptureScreenshot(point);
+            var bitmap = CaptureScreenshot();
             Child = new Image { Source = bitmap.ToBitmapImage() };
         }
 
-        private Bitmap CaptureScreenshot(Point point)
+        private Bitmap CaptureScreenshot()
         {
-            var magnifySourcePercentage = Settings.Default.MagnifySourcePercentageOfScreen / 100d;
-
-            var captureWidth = magnifySourcePercentage * screen.Bounds.Width;
-            captureWidth = captureWidth.CoerceToUpperLimit(screen.Bounds.Width);
-
-            var captureHeight = magnifySourcePercentage * screen.Bounds.Height;
-            captureHeight = captureHeight.CoerceToUpperLimit(screen.Bounds.Height);
-
-            var captureX = point.X - (captureWidth / 2d);
-            captureX = captureX.CoerceToLowerLimit(screen.Bounds.Left);
-            if (captureX + captureWidth > screen.Bounds.Right)
-            {
-                captureX = screen.Bounds.Right - captureWidth;
-            }
-
-            var captureY = point.Y - (captureHeight / 2d);
-            captureY = captureY.CoerceToLowerLimit(screen.Bounds.Top);
-            if (captureY + captureHeight > screen.Bounds.Bottom)
-            {
-                captureY = screen.Bounds.Bottom - captureHeight;
-            }
-
-            //Calculate source area
-            sourceArea = new Rect(captureX, captureY, captureWidth, captureHeight);
-
             if (sourceArea.Width <= 0)
                 throw new ArgumentOutOfRangeException(nameof(sourceArea.Width), $"SourceArea.Width was {sourceArea.Width}, which is invalid");
 
@@ -202,7 +195,7 @@ namespace JuliusSweetland.OptiKey.UI.Controls
                 throw new ArgumentOutOfRangeException(nameof(sourceArea.Height), $"SourceArea.Height was {sourceArea.Height}, which is invalid");
 
             //Create the bitmap to copy the screen shot into
-            var bitmap = new Bitmap(Convert.ToInt32(captureWidth), Convert.ToInt32(captureHeight));
+            var bitmap = new Bitmap(Convert.ToInt32(sourceArea.Width), Convert.ToInt32(sourceArea.Height));
 
             //Copy the screen image to the bitmap
             using (var graphic = Graphics.FromImage(bitmap))


JuliusSweetland:
@AdamRoden Looks good although I think the settings are a bit difficult to understand. If the Kalman filter is not enabled, should the smoothing level be available or hidden? If it is enabled then what does a low or high smoothing level actually mean?

AdamRoden:
It's more like Smoothing level preempts Kalman Filter. I guess in MC it
should go above it. You can turn on KF in addition to Smoothing.
Level 1 is a fast cursor, level 9 is a slow cursor. I guess it is a lot
like Windows how you can adjust your mouse sensitivity. Though, I don't
know if 0-9 the way I did it is user friendly.

Adam Roden

JuliusSweetland:
Ok makes sense. Is there a way to convey that concept of fast/slow and
sensitive/insensitive?

What does the KF do in addition to the smoothing? It's difficult to know
what change I would might be making with these settings.

AdamRoden:
What if I relabel Gaze Smoothing as "Cursor Sensitivity" and turn it into a drop-down of:
0 - Realtime
1 
2 
3
4
5 - Slow Motion

I don't know what to say about KF, other than it is an alternate version of smoothing. It's like suggestion method basic vs ngram vs presage.

JuliusSweetland:
I like realtime to slow motion, but cursor sensitivity implies it's only
going to affect mouse control. We'd need something like eye movement
sensitivity?

KF it's tricky...

AdamRoden:
How about this?

image

JuliusSweetland:
Better. What do you think of:

Source:
Source smoothing amount (before optikey):
Additional smoothing amount (within optikey):
Additional Kalman Filter smoothing (within optikey):

I still think the last two are not clear. I wouldn't, as a user, know what
the difference between them is and how they might combine. I can't remember
if the final/current state is to have to smoothing classes (gaze smoothing
and KF) or if they've been combined? If they are combined then I think we
just need a "Additional smoothing amount (within optikey):" setting. If
separate we need a way of conveying if combining them is meaningful, rather
than choosing between two methods.

AdamRoden:
From what I can see, KF is 100% within Optikey. If that is the case we can simply remove the checkbox.

Source
Data Stream Processing Amount (before Optikey):
Point Smoothing Amount (within Optikey):

We need to choose our smoothing preference (KF on/off) so users won't have to.

JuliusSweetland:
Is the point smoothing amount cooling the KF? If it is then that scale
could be the smoothing effect: none, low, medium, high scale?

AdamRoden:
> Is the point smoothing amount cooling the KF? If it is then that scale could be the smoothing effect: none, low, medium, high scale?

Yes. For KF I created an expanded scale: 0-None to 5-Highest.

AdamRoden:
> Is the point smoothing amount cooling the KF? If it is then that scale could be the smoothing effect: none, low, medium, high scale?

Yes. For KF I created an expanded scale: 0-None to 5-Highest. 

Is this what you mean:
Source
Data Stream Processing Amount (before Optikey):
Smoothing Effect (within Optikey):

JuliusSweetland:
Yes I think that is easy to understand. What do you think?

If you agree we just need to rtest with Kirsty and her changes and agree
the defaults.

Looking good!

AdamRoden:
Ok. I'm cleaning up a few things now and will have some new code in this PR soon.

AdamRoden:
New code provided. I'm pretty happy with how things are named and the variance of the smoothing effect from low to highest in this version. 

image

JuliusSweetland:
That looks great. I'm give things a test shortly. After that I may hold off
until Kirsty has taken a look as there could be more changes to the KF

JuliusSweetland:
@AdamRoden I really like the Management Console settings and the smoothing feels good. The only thing I've noticed is if you turn up the smoothing to max, turn on magnetic cursor, and then attempt to fixate on points on the screen I find the cursor makes a nicely smoothed movement towards the point I am looking at, but stops just short of where I am looking. You can coerce the cursor to where you want it by looking *past* the point you want, but it feels like there's something slightly wrong in there.

Can you replicate this?

AdamRoden:
> @AdamRoden I really like the Management Console settings and the smoothing feels good. The only thing I've noticed is if you turn up the smoothing to max, turn on magnetic cursor, and then attempt to fixate on points on the screen I find the cursor makes a nicely smoothed movement towards the point I am looking at, but stops just short of where I am looking. You can coerce the cursor to where you want it by looking _past_ the point you want, but it feels like there's something slightly wrong in there.
> 
> Can you replicate this?

@JuliusSweetland Yes. That is kind of the point of the "Highest" setting, for it to require you to look past wherever you want the cursor to go. However, I agree with you that if you're moving to a new target, (i.e. across the screen,) it should not slow down like that until you are all the way there, but it feels like it always starts hitting the brakes before fully reaching your target.

I will tweak the algorithm some more with that in mind.

JuliusSweetland:
Great, thanks Adam. I think this new smoothing algorithm is very promising.

AdamRoden:
I updated the code and altred some terminology. 
Called highest settings Ultra and Extreme. 
Made large movements capable of more speed on the higher levels so they can jump across the screen but still be able to have very fine movements once you lock in your focus.

AdamRoden:
@JuliusSweetland @kmcnaught I put a lot of effort into refining SmoothingEffect. Finding the right balance was quite challenging, but I think I'm happy with this version and am ready to put it to bed. Hopefully, you both agree and no more changes are needed.

JuliusSweetland:
I'll test again this evening. I've been working on adding Hindi for the last few nights.

kmcnaught:
Also taking a look tonight

AdamRoden:
I guess I'm not ready to put it to bed. 
I've been tuning it further today and have a couple more tricks to improve how it moves. Also, I've decided it make more sense to unhook from the point selection's lock-on and fixation radius values, and instead created a new Smoothing Effect Deadzone setting.
If you've tried it out I'd appreciate your feedback. If not, you might want to wait one more day.

kmcnaught:
Definitely agree on removing the interaction between point fixation settings and smoothing - it was bothering me a bit that these were linked. 

In general I think it's important to think about gaze filtering and point selection functionality separately (although it's possible that in the end there will be some interplay, for instance perhaps increasing smoothing for part of the point selection process). The historical point selection method can be a bit unforgiving in a way that is particularly problematic for a jumpy eye tracker. The extra smoothing definitely helps (a lot) but if you find you're adding lots of extra logic to your smoothing to improve point selection specifically, it might be better to rethink the way the lock-on happens with point selection. For instance, Tobii's Gaze Point app doesn't lock the cursor position at all while you're dwelling, it just resets the dwell if you move "too far". This lets you make minor adjustments while the dwell process is happening. I'm not saying this is necessary the right way to do things, but something closer to this might help improve things.

JuliusSweetland:
I'll hold off for another day then. I've got a final bit of Hindi to do.

AdamRoden:
@kmcnaught What are your impressions of this smoothing effect? I know it gives me an awesome magnetic cursor and an improved mouse clicking experience, but what about you on your system?

kmcnaught:
(Note that I've not had a chance to play with changing any settings, so I'm just working with default settings so far)

It's definitely a significant improvement on what we had before, and now works pretty well out the box with my EyeX (where previously I would often fail to make mouse dwells without changing some settings, either reducing dwell time and/or changing lockon etc radii)

I still have some issues with how the mouse cursor gets locked for a mouse action - the experience is great if your initial saccade hits the spot, but once you have any calibration bias it can be difficult to make corrections. This is a lot better since the smoothing improved, but still something I think could do with some thought (separate to this PR). 

With magnetic cursor it's generally massively improved, but I'm having a few issues with when it locks completely - I'd like to always be able to nudge it, for instance when I'm moving along some adjacent icons looking at the hover labels. As it is, you end up with a jumpy cursor for small movements. On balance I think the super-smooth-not-moving state looks impressive when you get it right, but is subject to cause problems when it's not quite what a user wants. It looked amazing when I first tried it, and then caused UX grumbles at various points when testing.

I recorded a couple of videos to show what I'm seeing. If you'd like to try my KF implementation, it hasn't been tuned much, still has some hardcoded stuff and I haven't touched any defaults for mouse actions, but it's functional here: [kf/exponential](https://github.com/kmcnaught/OptiKey/releases/tag/kf%2Fexponential). You can turn it on and off with Ctrl+K and Ctrl+Shift+K (so long as OptiKey has focus) [this is just for testing!]. I'll send over some notes by email.

Your version: https://www.youtube.com/watch?v=rW36TdNBQIA
My KF version, for reference: https://www.youtube.com/watch?v=aNzGeXEq7yA

AdamRoden:
@kmcnaught  Thank you for your code. It was exactly what I needed to see. By taking some of your code and combining it back into mine, I think the final product is just right. I'm very satisfied now.

Highlights:
Utilizes a KF algorithm for Point Source.
Skips KF when looking at a key.
Only 1 setting to worry about: Smoothing Effect
Options: 0-Off to 5-Extreme to influence the acceleration of your mouse movement and the deadzone size.

kmcnaught:
Incidentally, I tried turning off the locking behaviour for point selection and find it a lot more effective with my setup, particularly combined with the exponential KF:
https://github.com/kmcnaught/OptiKey/commit/5f6b7f9e6921ea1d0ad6fdb2916fa18b558f4b57

You still get the correct behaviour wrt the dwell resetting if you move somewhere else, but you're able to make small adjustments easily during the dwell process.

JuliusSweetland:
What smoothing are you using for the first video in the data section where
you move between 2,3,4 keys in Minecraft? The accuracy and movement looks
great, but the link appears to be in the section using the old KF
implementation?

kmcnaught:
@JuliusSweetland I'm using the kfExponential version in the video, aka my Proposal 2. It was supposed to an illustration showing the experiment, not associated with a specific version.

JuliusSweetland:
It's very impressive. I'll leave the details of what makes the final cut to
you two.

Great write up as well. Really detailed. I might see if you've got time
next week to explain some of it to me!

AdamRoden:
@kmcnaught I reset my code again so the commits can tell the clear story. 
First, I merged your smoothed pointer fixation into my branch. 
Second, I pushed a commit using what I call Benford smoothing. I incorporated a 'Gain' coefficient based on the smoothing level from settings. It works nicely and I'm curious about how it performs in your test. 
Third, I overwrote the Benford smoothing example in favor of the Kalman smoothing we've been talking about. 
This KF is based on your exponential model, but enhanced and tuned to be 2D and to be influenced by Smoothing Level. As you will see, it includes my bypasses when you look at a key, but per your suggestion, I've removed all pointer locking from the Smoothing Effect.

That's the summary. You can review both. For best performance in your test, make sure to have the Smoothing Level set to (1 - Low). All higher settings are likely detrimental to dwell clicking and are primarily aimed at the use of the Magnetic Cursor combined with an accessible switch to click. 

For reference:
Benford: 0ea1fad446bec0c366e1de97917fde50b92b2ac8
Kalman: ba06cf5184b02e7d0d9dd39de2e0f4fd036ee422

JuliusSweetland:
@AdamRoden @kmcnaught  I'm going to test this properly over the weekend. I've reviewed everything and understand your changes. The levels of smoothing confuse/worry me slightly as you say that anything over low is not suitable for anything other than magnetic mouse. Would it not be better to have 2/3 useable levels of smoothing for the general use case, and then one "ultra" level which is clearly designed for use in a certain scenario? Thoughts?

JuliusSweetland:
@AdamRoden Hi Adam, I've dug into your PR in more detail and have some feedback ideas for you:

1.To keep the git history cleaner and more understandable could you please remove Kirsty's commit (as a separate PR exists which I will merge) and squash your other changes into one? The full history can exist on your fork, but my worry is that large changes are added and removed over a few commits and it is difficult to follow the flow as an external observer.

2.You say "make sure to have the Smoothing Level set to (1 - Low). All higher settings are likely detrimental to dwell clicking and are primarily aimed at the use of the Magnetic Cursor combined with an accessible switch to click.". This suggests to me that the higher smoothing levels would be broadly unuseable for general use cases. I would like to avoid this. Would it be possible to tweak values to that all levels are useable for all use cases, i.e. the higher the level, the more it feels like the gaze point is dampened. It is then up to the user to get the "feel" for a smoothing level and to decide which is best for their use case.

3.I'm not convinced by the need for 3 branches in the smoothing code. I understand the first two cases: a user stays on the same key, or they don't. I'm not sure we can be so black and white with the lower quality eye trackers. It would be fairly common for the user to be fixating towards a key edge, have a point occur on another key, and then lose their smoothing history because of a rogue point. I always tend towards keeping it simple; if the distribution in the main catch all case is correct, then it should work everywhere; whether the user is directing their attention to keys or to select a point. I'm happy to be proved wrong, but I worry about edge cases (no pun intended) where the user is directing their attention to the middle of a key, but due to a bad calibration or whatever the registered points from the tracker are hovering around the intersection of 2/4 keys. This ruins the smoothing effect. If we dropped the first two cases it makes the model easier to understand, the effect easier to predict for the user, and gives us the opportunity to assess the usefulness in isolation. 

Lastly, there are some magic numbers in your smoothing algo (5 and 3000) - @kmcnaught are you in a position to opine on their values and whether there is enough meaning to extract them into named variables so it is easier to understand?

Overall - it's all great stuff! I'm really excited about getting this tested and finalised. Happy to discuss anything - on a call if easier?

AdamRoden:
Great input! God knows I wish I could talk on a call, but you wouldn't
understand a word out of my mouth.
1. Yes. I usually try to squash my commits to show compartment changes. I
hate seeing too much piecework in the central source.
2. I believe I addressed in my email about laying the groundwork for a
future option, though I'm also fine returning to a filter on/off setting at
this stage.
3. Agree. Only needed for high-level smoothing, which feels more and more
like a separate feature of magnetic cursor.

Magic numbers are... magic. Exposing them in MC could have
a disastrous impact on the user. They're all about tuning the algo, which I
think still has room for improvement.

Overall, I love the dialogue, and although I am eagerly sharing ideas, I
don't want to rush to a solution that's not proper.

Adam Roden

JuliusSweetland:
1.Great

2. I think on/off might be taking it back too far. I'd ideally like to see some workable levels. I imagine (overly simplistically maybe!) that it ranges from off, through lightly smoothing the jitteriness, to feeling like my cursor is being dragged through treacle (a bit, nothing too crazy!). This gives users a really easy to understand experience. We would then decide on an appropriate default (perhaps per eye tracker as some are noisier than others).

3.Great

No I don't want the magic numbers exposed - I meant refactoring them up into variables/constants with meaningful names. The users won't see them - it's for us techies. 3000 means nothing to me, but an aptly named variable might...

kmcnaught:
On the magic numbers:

**3000**
This is the measurement noise, which is a very standard KF term. 

**5**
This is a bit trickier to name, as it's specific to my exponential process model. It's a scaling factor. Perhaps `ProcessNoiseScale` with a comment:
```
// This defines how the assumed process noise grows with distance from the last point
// A larger number will increase the influence of the current estimate, resulting in smoothing across a larger radius
ProcessNoiseScale = 5*SmoothingLevel
```

I'd also be keen to add this comment to the exponential equation:
```
// We want process smoothness to dominate within the range of microsaccades, but to converge
// towards a uniform prior for larger jumps. An exponential mapping lets this happen smoothly.
```

I also noticed that `EstimationNoise` isn't initialised - this should be initialised to a big number. 

I'm quite happy with how things are looking, but I won't get a chance to try it out until the weekend. I agree with Julius that we'd ideally have a few different smoothing levels suitable for majority-use (allowing for personal preference for speed vs control). Ideally Smoothing = 1 would be fairly minimal (gets rid of worst jitter but unlikely to cause _any_ complaints about lack of responsiveness), and the next levels would allow increasing control/sluggishness for those who prefer it. I think we all need to try out the different smoothing levels and see how they feel before we finalise this. 

In terms of extra behaviours for key selection and/or magnetic cursor, I think we should definitely keep ideas flowing for future enhancements, but I'm much happier to keep things simple for now and see how everyone gets on with it. 

Finally, are you able to remove all the Kalman filter settings from Settings.cs (and anywhere else they might be referenced) since they are now unused?

AdamRoden:
Okay, today I'll reset my code to wipe previous commits. Then, I'll strip the key checks, reduce the levels to none-low-medium-high, give names to numbers, and see if I can further tune the formula for an update to the PR. I'm thinking something more can be done to reduce the floatiness of the cursor when you are trying to hover, especially the longer you stay on a point. However, it has to be done in a way that does harm the quick movement to a new point. So far, the silver bullet has eluded me.

Any other adjustments today?

JuliusSweetland:
@AdamRoden Not from me. That all sounds great.

kmcnaught:
@AdamRoden just a thought, would it make sense for the smoothing scaling level to be a (simple) function of both the chosen smoothing level and the dwell amount, something like:
`ProcessNoiseScale = 5*(SmoothingLevel + pointSelectionDwellAmount)`

this would let you bump up the smoothing between your chosen level and one level higher. You'd still benefit from the exponential KF's inbuilt mechanic for allowing quick saccades, but over the course of the dwell your "smoothing window" would effectively be gradually increased.

AdamRoden:
> @AdamRoden just a thought, would it make sense for the smoothing scaling level to be a (simple) function of both the chosen smoothing level and the dwell amount, something like:
> `ProcessNoiseScale = 5*(SmoothingLevel + pointSelectionDwellAmount)`
> 
> this would let you bump up the smoothing between your chosen level and one level higher. You'd still benefit from the exponential KF's inbuilt mechanic for allowing quick saccades, but over the course of the dwell your "smoothing window" would effectively be gradually increased.

This assumes a point selection is in progress? We would need to apply this concept to the magnetic cursor as well, only in that case do not trigger a click. I like it

kmcnaught:
Yeah sorry I was implying the variable pointSelectionDwellAmount would be zero if there's no point selection dwell in progress - off the top of my head I don't know where you'd be getting that information from.

What kind of scenarios do you use magnetic cursor for, and what behaviour are you trying to achieve? I can imagine we might at some point want to allow a different smoothing level for magnetic cursor vs [everything else] but since we're not dealing with dwells I'm not sure why you'd want special treatment above what the KF is doing? 

I'm keen that we all try out the current version with the different levels of smoothing before adding any other complexity. I imagine we might try them out and think "3 is great for magnetic mouse but 2 is great for everything else" or "3 is great for mouse clicks but too much for key selection". If this is how we're feeling we can think about some special scaling like this, but I'd much prefer we'll all familiar with how the simplest option works first. 

Could you please let us know when you've got a version with your current preferences for each of the smoothing levels?

AdamRoden:
@JuliusSweetland @kmcnaught I think we're closer than ever now. Commit a74ac5ac50f42a9382ca0a559250ae84f168f4d2 is a big step forward. Compared to previous models, it is superior in almost every regard. The combination of speed and smoothing is something you'll have to experience yourself. Give it a go and let me know how it works for you.

src/JuliusSweetland.OptiKey.Core/Observables/TriggerSources/PointFixationSource.cs
```diff
@@ -133,6 +133,7 @@ public IObservable<TriggerSignal> Sequence
                                         //Publish a high signal if progress is 1 (100%), otherwise just publish progress (filter out 0 as this is a progress reset signal)
                                         if (progress > 0)
                                         {
+                                            fixationCentrePointAndKeyValue = point.Value;
```
JuliusSweetland, line 136: Things may still be in progress, but can you remove Kirsty's commit from this PR? Thanks.

JuliusSweetland:
@AdamRoden One small comment, but otherwise it looks clean and exactly what we talked about. Great stuff. Is this ready for testing?

src/JuliusSweetland.OptiKey.Core/Observables/TriggerSources/PointFixationSource.cs
AdamRoden: Done and my comment edited to new commit.

AdamRoden:
> testing?
Ready to test. I am running it on Low and it keeps up nicely. Medium is good but too slow to keep up with the way I type. High is workable for slow typing, but again too slow for me.

AdamRoden:
> @AdamRoden One small comment, but otherwise it looks clean and exactly what we talked about. Great stuff. Is this ready for testing?

Yes, it is ready to test. I am running it on Low and it keeps up nicely. Medium is good but too slow to keep up with the way I type. High is workable for slow typing, but again too slow for me.

JuliusSweetland:
Excellent, I'll aim to test over the weekend if I can. For the higher
smoothing levels imagine a 75 year old with advanced MND as an example
user. They want everything slowed down, dampened and deliberate. It gives
them time to figure out what they want to select and get there with the
brakes half on. I was on a call today with a family in this exact scenario
- a lack of computer experience, coming from Macs to Windows (so everything
is new and confusing), plus the complexity of Optikey and the concepts that
it introduces. Then, as their confidence grows they ramp up through the
gears by dropping the smoothing level and decreasing the dwell times.

I can see these settings eventually being combined into a simplified
approximation of the management console with setting groups like Beginner,
Intermediate, Advanced. That would be cool.

AdamRoden:
@JuliusSweetland I think you know the general audience better than me, so keep them in mind when you test the high setting. It has to jump at a certain point in order to keep up with your eyes if you are keying, but the jumps bother me when I have the magnetic cursor turned on. We can sacrifice speed for smoother movement by changing part of the formula for measurementNoise from the smoothingBase^3 to something less. It is a drastic change5 but might help a new user. (I don't know. Just pointing it out.)

JuliusSweetland:
Let's reconvene after testing. Thanks

src/JuliusSweetland.OptiKey.Core/GazeSmoothingFilter.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public class GazeSmoothingFilter
+
+    {
+        private double EstimationNoise;
+        readonly double ProcessNoise;
+        private double Gain;
+        private double SmoothingLevel;
+        private Point EstimatedValue;
+
```
AdamRoden, line 17: Commenting needed:
>>EstimatedPoint: The point that will be returned for use after applying the filter to the delta of the measured point (n) and the previous estimate (n - 1)
Gain: Calculated each iteration and used to scale (0 to 100%) the delta between point (n - 1) and (n) 
SmoothingLevel:  The setting that a user can adjust to influence the filter behavior where a larger value equates to a less responsive more "floaty" cursor
EstimationNoise: Requires a large initial value for (n = 0) but will then fluctuate every iteration 
ProcessNoise: Constant used to fine-tune the filter where a larger value equates to slower movement. Anything from 400 to 3600 seems to work well in this model

src/JuliusSweetland.OptiKey.Core/GazeSmoothingFilter.cs
```diff
@@ -0,0 +1,39 @@
...
+            EstimatedValue = new Point(0, 0);
+            EstimationNoise = 1000;
+            ProcessNoise = 500;
+        }
+
+        public Point Update(Point measuredValue)
+        {
+            SmoothingLevel = Settings.Default.GazeSmoothingLevel;
+            var smoothingBase = (measuredValue - EstimatedValue).Length / SmoothingLevel;
+            var measurementNoise = ProcessNoise / (1 + Math.Pow(smoothingBase, 3));
```
AdamRoden, line 29: // We want process smoothness to dominate within the range of microsaccades, but to converge
// towards a uniform prior for larger jumps. An exponential mapping lets this happen smoothly.
// A power of 2.5 would be slow and smooth, but a power of 3 is fast and free of jitters

src/JuliusSweetland.OptiKey.Core/GazeSmoothingFilter.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            EstimatedValue = new Point(0, 0);
+            EstimationNoise = 1000;
+            ProcessNoise = 500;
+        }
+
+        public Point Update(Point measuredValue)
+        {
+            SmoothingLevel = Settings.Default.GazeSmoothingLevel;
+            var smoothingBase = (measuredValue - EstimatedValue).Length / SmoothingLevel;
```
AdamRoden, line 28: smoothingBase = distanceMoved = delta
This is the base number we will use in our exponential

JuliusSweetland:
@AdamRoden @kmcnaught Testing thoughts:

1. If the Kalman filter is no longer used can we remove all references? If you search you'll find properties like "KalmanFilterSupported" all over the place - are these needed? Also can we delete the old KalmanFilter class?
2. The new GazeSmoothing filter should be in the DataFilters folder and namespace?
3. In terms of the smoothing effect - I think it is really, really good. The difference between the filter level being off and on is night and day. I do, however, wonder if the differences between the smoothing levels are a bit too large. I felt that low was really nice to work with, but I almost wanted a level somewhere between none and low, and between low and medium. The step up between them was just a bit too large for me. I like how deliberate High feels. Medium felt quite similar to me. Maybe if we move to 5 levels of on (plus level 0 -None) then we could have something like:
0 - None
1 - Very low - this is new
2 - Low - as it is now
3 - Medium - this is new
4 - High - this is the current Medium
5 - Ultra - this is the current High?

I didn't feel like I wanted any more smoothing than the top level as that results in key fixations being on the wrong key sometimes and then eventually catching up and finding where I was looking. If this were slower I think selections on the wrong key would be more common.

FYI you're probably already doing this, but if you open the Tobii Experience app in the background so you can see where the tracker thinks you are looking it gives good insight into how the smoothing is behaving behind the scenes.

Overall, I'm really happy with the improvement, especially the way you can now adjust your mouse selections in the magnifier during the fixation animation. This is comparable to Tobii's zoom click behviour I feel.

src/JuliusSweetland.OptiKey.Core/GazeSmoothingFilter.cs
JuliusSweetland: @AdamRoden If you coud decorate the code with these comments that would be very helpful.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/Management/PointingAndSelectingViewModel.cs
```diff
@@ -129,6 +129,20 @@ public PointingAndSelectingViewModel()
...
+                return new List<Tuple<string, int>>
+                {
+                    new Tuple<string, int>("0 - " + Resources.NONE, 0),
+                    new Tuple<string, int>("1 - " + Resources.LOW, 1),
+                    new Tuple<string, int>("2 - " + Resources.MEDIUM, 25),
+                    new Tuple<string, int>("3 - " + Resources.HIGH, 50),
+                };
+            }
+        }
+
```
AdamRoden, line 145: @JuliusSweetland For the sake of organization and consistency in design, is this how you would supply the GazeSmoothingLevels, or would you create a class, or something else?

JuliusSweetland: Yes this is fine. I would use a Tuple of something noddy like this, rather than create another class which is really just a KeyValuePair.

JuliusSweetland: Although having said that, just above it is an example of using a KVP: new KeyValuePair<string, DataStreamProcessingLevels>

You could switch to that for a more explicit binding in the XAML - that way you really have Key and Value properties to bind against, instead of Item1, Item2.

JuliusSweetland:
@AdamRoden FYI there's a slight conflict on this branch now that the Hindi branch has been merged in. It should be an easy thing to resolve - you just need to merge the new Hindi strings with your Resource changes/additions. Nothing should be lost.

JuliusSweetland:
@AdamRoden @kmcnaught I've done a bit more testing. Really pleased still. One more consideration: once this is finalised, agreed, and merged we need to discuss the new defaults. I think this release should be like Android's "project butter" where they tried to tighten everything up for users. What do you think of turning on Look to Scroll as the default scrolling mechanism? I find the scroll a bit at a time default to be slow and annoying, but Look to Scroll is complex and potentially confusing. Thoughts?

AdamRoden:
@JuliusSweetland @kmcnaught I've just realized that my Data Stream Processing Level has a significant impact on gaze smoothing performance. I may have been experiencing better results because of it. When set to None, smoothing is not nearly as nice. So... what about you two?

Regarding Look to Scroll, it should absolutely be the default with Free Scrolling and Screen at Point. Deprecate Window at Point, and Window at Center. Deprecate display of Window Bounds. Update the Deadzone bounds to use an optional icon. I have a few examples I made for my AutoHotKey scrolling.
MiddleWheel
Wheel
WheelDown
WheelUp
image

JuliusSweetland:
Let's move the scroll to look to a new conversation as it's separate. I'm excited to discuss it after we finalize this. Let me test quickly with various stream processing levels.

JuliusSweetland:
@AdamRoden Two things:
1. I don't suffer any major problems with any of the stream processing amounts. I am using a 14" screen with a Tobii Eye Tracker 5 which isn't even fixed in position correctly. With no stream processing the cursor was jittery, but adding smoothing helped a fair amount. I felt the accuracy was a bit worse than usual. With high stream processing levels (and various smoothing levels everything was fine again). I'm pretty happy with it.
2. Can we default the smoothing level property. I'm sure the first time I ran Optikey the drop down was empty (as I wouldn't have had any stored value for that setting). Can we default it so it is populated?
3. What did you think about my earlier suggestions: https://github.com/OptiKey/OptiKey/pull/726#issuecomment-699496402
4. I ahve created a new issue to discuss scroll to look: https://github.com/OptiKey/OptiKey/issues/731

AdamRoden:
@JuliusSweetland 2 things, aka 3.
1. I know we can do better. My new goal is to eliminate the need for stream processing, and I'm close.
2. Yeah, we'll do something before merging for release.
3. I think two more settings is overkill. I think low smoothing should eliminate jitter, medium should add a bit of noticeable cursor smoothing, and high should be for introducing the software, but annoying for full-time use.

JuliusSweetland:
@AdamRoden Ok - let's see what you come up with...

By the way I'm so impressed with this I think I'll bump up to v4 as this is will be a game changer for a lot of people.

JuliusSweetland:
Forgot to add: whilst useful levels of smoothing is obviously what we want, spending time trying to replace the stream processing is not necessarily required. For some trackers it will be useful, but the Tobii trackers expose good streams of processed data (which is actually processed on chip, so it reduces the CPU load and gives us the option of doing no smoothing on the device if it is too underpowered to handle it) - we should be using it if we can as I would assume that it will only get better over time

AdamRoden:
Agree. I'm mainly trying to tune what I've got to get rid of the jitter because before the device was doing most of the work preventing me from recognizing whether I was tuning properly.

src/JuliusSweetland.OptiKey.Core/GazeSmoothingFilter.cs
kmcnaught: A bit more precise on a couple of these:

ProcessNoise: the amount of uncertainty in our model of eye movements. This is low when we believe the user is fixating and large when we believe they've moved to a new location.

EstimationNoise: the amount of uncertainty in our current running estimate. This feeds into calculating how much we should weight new data coming in.

kmcnaught:
Hey, afraid I haven't quite got to the point of trying it out, but I will do later today. I've taken a look at the code and I have one (long) comment...

In the latest changes to the algorithm, the meaning of the terms has changed significantly. I'm quite happy if the final algorithm ends up not exactly following the KF framework, but in that case the terminology that I was suggesting becomes misleading. 

If we're thinking of it from a KF-like perspective, we've got two sources of uncertainty. One is uncertainty about how much the eyes _could_ have moved since the last estimate - in the original KF framework the assumption was "probably not very far", in my exponential process model the assumption was "the further they've moved the less certainty we are of where they might have moved to" which is a smoothed version of "not very far if fixating, OR all bets are off if saccading". 

Totally separately from this we have the concept of measurement noise, i.e. *given* a particular measurement out of the eye tracker, how much uncertainty is there about the true underlying position?

The balance of these two uncertainties tells us how much to update our estimate - if the process noise is much greater than the measurement noise, we trust the new measurement very highly, but if the process noise is higher we trust our running estimate more than the new measurement.

With my exponential KF I was saying: the measurement uncertainty is constant regardless of where you're looking (this is eyetracker noise plus micro-saccades, neither or which are going to vary much for long vs short movements). The process noise however increases the further your eyes jump - basically, once we're totally sure we have moved to a new location, our running estimate + process model are completely useless and we should trust the new measurement.

With your new calculations and variable names you're effectively saying the *measurement* noise goes down the further you've moved, which is hard to justify. Nothing about the eye tracker has changed. The end result of the calculations is very similar to the KF-like version, because it changes the _relative_ size of measurement noise and process noise, so it works out well, but it doesn't describe a meaningful model, certainly not the one implied by the variable names. 

I don't want to force your algorithm into a particular formulation - like I said I don't mind if it isn't KF-like any more. But you do need to think about what assumptions or goals each of your calculations is encapsulating, and make sure the variables are named and/or commented appropriately so that next time someone else tries to understand this code they can understand what you've implemented. Or keep the names and rearrange your mathematical expressions so that you're increasing process noise instead of decreasing measurement noise ;)

AdamRoden:
@kmcnaught I agree with my terms being off. I'd like to clean up the algo using your expertise.

kmcnaught:
I've spent a few hours trying this out and trying to figure out what's not quite working for me. The main problem for me is that my setup is a lot worse than yours, so the smoothing levels really aren't calibrated for my setup. This makes it difficult to compare notes with you but is actually a blessing in disguise, since it means we're exploring the full problem space better, right? :)

I know part of this is that I'm using an old EyeX, and also my environmental setup isn't ideal. I wonder if part of it is also down to screen resolution? I know we're still hardcoding tuning numbers regardless of screen resolution, so the smoothing behaviour is based on pixel deltas. Ideally it would be based on "degrees of visual angle" but as a proxy I'd consider "% of screen", since the angle of a screen is likely to be more similar from one person to the next than their screen resolution. This might help.

Aside from this I've been trying to explore the mappings you've come up with. I've plotted what your gain curve looks like for a particular scenario (SmoothingLevel 3 and current EstimationNoise = 100)

What I'd be looking for (scaled to my setup at least) is a gain plot that shows this behaviour (forgive my paint skillz):
image

What I'm seeing from your new algorithm is this:
image

The exact graphs vary depending on current estimation noise, for example with estimation noise = 500:
image

The behaviour your algorithm is exhibiting the right overall shape, but consistently (a) scales up rapidly at smaller pixel distances meaning less smoothing at a fixation, but also (b) still stays < 1.0 for a long time meaning some residual smoothing remains after a fixation. 

I also notice the behaviour is a bit inappropriate if the current estimate uncertainty is very high. For instance, hypothetically assume your current estimate noise = 1,000,000 (i.e. your current guess has error bars with standard deviation = 1000 pixels). In this case your running estimate is almost meaningless, and you'd like to completely ignore it. However, your version still has some smoothing until 5-10 pixels away. The end result is fine - because the estimate is close, it doesn't really matter that we're using the estimate, but it illustrates some concerns with the underlying model.
image

Since you're obviously not privy to what my eyetracker inputs are looking like (i.e. a lot worse than yours) I've made a short video showing the behaviour I'm seeing. Overall I'm finding your algorithm (at SmoothingLevel 3) is both a bit less smooth than the KF but also a bit less snappy on intermediate-size saccades. This is consistent with the slower slope on the gain plots above.

[https://youtu.be/YRl4tHLsV2A](https://youtu.be/YRl4tHLsV2A)(still processing at time of posting..)

Would you mind leaving this with me for a few days and I'll see if I can get my model framework into your smoothing level framework? Then I can propose a range of numbers which work well for me, and get you and Julius to comment on how well they work for you? Maybe we can then meet in the middle somehow.

kmcnaught:
Looking at this again in the context of your comments on floating at the fixation - your gain curves generally start lower (more smoothing) at really small distances, but then overtake mine quite quickly (less smoothing). This fits exactly with what we've been observing - at ~0-5 pixels yours has much higher smoothing (no floatiness) but beyond ~5-20 pixels (depending on SmoothingLevel) the smoothing suddenly drops. For me, with a noisy eye tracker, this meant I had a lot of small jumps between stationary positions. 

I realise this version is now obselete so I'm not going to over-analyse these curves, I'm just trying to work out what I should be aiming for - it sounds like we want something in between the two in terms of gain curves at the lower end (near 0-10 pixels). All the extra exponential filtering steps in the latest revision make it a lot harder to do this particular comparison.

PR #728, opened 2020-09-24
Disable locking behavior during point fixation in favor of continuous…

AdamRoden:
… smooth motion

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/UI/Controls/CursorPopup.cs
+++ b/src/JuliusSweetland.OptiKey.Core/UI/Controls/CursorPopup.cs
@@ -77,12 +77,16 @@ namespace JuliusSweetland.OptiKey.UI.Controls
             };
             mainViewModel.OnPropertyChanges(vm => vm.ShowCursor).Subscribe(calculateIsOpen);
             calculateIsOpen(mainViewModel.ShowCursor);
-            
+
             //Calculate position based on CurrentPositionPoint
             mainViewModel.OnPropertyChanges(vm => vm.CurrentPositionPoint)
-                .Where(cpp => IsOpen && SelectionProgress == 0) //Only set Point if popup is open and there isn't a current fixation in progress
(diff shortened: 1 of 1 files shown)


JuliusSweetland:
I like this. @kmcnaught Are you happy with this change? There are a lot of concurrent changes and discussions going on so I'd like input from everyone before I blindly merge things in.

AdamRoden:
This should have been the first place I went when Kirsty brought up the issue of the cursor not being smooth. 
Also, some food for thought, perhaps it is more appropriate to have our input stream update be purely Kalman based, and to have addition slow-motion and cursor locking effects applied either here, or if the Magnetic Cursor is active, in its event handler. 
We'll see what Kirsty says.

JuliusSweetland:
Interesting ideas. If we go down that route my initial thought would be to
toggle the KF on/off (or increase/decrease its effect) from the code which
opens/closes the magnified popup or starts/stops the magnetic cursor, so
the input service handles all requests like that.

AdamRoden:
So, I guess this little change only works nicely when paired with the Kalman Filter. After more testing without raw unfiltered points, the results were ugly.

JuliusSweetland:
Maybe we could work out the minimum amount of smoothing required to be able
to turn off this locking and link the two behaviours?

kmcnaught:
I'm 100% behind the idea of removing the locking behaviour, but I don't think this commit does it correctly - have you tried my attempt? https://github.com/kmcnaught/OptiKey/commit/5f6b7f9e6921ea1d0ad6fdb2916fa18b558f4b57 - you may wish to cherry-pick the commit onto your preferred filtering setup. 

When I try this PR I get quite jittery behaviour - I think you're correcting the position of the cursor visualisation rather than the point fixation centre. If you remove lockon on the point fixation centre instead (as per 5f6b7f) then the cursor naturally follows

AdamRoden:
@kmcnaught Your change is better. PR that.

PR #729, opened 2020-09-24
Allow fixation point to be updated during dwell

kmcnaught:
This removes the lockon behaviour during a point selection, letting you continue updating the point (within the appropriate radius).


kmcnaught:
This should be independent of any changes to filtering

AdamRoden:
@JuliusSweetland I approve of this change.

PR #732, opened 2020-09-27
GazeSmoothingFilter

AdamRoden:
Updated on October 5.
The changes feature 6 choices for "Adaptive smoothing level" and an additional option to enable "Smooth movement to new gaze targets" in MC. 
The highest levels are intentionally slow to contrast the low settings.

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/DataFilters/KalmanFilter.cs
+++ b/src/JuliusSweetland.OptiKey.Core/DataFilters/KalmanFilter.cs
@@ -1,29 +1,65 @@
 ﻿// Copyright (c) 2020 OPTIKEY LTD (UK company number 11854839) - All Rights Reserved
+using JuliusSweetland.OptiKey.Properties;
+using System;
+using System.Windows;
+
 namespace JuliusSweetland.OptiKey.DataFilters
 {
     public class KalmanFilter
     {
-        double ProcessNoise; //Standard deviation - Q
-        double MeasurementNoise; // R
-        double EstimationConfidence; //P
-        double EstimatedValue; // X 
-        double Gain; // K
+        private Point EstimatedPoint;     //The point used after applying the filter.
+        private double EstimationNoise;   //This value fluctuates to balance each new measurment with the previous estimate.
+        private double Gain;              //Scale from 0% to 100% applied to the movement delta when updating the Estimate
+        private Point Measurement1;
+        private Point Measurement2;
+        private readonly double MeasurementNoise;
 
-        public KalmanFilter(double initialValue = 0f, double confidenceOfInitialValue = 1f, double processNoise = 0.0001f, double measurementNoise = 0.01f)
+        public KalmanFilter()
         {
-            this.ProcessNoise = processNoise;
-            this.MeasurementNoise = measurementNoise;
-            this.EstimationConfidence = confidenceOfInitialValue;
-            this.EstimatedValue = initialValue;
+            EstimatedPoint = new Point(0, 0);
+            EstimationNoise = 1000;
+            MeasurementNoise = 2000;
         }
 
-        public double Update(double measurement)
+        public Point Update(Point measurement)
         {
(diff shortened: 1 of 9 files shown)


JuliusSweetland:
@kmcnaught I'm happy with this behaviour, with the exception of the naming of the "smoothing boost" setting. What do you think?

kmcnaught:
Looks (and feels) good to me.

In my head the "smoothing boost" is "a little bit of extra global smoothing on top of the adaptive smoothing". It improves fixation-stickiness but decreases sacadde-responsivement slightly. I can't immediately think of a better user-facing term than "smoothing boost" or "extra smoothing" or similarly vague terms. Ideally though we'd disambiguate between the two things better - otherwise it's very hard for a user to understand the difference. For example, I personally prefer Level 2 vs Level 1 + extra smoothing, even though they feel similar in terms of "amount of smoothing" - the latter isn't as rapid when saccading.  

Some more brainstorming
- Rename "Smoothing Effect" -> "Adaptive Smoothing Level" to disambiguate slightly
- "Extra Stickiness"
- "Extra Friction" (sounds technical but I think it's used in mouse cursor contexts?)

AdamRoden:
I like "Smoothing Level" or "Adaptive Smoothing Level". 
I think I prefer Level 2 with no extra.
I like the terms Extra Friction or Suppress Instant Movement or Prevent Pointer Jumping

JuliusSweetland:
Posted here as well as on the chat: Additional smoothing of small movements?

JuliusSweetland:
React to gaze movements instantly?
Enable instant movements?

AdamRoden:
"Enable Instant Movement"
Sounds good to me. 
Are we keeping 6 smoothing levels? Should any be adjusted up or down?

JuliusSweetland:
I like the 6 levels and thought they all felt like they were fairly evenly
spaced.

kmcnaught:
I'm happy with the levels, and "Enable instant movements". Would also accept "React to gaze movements instantly?" but it's a bit more wordy,

JuliusSweetland:
@AdamRoden I like "React to gaze movements instantly?" if it will fit. Slightly easier to infer the meaning.

AdamRoden:
The thing is, we always react instantly. For me, the most noticeable difference is when moving the cursor this setting controls whether the cursor teleports around the screen as you quickly change your target, or it glides across the screen from one point to the next. To understand what I mean try using move-and-click and rapidly switch your target before it clicks and observe how the cursor moves. Do it using Low with and without this setting.

That said, I think "Enable Instant Movement to Target" would more accurately convey meaning.

@JuliusSweetland @kmcnaught Would that be okay?

JuliusSweetland:
Nitpick, but I think I'd prefer the setting to be the user enabling some
extra smoothing. What about "Smooth when changing gaze target", or  "Smooth
movement when changing gaze target?

AdamRoden:
I like "Smooth when changing gaze target"

JuliusSweetland:
@kmcnaught Is there anything remaining on this PR? I'm happy to merge if there's nothing else (and all final changes we've discussed are in place)?

kmcnaught:
Nothing remaining as far as I'm aware.

JuliusSweetland:
@AdamRoden Nothing outstanding from us. Could you let me know when the final changes are all in place on this PR (e.g. "Smooth when changing gaze target") and I'll create a pre-release for testing. Thanks!

AdamRoden:
We neglected one thing.
How do we differentiate between "Data stream processing amount" and "Adaptive smoothing level" so they are distinguishable from a user perspective?
Would "Input device processing amount" and "Application smoothing level" be more appropriate?

JuliusSweetland:
Those work for me

AdamRoden:
Okay. The code with "Input device processing amount" and "Application smoothing level" and "Smooth when changing gaze target" is posted.

JuliusSweetland:
@AdamRoden @kmcnaught Lovely stuff guys. I'll create a pre-release tonight.

PR #737, opened 2020-10-10
Make look to scroll default

AdamRoden:
Take it for a test drive and let me know how it all feels. There were a lot of changes, so I won't be surprised to get a list of things to keep working on before calling this good.


JuliusSweetland:
@AdamRoden I didn't get a notification you'd pushed this. I'll take a look this evening...

JuliusSweetland:
Hey Adam,

It's looking really good. I like the new mouse keyboard layout. Overall it's pretty close - nice work. Here are the small things I spotted:

Vertical zone was not being drawn (in Horizontal scroll mode or cross mode)
Default back to free or cross?
Should any border and deadzone lines be drawn in free mode? There were none.
Should there be a preview in the MC of the boundary border colour and thickness settings? (I really like the deadzone border preview)
Automatically resume scrolling after performing mouse actions: was disabled, but scrolling kept resuming.

I really like the circular deadzone/free scroll indicator - looks cool. I changed it to Teal and 5 pixels - it looks good.

If you look for any references to "if (Settings.Default.LookToScrollEnabled && Settings.Default.LookToScrollShowOverlayWindow)" - change these to remove the "Settings.Default.LookToScrollEnabled && " part and the overlays get created, although the screen border lines don't seem to be drawn. I'm not 100% sure they should be though...

Again, great looking PR. Just a few issues to iron out and then it'll be there.

Thanks Adam!

JuliusSweetland:
@AdamRoden FYI I fixed the missing setting problem on this branch if you want to check anything: https://github.com/OptiKey/OptiKey/tree/AdamRoden-master

AdamRoden:
@JuliusSweetland Here is an update to address the issues you found.

---
>Vertical zone was not being drawn (in Horizontal scroll mode or cross mode)

I do not have this problem.

---
>Default back to free or cross?
Should any border and deadzone lines be drawn in free mode? There were none.

LTS Free is the default and it uses corner rounding to make the deadzone borders round.

JuliusSweetland:
Thanks Adam. I'll restore all of my default settings and will test again.

JuliusSweetland:
@AdamRoden I'll do some more testing later, but this looks great. One thing; the default colour of "Grey" (whilst it is the spelling in England) is not recognised and should be "Gray" for Microsoft's benefit. If you delete your user settings and try running Optikey it will crash on startup.

The only other suggestion I have would be to increase the thickness of the deadzone and (possibly) border lines. I think they look better when they are more deliberate. What do you think?

Looking great Adam!

AdamRoden:
@JuliusSweetland I think resume scrolling after a mouse action should also re-center the mouse when the center mouse at scroll location option is enabled. Otherwise, I'll change the color and thickness and look at adding some dynamic opacity adjustments so that as you scroll the border is apparent, but as you are still, it fades down.

JuliusSweetland:
Agree with all of that. Looking forward to seeing it!

AdamRoden:
@JuliusSweetland I cannot figure out why my opacity property is not making it from MVVM.LTS to LTS.xaml. Can you take a look and tell me what I should have done?

JuliusSweetland:
@AdamRoden I can't replicate the problem - I think it's all hooked up correctly. If you change MainViewModel.LookToScroll.cs L469 to L471 to the below you can see opacity changes as you look in/out of the deadzone circle in the centre:

```
                Opacity = 0.2;
                if (velocity.Length > 0 && Opacity < 1.0)
                    Opacity = 0.5;
```

AdamRoden:
@JuliusSweetland I did it like you and did a force push of all changes grouped into a single commit.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.LookToScroll.cs
```diff
@@ -59,6 +59,13 @@ public Thickness ActiveLookToScrollMargins
             private set { SetProperty(ref activeLookToScrollMargins, value); }
         }
 
+        private Timestamped<double> opacity = new Timestamped<double>(1.0, DateTime.Now);
```
JuliusSweetland, line 62: @AdamRoden What's the purpose of Timestamping the opacity double? It doesn't seem to be used.

AdamRoden: Good point. I had something else in mind when I set it up that way. No need now.

JuliusSweetland: Do you want me to wait for this correction from you, or I am happy to remove it after merging? I will only be able to look this evening though.

AdamRoden:
@JuliusSweetland Use my new commit. It implements the fadeout effect I was going for.

JuliusSweetland:
@AdamRoden I like the new fade in/fade out. Very nice PR. Thank you sir.

PR #741, opened 2020-10-19
Key Completion Review

AdamRoden:
@JuliusSweetland Here you can take a look at how I propose implementing this change. It's actually a nice step toward unity between built-in and dynamic keyboard capability. I suggest copying the current field into the new setting on application launch.

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/OptiKeyApp.cs
+++ b/src/JuliusSweetland.OptiKey.Core/OptiKeyApp.cs
@@ -115,6 +115,13 @@ namespace JuliusSweetland.OptiKey
                 Settings.Default.Reload();
             }
 
+            //Migrate the completion time setting to the new field
+            if (Settings.Default.KeySelectionTriggerFixationDefaultCompleteTimes == null)
+            {
+                Settings.Default.KeySelectionTriggerFixationDefaultCompleteTimes
+                    = Settings.Default.KeySelectionTriggerFixationDefaultCompleteTime.TotalMilliseconds.ToString();
+            }
+
             //Adjust log4net logging level if in debug mode
             ((Hierarchy)LogManager.GetRepository()).Root.Level = Settings.Default.Debug ? Level.Debug : Level.Info;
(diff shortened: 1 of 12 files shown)


JuliusSweetland:
I'll take a look as soon as possible, but I've got a full on week so please
bear with me.

src/JuliusSweetland.OptiKey.Core/OptiKeyApp.cs
```diff
@@ -115,6 +115,18 @@ protected void Initialise()
                 Settings.Default.Reload();
             }
 
+            //Migrate the completion time setting to the new field
+            var timespan = Settings.Default.KeySelectionTriggerFixationDefaultCompleteTime;
+            if (timespan> TimeSpan.Zero)
+            {
+                Settings.Default.KeySelectionTriggerFixationDefaultCompleteTimes = timespan.TotalMilliseconds.ToString()
+                    + "," + Math.Round(0.75 * timespan.TotalMilliseconds).ToString()
```
JuliusSweetland, line 123: I'm aiming to look at this PR (and your other one) in the next few days. Can you clarify what this logic is doing?

AdamRoden: I can clarify the comment before finalizing the PR, but the intent here was to convert the old single value Completion Time into a list with the new format to nudge users toward exploring the feature. Another idea is to add conditional highlighting to the field in MC and show a message that repeat rates can be added.

JuliusSweetland:
Ah I see. If this is for all key selection times then I wouldn't default
their old value to one which increases in selection speed. I fear some
users will not understand why keys are suddenly being selected multiple
times in faster succession. I think a mask on the textbox + a hint would go
a long way though.

AdamRoden:
I added additional language to help users understand what they can enter, highlighted the field if it does not have multiple values, and added input validation.

JuliusSweetland:
@AdamRoden Very nice PR. Merged. Thank you.

PR #749, opened 2020-11-17
EyeGesture Test

AdamRoden:
@JuliusSweetland I've put together a conceptual prototype of my idea. When you are focused on a point Optikey's cursor will appear. Then, you can glance left and back or right and back to do a left or right mouse click.

Give it a shot and then let's chat.

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Enums/EyeGestureStepTypes.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Enums/EyeGestureStepTypes.cs
@@ -1,0 +1,11 @@
+﻿// Copyright (c) 2020 OPTIKEY LTD (UK company number 11854839) - All Rights Reserved
+namespace JuliusSweetland.OptiKey.Enums
+{
+    public enum EyeGestureStepTypes
+    {
+        Fixation,
+        LookAtArea,
+        LookInDirection,
+        ReturnToFixation
+    }
+}
--- a/src/JuliusSweetland.OptiKey.Core/Enums/KeyCommands.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Enums/KeyCommands.cs
@@ -3,8 +3,8 @@ namespace JuliusSweetland.OptiKey.Enums
 {
     public enum KeyCommands
     {
-        Action,
         ChangeKeyboard,
+        Function,
         KeyDown,
         KeyToggle,
         KeyUp,
--- a/src/JuliusSweetland.OptiKey.Core/JuliusSweetland.OptiKey.Core.csproj
+++ b/src/JuliusSweetland.OptiKey.Core/JuliusSweetland.OptiKey.Core.csproj
@@ -205,6 +205,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="Enums\AppType.cs" />
+    <Compile Include="Enums\EyeGestureStepTypes.cs" />
     <Compile Include="Enums\KeyboardLayouts.cs" />
     <Compile Include="Enums\KeyCommands.cs" />
     <Compile Include="Enums\LookToScrollBounds.cs" />
@@ -260,6 +261,11 @@
     <Compile Include="Extensions\UiElementExtensions.cs" />
     <Compile Include="Extensions\WindowExtensions.cs" />
     <Compile Include="Models\DynamicKeyboardFolder.cs" />
+    <Compile Include="Models\XmlKeyCommand.cs" />
+    <Compile Include="Models\EyeGestureShape.cs" />
+    <Compile Include="Models\XmlEyeGestures.cs" />
+    <Compile Include="Models\EyeGesture.cs" />
+    <Compile Include="Models\EyeGestureStep.cs" />
     <Compile Include="Models\ILastMouseActionStateManager.cs" />
     <Compile Include="Models\KeyboardOutputServiceState.cs" />
     <Compile Include="Models\KeyCommand.cs" />
@@ -334,6 +340,7 @@
     <Compile Include="Observables\PointSources\IPointSource.cs" />
     <Compile Include="Observables\PointSources\MousePositionSource.cs" />
     <Compile Include="Observables\PointSources\PointServiceSource.cs" />
+    <Compile Include="Observables\TriggerSources\EyeGestureSource.cs" />
     <Compile Include="OptiKeyApp.cs" />
     <Compile Include="Properties\Resources.Designer.cs">
       <AutoGen>True</AutoGen>
@@ -511,6 +518,7 @@
     <Compile Include="UI\ViewModels\MainViewModel.cs" />
     <Compile Include="UI\ViewModels\ManagementViewModel.cs" />
     <Compile Include="UI\ViewModels\Management\DictionaryViewModel.cs" />
+    <Compile Include="UI\ViewModels\Management\GesturesViewModel.cs" />
     <Compile Include="UI\ViewModels\Management\PointingAndSelectingViewModel.cs" />
     <Compile Include="UI\ViewModels\Management\SoundsViewModel.cs" />
     <Compile Include="UI\ViewModels\Management\FeaturesViewModel.cs" />
@@ -876,6 +884,12 @@
     <Compile Include="UI\Views\ManagementView.xaml.cs">
       <DependentUpon>ManagementView.xaml</DependentUpon>
     </Compile>
+    <Compile Include="UI\Controls\GestureEditor.xaml.cs">
+      <DependentUpon>GestureEditor.xaml</DependentUpon>
+    </Compile>
+    <Compile Include="UI\Views\Management\GesturesView.xaml.cs">
+      <DependentUpon>GesturesView.xaml</DependentUpon>
+    </Compile>
     <Compile Include="UI\Views\Management\OtherView.xaml.cs">
       <DependentUpon>OtherView.xaml</DependentUpon>
     </Compile>
@@ -986,6 +1000,9 @@
     <None Include="NOTES AND DISCLAIMER.txt">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
+    <Content Include="Resources\EyeGestures\EyeGestures.xml">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
     <Content Include="Resources\DynamicKeyboards\AltAlphaHorizontalKeyboard.xml">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
@@ -1978,6 +1995,14 @@
     <Page Include="UI\Views\ManagementView.xaml">
       <SubType>Designer</SubType>
       <Generator>MSBuild:Compile</Generator>
+    </Page>
+    <Page Include="UI\Controls\GestureEditor.xaml">
+      <Generator>MSBuild:Compile</Generator>
+      <SubType>Designer</SubType>
+    </Page>
+    <Page Include="UI\Views\Management\GesturesView.xaml">
+      <Generator>MSBuild:Compile</Generator>
+      <SubType>Designer</SubType>
     </Page>
     <Page Include="UI\Views\Management\OtherView.xaml">
       <Generator>MSBuild:Compile</Generator>
--- a/src/JuliusSweetland.OptiKey.Core/Models/EyeGesture.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Models/EyeGesture.cs
@@ -1,0 +1,33 @@
+﻿// Copyright (c) 2020 OPTIKEY LTD (UK company number 11854839) - All Rights Reserved
+using System;
+using System.Collections.ObjectModel;
+using System.Windows;
+using System.Xml;
+using System.Xml.Serialization;
+
+namespace JuliusSweetland.OptiKey.Models
+{
+    public class EyeGesture
+    {
+        public EyeGesture() { }
+
+        public string Name { get; set; } = "New Gesture";
+
+        public double Cooldown { get; set; } = 2000;
+
+        [XmlIgnore] public bool enabled { get; set; } = false;
+
+        public string Enabled
+        {
+            get { return enabled.ToString(); }
+            set { enabled = !bool.TryParse(value, out var result) || result; }
+        }
+
+        [XmlElement("Step")] public ObservableCollection<EyeGestureStep> Steps { get; set; }
+
+        [XmlIgnore] public Point FixationPoint { get; set; }
+        [XmlIgnore] public int StepIndex { get; set; } = 0;
+        [XmlIgnore] public Point PointStamp { get; set; }
+        [XmlIgnore] public DateTime TimeStamp { get; set; } = DateTime.Now;
+    }
+}
--- a/src/JuliusSweetland.OptiKey.Core/Models/EyeGestureShape.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Models/EyeGestureShape.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) 2020 OPTIKEY LTD (UK company number 11854839) - All Rights Reserved
+
+using System.Windows.Media;
+
+namespace JuliusSweetland.OptiKey.Models
+{
+    public class EyeGestureShape
+    {
+        public Brush Background { get; set; }
+        public double Radius { get; set; }
+        public double X { get; set; }
+        public double Y { get; set; }
+        public double Left { get; set; }
+        public double Top { get; set; }
+        public double Width { get; set; }
+        public double Height { get; set; }
+    }
+}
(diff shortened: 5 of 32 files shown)


JuliusSweetland:
That's a really interesting demo. Initial rough thoughts:

1.This would need to be disabled when using the mouse for input (mouse as primary point source, and manual mode).
2.The UX could be more intuitive. I think the idea that Optikey has recognised that there is focus on a point could be done without the cursor, and remain until the user completes or discards the gesture. What do you think about something akin to the Tobii context menu which gives you hints on what you can achieve, as well as feedback on how far/how long you need to look to complete gestures?
3.We should consider gestures or settings that include the magnifier - it might be good to gesture a rough position and then use a magnified image with either:
a)another gesture 
or 
b)the usual selection mechanic on/outside the magnified image
4.This could be turned on/locked on by a key. That gives the user more control to disable this when they don't want it.

Compelling stuff though.

AdamRoden:
@JuliusSweetland Yes to all those things. This code is not how the feature would be built, it was just the easiest way to demo the idea. I can either keep working on this if you think it's worthwhile, or I can put it on standby and focus my energy elsewhere.

JuliusSweetland:
I think it is worthwhile, yes.

What else is on the maybe pile? I've had a few requests for the ability to
pause/sleep when magnifying to give the user a chance to read the magnified
section. This could be achieved with an optional sleep key next to the
magnification, with a setting to turn on this behaviour?

AdamRoden:
@JuliusSweetland I pushed my current state of code. It has issues that need to be resolved but is a big step toward the end goal.
EyeGesture is a collection of steps.
Each step has a position, shape, and timeout.
At the completion of each step, you can trigger events.
Events are currently limited to FunctionKeys, but with further work they will be mutual in function to dynamic keys. (I need to update the xml serialization.
My parallel ambition is to use the same concepts of my EyeGesture management gui to build a dynamic keyboard management gui.

JuliusSweetland:
@AdamRoden Cool - I'll give it a try this evening.

JuliusSweetland:
@AdamRoden I've just had a quick look and I'm a bit stuck! What do I do to get started?

image

Also I think the Gestures tab should move in before "Extra Features".

AdamRoden:
@JuliusSweetland I added an EyeGestures.xml file to Resources, but had problems auto copying it from there to file system, so broken for now. I will move it to Extra Features.

AdamRoden:
[new 5.txt](https://github.com/OptiKey/OptiKey/files/5745334/new.5.txt)

JuliusSweetland:
@AdamRoden Sorry for the delay - even on holiday I'm a bit swamped with work and family stuff!

The PR looks good. I tested by loading the 5.txt file and then (using the mouse, just for testing) moved between areas 1 and 2 to trigger the two different actions. It works well. Here are a few bits of feedback:

1.Set the Build Action on the bundled EyeGestures.xml resource to Content (from Resource) and the file will be copied to the output folder. From there you can choose to copy the file (and other files in the run location\Resources\EyeGestures folder) to the C:\Users\USER_NAME\AppData\Roaming\Optikey\Optikey\EyeGestures folder. We do this with dynamic keyboards, symbols, dictionaries, etc. Eg:
image

2.Some clarifications: what do Floating and Start Speed do?
image

3.What does EVE in "ADD EVE" mean?

4.I know this is a work in progress, but there is no visual effect from clicking Enabled or Round (or floating, but I don't know what that should do).

AdamRoden:
@JuliusSweetland I could really benefit from your expertise to help get the wpf stuff working well. 
I struggled with how to pass a parameter with a button click using mvvm.
I am also struggling with how to have a preview that will scale properly and update as you update properties. Right now I'm using grids created programmatically because canvas doesn't scale, but that I'd rather do it all in wpf.

JuliusSweetland:
@AdamRoden Sure. I can probably take a look later on today. Can you give me a specific query, e.g. when I click _this_ button I want to pass this info _here_ so I can use it for _this_

JuliusSweetland:
Hi Adam. I like the new layout - it's cleaner although (I'm sure you're aware) the previews are not showing:

image

I've looked over how things are structured. If we take the example of how previews work you have the XmlEyeGestures collection in the GesturesViewModel. An XmlEyeGestures then contains an EyeGestures collection which builds its own GesturesPreview.

Anything in the view model (including models like XmlEyeGestures and EyeGestures) should not contain any UI logic, or anything UI specific. Instead I would expose properties from the models which fully describe the preview, e.g. where the areas are, whether they are rounded, etc. You then bind your view (or controls insides the view) to this data. A possible solution would be to copy something like the CursorPopup code behind logic - in the OnLoaded method (which you should hook up to run your logic, but then unhook as it probably only needs to run once - see KeyboardHost for an example of only handling OnLoaded once as it unhooks the event handler at the end of the handler) you attempt to cast the DataContext to the GesturesViewModel, and then you can bind to it (e.g. gesturesViewModel.OnPropertyChanges()...). You then move the UI logic there, and build the visual representation when the data model changes.

Does that make sense?

AdamRoden:
@JuliusSweetland "Does that make sense?"
Yes and no. It is a good hint to steer me in the right direction. 
Your previews do not show up because your file is not compatible with my update. You will need to copy the latest file from /Resources/EyeGestures. I don't yet have good errors and warnings for bad files. I will continue working today.

JuliusSweetland:
Ah I see. I'll try the new file. Thanks.

Let me know if you have other questions. And Happy New Year!

AdamRoden:
@JuliusSweetland It didn't quite click. I'm still lost.

JuliusSweetland:
@AdamRoden So I'd break it down as:

1.When you click the button to add a zone the command should trigger logic to add a zone object to the model, e.g. add an EyeGazeZone to a collection of EyeGazeZones in the GestureViewModel (or in a child property).

2.The constructor in the code behind the GestureView control attached an event handler to the OnLoaded event

3.The OnLoaded event attaches property change handlers to properties on the GestureViewModel (case from the DataContext). For now it would just listen to the EyeGazeZones changes. The same handler should unhook itself as part of the OnLoaded handler (e.g. "this.OnLoaded -= OnLoadedHandler" if "OnLoadedHandler" is the name of the local function or method setting up the property change listener).

4.When EyeGazeZones collection changes are detected, or changes to any item in the EyeGazeZones collection, then redraw based on that collection.

Is that a bit clearer?

AdamRoden:
>OnLoaded event attaches property change handlers to properties on the GestureViewModel (case from the DataContext)

@JuliusSweetland This is the piece I'm not sure about. Specifically, what you mean by "(case from the DataContext)"? How does that translate to code?

JuliusSweetland:
Sorry there was a typo in there. CAST from DataContext to the expected view
model type. Have a look at KeyboardHost for an example of only handling
OnLoaded once as it unhooks the event handler at the end of the handler.

AdamRoden:
@JuliusSweetland I've been tuning away and while there is always room for improvement, especially with this particular set of features, I'd like you to sample the new GUI in MC along with the new file. It will give you a good feel for things.

JuliusSweetland:
Hey Adam. Absolutely. Let me know when the PR is ready for a spin. I get
updates when you push, but the last one didn't build so I assumed it was
in-progress.

Hope you're doing well.

AdamRoden:
@JuliusSweetland Try, try, try again. My code keeps violating the unit tests, so although it works, I keep having to revise how I achieve my goal. :-/

AdamRoden:
@JuliusSweetland All checks passed. Give it a whirl. I have a few big ideas, so getting your feedback will help me prioritize my next actions.

JuliusSweetland:
I'll check it out this weekend. Thanks Adam.

JuliusSweetland:
@AdamRoden Hey - I've had a look over things. The new gesture previews look great - nice one.

Here are a few things to discuss:

1.When selecting config file location the location text is not updated
2.Should ADD STEP not add at the end of the sequence? If you edit an existing set of 2 steps and add another you get a new step 1, which seems wrong.
3.I found "Edit from file", "Save to file", "Edit from settings", "Activate changes" confusing. The difference between edit from file and settings may be important to my understanding, but I would think the flow should be to Enable Gestures, which will automatically load from a file (or create one if no settings file exists). The user then edits, and when they click the general OK button for the Management Console then the settings should be saved to the EyeGestures.xml (or whatever is selected). I don't think users will understand loading manually and activating manually, etc.
4.The unit tests are failing as the GestureViewModel was extending a ContentControl which is a UI element and can only be created on the STA thread (the main UI thread). I have fixed this. There are a few small issues with the same VM, e.g. the visibilitity properties: these should not be of type Visibility in the VM, but booleans instead;  i.e. IsVisible. You then bind the Visibility property in the View to the VM property and use a ValueConverter to convert the value to a Visibility (e.g. the InvertedBoolToVisibilityConverter). If you delete the references at the top of the file to these then you will see which UI elements need to be refactored out:

using System.Windows;
using System.Windows.Controls;
using System.Windows.Media;

JuliusSweetland:
@AdamRoden FYI my small fix is on the latest commit on this branch: https://github.com/OptiKey/OptiKey/tree/AdamRoden-master

AdamRoden:
@JuliusSweetland Good feedback. Thank you! I will clean those things up and try again.

JuliusSweetland:
Cool. You know where I am if you have any questions.

JuliusSweetland:
@AdamRoden Please let me know when you want me to test again. I get notifications for commits, but I don't know when you're ready so I'll wait to hear from you.

AdamRoden:
You can have another look at the UI now. It is a better, but I tried a couple fresh looks that you may want to discuss.

JuliusSweetland:
Hey @AdamRoden - this is a cool step forward. Questions:

1.Should the On/Off button in the top left do anything? It appears to be linked to playing/stopping.
2.What is the play/stop doing? I presume it's running through the sequence so the user can see what they have defined?
3.I think SAVE FILE should be implicit when clicking OK on the management console, rather than allowing the possibility for users to do lots of work, click ok and lose their progress.
4.In the top of my screenshot you can see that a startup file is missing - what is this?
5.If the action is to "Change Keyboard" what should the user enter in the free text box? Is this a keyboard name? If so would it be better to supply a list of valid keyboards?

The change to insert and re-order the steps looks good. As does the responsiveness of the UI. Great stuff!

image

AdamRoden:
@JuliusSweetland 
> 1.Should the On/Off button in the top left do anything? It appears to be linked to playing/stopping.
> 2.What is the play/stop doing? I presume it's running through the sequence so the user can see what they have defined?
> 3.I think SAVE FILE should be implicit when clicking OK on the management console, rather than allowing the possibility for users to do lots of work, click ok and lose their progress.
> 4.In the top of my screenshot you can see that a startup file is missing - what is this?
> 5.If the action is to "Change Keyboard" what should the user enter in the free text box? Is this a keyboard name? If so would it be better to supply a list of valid keyboards?
1. I just thought it served as a nice indicator of whether or not gestures were running. I could use that with no start top buttons. What style do you prefer?
2. Play/stop activates the changes without permanently saving them.
3. OPEN/SAVE need to be relabeled. They are really IMPORT/EXPORT. The OK button is what commits the changes internally to Settings.
4. I think that happens when you have an dynamic keyboard that with an improperly closed xml tag somewhere.
5. Ideally it would a dropdown, but I didn't figure out how to also support dynamic keyboards.

I think what is really needed is a new list of command types for the UI (This would also be used for dynamic keyboards):
- Change Keyboard (Sub options: Built in, Dynamic, BackAction)
- Change Language (Sub option: dropdown of languages)
- Change Setting NEW TBD (Sub option: dropdown of allowed settings and a textbox for the new value)
- Function (Optikey specific trimmed down version of current list. Add Wait to this list. Strip out language, keys, keyboard, and mouse types. Sub option: textbox for arguments)
- Keyboard Input (Sub options: a dropdown of up/down/toggle/press and a dropdown of virtual key codes)
- Mouse Input (Stripped from Function List)
- Text
- (Not sure how I want to handle Loop and Plugin as they are complex advanced features).

JuliusSweetland:
>     1. I just thought it served as a nice indicator of whether or not gestures were running. I could use that with no start top buttons. What style do you prefer?
I think I prefer the play/stop style. It's clear that this will run the actions. Separately is there an option to enable/disable gestures across the board? That would be good. If disabled the rest of the controls should be disabled (greyed out) to reflect the state. That makes everything very clear to the users.

>     2. Play/stop activates the changes without permanently saving them.
Do you mean it dry runs through the gesture stages?

>     3. OPEN/SAVE need to be relabeled. They are really IMPORT/EXPORT. The OK button is what commits the changes internally to Settings.
Ok that makes sense.
 
>     4. I think that happens when you have an dynamic keyboard that with an improperly closed xml tag somewhere.
Understood.

>     5. Ideally it would a dropdown, but I didn't figure out how to also support dynamic keyboards.
I guess it would be a combination of the derived list of keyboard enum values and the logic which detects all loaded dynamic keyboards to build the dynamic keyboards menu. This would be done in the VM to build a list which a drop down could bind to.
 
> I think what is really needed is a new list of command types for the UI (This would also be used for dynamic keyboards):
> 
>     * Change Keyboard (Sub options: Built in, Dynamic, BackAction)
> 
>     * Change Language (Sub option: dropdown of languages)
> 
>     * Change Setting NEW TBD (Sub option: dropdown of allowed settings and a textbox for the new value)
> 
>     * Function (Optikey specific trimmed down version of current list. Add Wait to this list. Strip out language, keys, keyboard, and mouse types. Sub option: textbox for arguments)
> 
>     * Keyboard Input (Sub options: a dropdown of up/down/toggle/press and a dropdown of virtual key codes)
> 
>     * Mouse Input (Stripped from Function List)
> 
>     * Text
> 
>     * (Not sure how I want to handle Loop and Plugin as they are complex advanced features).
That sounds like a great solution.

AdamRoden:
@JuliusSweetland 
> I think I prefer the play/stop style. It's clear that this will run the actions. Separately is there an option to enable/disable gestures across the board? That would be good. If disabled the rest of the controls should be disabled (greyed out) to reflect the state. That makes everything very clear to the users.

Okay, new plan. The Play/Stop will be moved to the Editor section and be used to test the selected gesture. Then, the toggle at the top will serve as a global toggle to enable/disable all gestures.

JuliusSweetland:
Yes I like that, but could we change the style of the enable/disable to a
standard checkbox which greys out all other controls when disabled? This
will be in line with other instances.

AdamRoden:
@JuliusSweetland How about this look?

image

image

JuliusSweetland:
@AdamRoden New design is looking great. Bit more feedback...

1.Stop ALL / Play ALL - what do you think about removing the "ALL"? I presume the play would trigger all steps? Am I misunderstanding?

2.Should the gesture's "Active" checkbox be triggered by the play/stop buttons? What does the active flag mean?

image

3.What should the PLAY button do? If I import the EyeGestures.xml file and select the 'Minimise' action should the play button result in the active window being minimised, or should I have to move the cursor to the zones to trigger the action?

AdamRoden:
@JuliusSweetland Okay, fixes for Minimise bug and the startup keyboard bug are both interwoven with additions for Eye Gestures. I updated the UI to be a bit more clever. The Start/Stop buttons are merely substitution for enable/disable checkboxes. I prefer the button style for a larger target and visual. 

image

JuliusSweetland:
@AdamRoden The new layout looks great, but I'm still confused by what "Activate" actually means, and the difference between "Activate All" and activating a single set of steps?

AdamRoden:
@JuliusSweetland Is this intuitive enough?

image

JuliusSweetland:
Ah yes, that's really good. UX is quite a tricky thing - something which
makes total sense to one person ends up confusing another. I think you've
nailed it here!

AdamRoden:
@JuliusSweetland One more round of UI tweaks in the bag. Do you think this is seeming polished enough prep for merge?

image

JuliusSweetland:
@AdamRoden I think this is basically ready. 2 small requests:

1.Could we change the stop icon to a cross, so it's a cross and a tick. In which case do you think a standard checkbox should be used? I don't mind either way - it's just a simplification.

2."Eye Gesture State" - if this is just a summary of how many gestures are enabled for the user then what about "Enabled Eye Gesture Count", or "Enabled Eye Gestures"?

Otherwise looking great.

AdamRoden:
@JuliusSweetland I like those two notes. Done.
Also, is there a good reason for having Other and Extra features as 2 tabs instead of 1. After trimming LTS complexity I saw fit to merge those two views, which is why the latest check failed. I can split them again, but I'd rather not.

JuliusSweetland:
No real need for other and extra anymore, no. I'll aim to check your PR
this evening.

JuliusSweetland:
@AdamRoden I've merged and fixed all merge issues (and a broken test). The new version is here: https://github.com/OptiKey/OptiKey/tree/AdamRoden-master

Could you please sanity check and then I'll merge in. Thanks!

AdamRoden:
> @AdamRoden I've merged and fixed all merge issues (and a broken test). The new version is here: https://github.com/OptiKey/OptiKey/tree/AdamRoden-master
> 
> Could you please sanity check and then I'll merge in. Thanks!

@JuliusSweetland The version history has some pointless commits that have been overwritten, but the end result is good.

PR #760, opened 2021-02-27
Multiple asserts and use of NUnit 3.0's Constraint Model of assertions

khusain:
In NUnit 3.x, assertions are primarily written using constraint-based Assert model. (see https://docs.nunit.org/articles/nunit/writing-tests/assertions/assertion-models/constraint.html). There is no functional difference between the classic and constraint-based model. The new way improves readability of tests and makes it clear what is the actual value and what the expected value should be.

In addition, NUnit has the concept of multiple asserts. (see https://docs.nunit.org/articles/nunit/writing-tests/assertions/multiple-asserts.html). Usually when an assert fails, the test terminates. Most of the tests in this solution do not depend on the passing of the previous test(s). As a result, the Assert.Multiple allows us to execute all tests within the block and collect all failures within a method so they may all be fixed at once.

### DIFF
--- a/src/JuliusSweetland.OptiKey.UnitTests/UI/ViewModels/MainViewModelSpecifications/WhenConstruct.cs
+++ b/src/JuliusSweetland.OptiKey.UnitTests/UI/ViewModels/MainViewModelSpecifications/WhenConstruct.cs
@@ -20,9 +20,9 @@ namespace JuliusSweetland.OptiKey.UnitTests.UI.ViewModels.MainViewModelSpecifica
 
         protected override void Act()
         {
-            MainViewModel = new MainViewModel(AudioService.Object, CalibrationService.Object, DictionaryService.Object, 
-                KeyStateService.Object, SuggestionService.Object, CapturingStateManager.Object, LastMouseActionStateManager.Object, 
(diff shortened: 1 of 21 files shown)


src/JuliusSweetland.OptiKey.UnitTests/UI/ViewModels/MainViewModelSpecifications/WhenConstruct.cs
```diff
@@ -547,8 +547,8 @@ protected override void Arrange()
         [Test]
         public void ThenKeyboardShouldBeSizeAndPosition()
         {
-            Assert.IsNotNull(MainViewModel.Keyboard);
-            Assert.AreEqual(typeof(OptiKey.UI.ViewModels.Keyboards.SizeAndPosition), MainViewModel.Keyboard.GetType());
+            Assert.That(MainViewModel.Keyboard, Is.Not.Null);
```
khusain, line 550: This is one of the few cases where it's incorrect to use the Assert.Multiple method. We need MainViewModel.Keyboard to be NOT NULL for the following assert.

JuliusSweetland:
@khusain Thanks very much. Nice PR.

PR #764, opened 2021-03-12
Add MoveWindow DynamicKey

AdamRoden:
@JuliusSweetland This a new dynamic key that is a necessary prerequisite to supplying my very best dynamic keyboard examples. It is also a stepping stone to some more comprehensive improvements I have in the works.

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Enums/KeyCommands.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Enums/KeyCommands.cs
@@ -9,6 +9,7 @@ namespace JuliusSweetland.OptiKey.Enums
         KeyToggle,
         KeyUp,
         Loop,
+        MoveWindow,
         Plugin,
         Text,
         Wait
--- a/src/JuliusSweetland.OptiKey.Core/JuliusSweetland.OptiKey.Core.csproj
+++ b/src/JuliusSweetland.OptiKey.Core/JuliusSweetland.OptiKey.Core.csproj
@@ -992,7 +992,17 @@
     <Content Include="Resources\DynamicKeyboards\AltAlphaVerticalKeyboard.xml">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
-    <Content Include="Resources\DynamicKeyboards\Samples\AR16-Qwerty.xml">
+    <Resource Include="Resources\DynamicKeyboards\BigHackerKeyboard.xml" />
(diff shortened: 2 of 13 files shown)


AdamRoden:
Well, this is supposed to be a draft. I need to also include the sample keyboards... Working on them soon.

AdamRoden:
@JuliusSweetland New keyboards supplied. Good to go.

JuliusSweetland:
@AdamRoden Could you please give me a bit of an overview of what this change achieves? How does the MoveWindow key work? Are these changes backward compatible or will users with existing dynamic keyboards have to recreate them?

Thanks

AdamRoden:
@JuliusSweetland It is fully backward compatible. Here's an example. Using the key defined below when browsing the internet would resize your web browser to take up the top left quarter of your screen. I use it in combination with Floating Optikey keyboards as an alternative to using Docking.
> \<DynamicKey Col="0" Row="0">\<Label>Resize Window\</Label>\<MoveWindow>0,0,50%,50%\</MoveWindow>\</DynamicKey>

JuliusSweetland:
@AdamRoden Thanks Adam. I will hold off from releasing until you've had a chance to look into this.

AdamRoden:
@JuliusSweetland I'm disappointed in myself because I know what causes the issue and it seems like the fix should be simple, but I cannot find the answer to fix it. 

The problem (which also affects look to scroll) is that when the code looking for the front-most window is running, it finds Optikey as the top-most window. The main reason for this is because I created an Overlay window to follow the user's gaze. Even though it is transparent like the look to scroll overlay, somehow it always results in Optikey becoming the active window on the taskbar. I feel like I should be able to use the user32.dll and window styles to filter, but my attempts to fix it have not been successful.

JuliusSweetland:
@AdamRoden I wouldn't be disappointed in yourself - this stuff can be really fiddly. Is there a way to obtain all windows under the point and then order by z axis, or ignore the handle of the overlay window? Otherwise what about minimising the overlay before locating the window and then restoring? Would something like that work?

AdamRoden:
@JuliusSweetland I fixed it by making the gaze indicator overlay a popup window and by exposing the window handle in Look to scroll to filter it. (It also helps to close my other instance of Optikey when testing.)

JuliusSweetland:
Great work!

JuliusSweetland:
@AdamRoden Merged

Issue #676, opened 2020-05-13
Changing font gives error fields


kmcnaught:
Just a small one, mainly aesthetic. When a font family is changed in the Management Console and the current font weight / stretch is no longer applicable, the fields go red with a `Value "" could not be converted` error. Ideally these fields would revert to a sensible default if the current value isn't supported. 

Video explains it better:
https://www.loom.com/share/4a3826de7816483c83c77a4637465f26

Note that Optikey continues to function fine, so this is fairly minor. This was present in 2.32.0 also, so it's not related to any management console refactorings in the 2->3 upgrade.

kmcnaught:
Thanks @SCEhardt

PR #779, opened 2021-12-15
Fixes for 676 & 747

SCEhardt:
Tested in Pro and Mouse

resolves 747


SCEhardt:
Added a fix for https://github.com/OptiKey/OptiKey/issues/676

PR #791, opened 2022-01-27
Add XInput/DirectInput trigger sources

kmcnaught:
Adds support for controllers/joysticks/gamepads using the DirectInput or XInput APIs


kmcnaught:
@annakirkpatrick have you had a chance to test this out at all?

annakirkpatrick:
@kmcnaught unfortunately, no, I haven't had a chance to try it out yet. But it is absolutely on my to do list! I'll get back to you.

kmcnaught:
Ooh, interesting! Thanks for the report, I will investigate this end as well. I had another user report that her Xbox Adaptive Controller didn't work until she restarted her PC (in addition to Optikey) so it sounds like something in the initialisation

kmcnaught:
PS - If you get into a not-working state again could you please save the logs for me

annakirkpatrick:
@kmcnaught I installed on a new computer and replicated the initialization issues. Logs are attached. I think the log.2 is the one where the error actually happened. But I will also include the log.3 that came before it and .log from when it worked after restart.

Here's the sequence of events.

1. I ran the new Optikey, opened the management console, and set both selection methods to direct input and button 1. At this point I realized I also need to configure my switch interface, and Optikey was getting in the way. So I closed Optikey. I think this is .log.3

2. After configuring my switch interface, I ran Optikey again. I tried selecting several different keys. My button presses were not recognized. I opened the management console and selected the specific name of my switch interface, as opposed to the general "all direct input controllers." This should be log.2


3. Automatic restart. And now everything seems to work.

Note: renamed log files so extension was .log, for github.
[OptiKeyPro.log](https://github.com/OptiKey/OptiKey/files/8073758/OptiKeyPro.log)
[OptiKeyPro.2.log](https://github.com/OptiKey/OptiKey/files/8073759/OptiKeyPro.2.log)
[OptiKeyPro.3.log](https://github.com/OptiKey/OptiKey/files/8073762/OptiKeyPro.3.log)

annakirkpatrick:
@kmcnaught I think I found another bug. When I select DirectInput for only the key selection source but leave the point selection source as dwell, the combo box to select the button doesn't appear.  See screenshot: 
OptiKeyDirectInput1

After changing point selection to also be DirectInput, the combo box for the button selection shows up (for both key and point):
OptiKeyDirectInput2

Log from when I was changing these settings:  [OptiKeyPro.2.log](https://github.com/OptiKey/OptiKey/files/8080575/OptiKeyPro.2.log)

Log after restart:  OptiKeyDirectInput2

Additional note: This was right after installing on yet another computer.  This time I selected my specific switch interface (instead of "all directinput controllers" right from the start. And the issue with not recognizing my switch presses didin't occur.  Not sure if it's related.  But perhaps the selection of the specific DirectInput device is relevant to the initialization issues....

kmcnaught:
Thanks, I've fixed the UI issue now (copy paste error!) and will investigate the initialisation issue further. I think it's only happening when "all directinput controllers" is used for both key and point source; having the two sources in parallel isn't working in that scenario

JuliusSweetland:
@kmcnaught FYI I'll hold off creating the release for this PR to be resolved. Pls @ me when there's a new version so I don't miss it. Thanks.

annakirkpatrick:
>  I think it's only happening when "all directinput controllers" is used for both key and point source; having the two sources in parallel isn't working in that scenario

This is consistent with the behavior I've seen.

kmcnaught:
@JuliusSweetland this is ready for your review, I will separately send a pre-release to @annakirkpatrick for testing also. 

fyi I have added a console app which just logs information from any controller events so people can have discover what API their hardware uses and check whether it's working outside of Optikey. This was very useful for initial diagnosis with @annakirkpatrick  but might be mainly superceded by the automatic detection / user feedback in the management console. Though I expect it's still useful in the case where someone's switch doesn't actually present itself as a gamepad controller at all. If we _are_ going to include it then we'll mention it in wiki documentation, but I don't mind if it gets dropped.

kmcnaught:
PS - I would need to bundle the test console app into the installer if you give me the go-ahead.

JuliusSweetland:
@kmcnaught Yes please bundle the console app. I will prepare the release test candidate once this is ready to merge. Thanks.

kmcnaught:
@JuliusSweetland now ready with bundled `TestControllers.exe`, please go ahead with pre-release

Some comments on installer-related commits:
- I committed the general file changes that were automatically picked up in AdvancedInstaller, feel free to drop that commit/recreate it yourself if you prefer
- I encountered an error installing: `"Source file not found .... OptikeyPro1.cab". Verify that the file exists and you can access it"`. According to https://www.advancedinstaller.com/forums/viewtopic.php?t=35885#p100391 this is an AdvancedInstaller race condition which I resolved by setting the property `ExtractFilesFirst`. Not sure if you've ever seen this? I once saw it in the wild with a fork I deployed.

JuliusSweetland:
@annakirkpatrick @kmcnaught The pre-release version of Optikey v3.2.0 is finally ready for testing if you have time to kick the tyres? https://github.com/OptiKey/OptiKey/releases/tag/v3.2.0_pre-release-1

Issue #787, opened 2022-01-05
Optikey is hanging on master branch


kmcnaught:
Using recent (undeployed) code on master, Optikey is very very laggy, quickly becoming completely unresponsive. I can't even right click on it to quit cleanly. 

I did a git bisect and the issue is introduced in c10fe5a9e8d. Looking at Task Manager, it's clear that Optikey is firing on all cylinders. I profiled the code before and after and have included some screenshots. 

### Profile without issue 
0898bd5e
![profile_before](https://user-images.githubusercontent.com/12151404/148300469-57d62358-6a60-494c-a919-ccd3a7d5b799.PNG)

### Profile with issue 
c10fe5a9e  - truncated because I am unable to quit cleanly so the profiler doesn't capture the whole session
![profile_after](https://user-images.githubusercontent.com/12151404/148300574-3c587724-cafa-4395-8191-490b476d8afe.PNG)

This looks suspicious:
![profile_deserialize](https://user-images.githubusercontent.com/12151404/148300592-239cb593-bf32-44dd-90e6-f1d45b5eb3d0.PNG)

Maybe points at it being to do with my settings file?? I blew away all my config files for this testing, but it's possible I had "accidentally good" config previously, if that makes sense?

I also see these logs in the Visual Studio output (running c10fe5a9e8d). 

> Exception thrown: 'System.Xml.XmlException' in System.Xml.dll
Exception thrown: 'System.Reflection.TargetInvocationException' in mscorlib.dll
Exception thrown: 'System.InvalidOperationException' in System.Xml.dll
Exception thrown: 'System.ArgumentNullException' in System.Core.dll

Weirdly I've been running this code for weeks now and the problem has only just arisen - I haven't been able to track down what has changed in my setup today. Some code I deployed yesterday is still working fine - but if I run the same code from Visual Studio I get the bad behaviour. 

I'm hoping this is to do with a bad state wrt settings rather than a fundamental performance limitation in the new code. I'd appreciate some other eyes on it.

kmcnaught:
Here's an excellently recorded phone video of Task Manager for a more visceral demonstration of the performance issue...

https://user-images.githubusercontent.com/12151404/148301654-c5ba5343-cf00-4734-8bd4-516035652312.mp4

kmcnaught:
[...]

PR #794, opened 2022-02-10
Fixes to gestures

### LINKED ISSUE #786: Dynamic keyboard attribute "BackReturnsHere" is wrong way round again
Since c10fe5a9e (@AdamRoden) the `BackReturnsHere` attribute on a `ChangeKeyboard` is flipped. 

See sample keyboard:
[test_back.zip](https://github.com/OptiKey/OptiKey/files/7807034/test_back.zip)

kmcnaught:
- Fixes loading of default gestures file if none found
- Add other guards (null values, missing directory, ..)
- Fixes saving of gestures from Management Console
- Prevents continuous deserialisation of invalid string/file (caused #787)
- Fixes #786

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Models/XmlEyeGestures.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Models/XmlEyeGestures.cs
@@ -1,4 +1,5 @@
 ﻿// Copyright (c) 2022 OPTIKEY LTD (UK company number 11854839) - All Rights Reserved
+using System;
 using System.Collections.ObjectModel;
 using System.IO;
 using System.Xml;
@@ -32,15 +33,17 @@ namespace JuliusSweetland.OptiKey.Models
 
         public static XmlEyeGestures ReadFromString(string xmlString)
         {
+            if (String.IsNullOrEmpty(xmlString)) { return null; }
+
             var gestures = new XmlEyeGestures();
             var serializer = new XmlSerializer(typeof(XmlEyeGestures));
             try
             {
-                gestures = (XmlEyeGestures)serializer.Deserialize(new StringReader(xmlString));
+                return (XmlEyeGestures)serializer.Deserialize(new StringReader(xmlString));                
             }
-            catch { }
-
-            return gestures;
+            catch {
+                return null;
+            }
         }
 
         public void WriteToFile(string filename)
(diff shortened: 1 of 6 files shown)


src/JuliusSweetland.OptiKey.Core/UI/ViewModels/Management/GesturesViewModel.cs
```diff
@@ -115,24 +119,35 @@ private void Load()
...
-                    break;
+                    return;
+            }
+            try
+            {
+                xmlEyeGestures = XmlEyeGestures.ReadFromFile(tempFilename);                
+            }
+            catch (Exception e)
+            {
+                Log.Error($"Error reading from gesture file: {EyeGestureFile} :");
```
JuliusSweetland, line 146: Should this not be logging the tempFilename as the EyeGestureFile has not yet been updated?

src/JuliusSweetland.OptiKey.Core/Models/XmlEyeGestures.cs
```diff
@@ -21,30 +22,28 @@ public static XmlEyeGestures ReadFromFile(string inputFilename)
...
+
             var gestures = new XmlEyeGestures();
             var serializer = new XmlSerializer(typeof(XmlEyeGestures));
             try
             {
-                gestures = (XmlEyeGestures)serializer.Deserialize(new StringReader(xmlString));
+                return (XmlEyeGestures)serializer.Deserialize(new StringReader(xmlString));                
+            }
+            catch {
+                return null;
```
JuliusSweetland, line 45: Would it be too spammy to log an error here to help diagnose issues later?

JuliusSweetland:
@kmcnaught Hey Kirsty. Sorry for the late review - I didn't get a notification that this PR was ready. I've added 2 very minor comments. Could you let me know and I'll merge once I have your replies? Thanks.

src/JuliusSweetland.OptiKey.Core/UI/ViewModels/Management/GesturesViewModel.cs
kmcnaught: yes good spot

src/JuliusSweetland.OptiKey.Core/Models/XmlEyeGestures.cs
kmcnaught: I think that would be fine but let me double-check

kmcnaught:
I've also just realised the bundled sample gestures file has all the gestures enabled, which overrides `Settings.Default.EyeGesturesEnabled=false`. That's both a bug (I think the global enabled state should override the loaded file if it differs) and something to change in the bundled file.

kmcnaught:
Ready for re-review

JuliusSweetland:
@kmcnaught Excellent, thank you very much!

PR #806, opened 2022-03-09
Performance improvement

AdamRoden:
@JuliusSweetland @kmcnaught 
This lightens the load. Nice job identifying that with profiler, Kirsty.


src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.LookToScroll.cs
```diff
@@ -20,6 +20,7 @@ namespace JuliusSweetland.OptiKey.UI.ViewModels
     partial class MainViewModel : ILookToScrollOverlayViewModel
     {
         private const int WheelUnitsPerClick = 120;
+        private bool lookToScrolActive;
```
kmcnaught, line 23: This variable name is a bit confusing, could it be renamed `previouslyActive` ? Then the condition `if (previouslyActive or active) ` would read much more logically

PR #822, opened 2022-03-29
Feature/try get

kmcnaught:
I've kept commits separate to show clearly what's changed, I will squash them when approved.

Two mitigations:
- Makes overrides dictionary concurrent
- Uses TryGet pattern to avoid race conditions

I've confirmed the behaviour is the same as before my changes (except hopefully without the sporadic crashes🙂)

Re: the concurrent dictionary, I'm not sure this is necessary so I'm happy to drop that change if you prefer. 

I hooked up a property changed handler (in testing) to `overrideTimesByKey`. As far as I can tell, it is only ever changed via `KeyboardHost:GenerateContent` which clears the dictionaries and then re-populates with any overrides from the dynamic keyboard XML. This is always on the same thread.

However, the `overrideTimesByKey` dictionary is exposed via the `IFixationTriggerSource` interface and widely accessible - `KeyboardHost`'s dictionary points to `mainWindow`'s dictionary which points to `inputService`'s dictionary which points to `fixationTrigger`'s dictionary which is the one I'm touching. So it's available pretty much globally and it might be reasonable to protect it against future changes by making it concurrency-safe. What do you think? Happy either way - I'll just drop the commit when I squash everything if you want to keep it as-is.

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Observables/TriggerSources/KeyFixationSource.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Observables/TriggerSources/KeyFixationSource.cs
@@ -245,9 +245,11 @@ namespace JuliusSweetland.OptiKey.Observables.TriggerSources
                                             {
                                                 t.Dispose();
                                             }
-                                                keystroke = (fixationCentrePointAndKeyValue.KeyValue.Equals(lastKeyValue)) ? keystroke + 1 : 1;
-                                                lastKeyValue = fixationCentrePointAndKeyValue.KeyValue;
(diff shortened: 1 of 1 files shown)


JuliusSweetland:
@kmcnaught I had to merge a change Adam made with your change and this is the result: https://github.com/OptiKey/OptiKey/pull/822/commits/44ee274cb8837c914a43664067468aacceeaf6f5#diff-e9b95dd8ad2edc339bf25f5079525deb21b26f3bce51c5ee7e010ec07a421768R205

Could you please sanity check that I haven't broken anything pls.

PR #824, opened 2022-03-30
Possible fix for disposal of sequence

### LINKED ISSUE #524: Manual mode does not work properly when keyboard key trigger mechanism is used
# How OptiKey was being used

Version 2.20 had just been installed. An able-bodied helper was attempting to help me with configuration. I instructed him to enable manual mode by right clicking on Optikey and selecting manual mode. Manual mode was successfully enabled.

# What the user expected to happen

I expected that my helper would be able to activate keys by clicking on them.

# What the user actually experienced

The orange highlight did follow the mouse, but clicking on keys did nothing. Pressing my trigger button (mapped to keyboard key) gave message "I cannot detect where you are directing your attention…" (Even if I was sitting in front of the screen and the Tobii tray icon indicated that my gaze was detected.)

# Version
 - Operating system and version: Windows 10 (64-bit)
 - OptiKey version 2.20.0.0

# Steps to reproduce the issue

1. Configure Optikey to use keyboard key trigger mechanism. I am using key F 23, but I have verified that the same behavior occurs when using key "I".
2. Right-click on Optikey and select "Toggle Manual Mode." Click yes in the dialog box.
3. Observe that the orange highlight follows your mouse cursor when you move it across Optikey.
4. Try to click on any key. Notice that nothing happens.
5. Try pressing your trigger key. Note the resulting message.

# Any other information that helps to describe and/or reproduce the problem

Just a quick note that this bug is not a big deal for me personally, so, as far as I am concerned, there is absolutely no urgency in this request. It is a very rare situation when I would have an able-bodied helper helping me with Optikey. (Most of the time I am quite independent with gaze and speech input. When I need help or am collaborating, I just hand my helper a physical keyboard.)

Nonetheless reporting this because I know it could be an issue for some. Would be happy to test any fixes, as I have the hardware to do so easily.

# Screenshots

N/A

kmcnaught:
Might fix #524, explanation in issue thread. If approved, I neef to make the same change to other trigger sources

### DIFF
--- a/src/JuliusSweetland.OptiKey.Core/Observables/TriggerSources/KeyboardKeyDownUpSource.cs
+++ b/src/JuliusSweetland.OptiKey.Core/Observables/TriggerSources/KeyboardKeyDownUpSource.cs
@@ -103,7 +103,11 @@ namespace JuliusSweetland.OptiKey.Observables.TriggerSources
                         .DistinctUntilChanged(signal => signal.Signal) //Combining latest will output a trigger signal for every change in BOTH sequences - only output when the trigger signal changes
                         .Where(_ => State == RunningStates.Running)
                         .Publish()
-                        .RefCount();
+                        .RefCount()
+                        .Finally(() => {                            
+                            sequence = null;
+                        });
+
                 }
 
                 return sequence;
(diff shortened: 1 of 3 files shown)


kmcnaught:
Fix now applied to three trigger sources that follow the same pattern / exhibit the same bug

PR #825, opened 2022-03-30
Remove NBug, fix logs of crashes

kmcnaught:
Removes NBug and makes sure:

- Auto-restart happens for new crash
- Auto-exit happens for repeat crashes
- Crash window shows in both cases
- Logs get flushed and closed correctly in both cases
- Logs do include the exception info (previously this was often missing)

We still need to remove any references to the (now untrue) fact that email bug reports may be sent.


kmcnaught:
I have now removed the installer message about crash emails. There are two files which contain that information, when I'm touching the installer next I will delete whichever one isn't used!

PR #828, opened 2022-04-05
Fix controllers app logging

kmcnaught:
Log4net was logging stuff for the TestControllers console app, which we didn't need or want. Generally this didn't matter, but sometimes there was an error with the log file rollover which left ugly error messages in the console. 

This removes log4net config from the core DLL itself, but does not affect the 4 Optikey apps which set up logging for themselves (including all the logs that the core library generates)


kmcnaught:
@annakirkpatrick fyi

JuliusSweetland:
@kmcnaught Just to double check are we ok to remove the log4net.config file? It's a deliberate commit so I presume it's fine.

kmcnaught:
There is still a config file in each of the apps, I have just removed the config file from the core library itself. I have checked that the logs still contain the lines logged from inside the core library. This means other apps (like the controller test exe) can consume the core DLL without logging if they so wish.

PR #835, opened 2022-05-06
x64 build + Tobii API upgrade

kmcnaught:
This is a new branch while we figure out the x64 upgrade, here for review, may be subject to change

- Upgrade dependencies to x64
- Remove some 32-bit-only dependencies for now
- Add Tobii StreamEngine support (loses calibration support)
- Some temporary console apps for testing

Still outstanding:
- [ ] Upgrade Irisbond Hiru libraries
- [ ] Upgrade Irisbond Duo libraries subject to x64 build being provided
- [ ] Remove MyGaze completely (API doesn't support 64 bit and is no longer developed)
- [ ] Test with a couple of different eye trackers to check 
- [ ] Test on PC without any Tobii libraries installed - does it fail gracefully?

To discuss:
- [ ] Consider using adjusted eye tracker time stamps with some basic clock synchronisation (we currently use the system clock instead of the eye tracker clock)
- [ ] Consider calling out to external Tobii calibration exe


kmcnaught:
@JuliusSweetland

PR #836, opened 2022-05-10
Performance + fix to gestures + refactor resources

kmcnaught:
- Cache expensive DpiX/Y query until screen changes - this should give significant speedup to gestures and gaze indicator as well as some core implementation.
- Slight rework of the way we copy resources over to APPDATA, since there was a lot of duplicate code.
- Fix "minimise.xml keyboard not found" error that was caused by using the built-in gestures


AdamRoden:
@JuliusSweetland I looks good to me. 

Seeing the "Minimised" keyboard does remind me of something. I had been using a Dynamic keyboard named "Minimise" so that I could set the size and position to below the screen. I think it would be cool if the built-in "Minimised" keyboard had a field in MC to control its size and position.

Anyway, I'm available to test once things are merged.

kmcnaught:
@JuliusSweetland this is ready to merge for a pre-release.

Ignore our conversation about the application gaze smoothing in the management console, I was wrong.

PR #837, opened 2022-05-12
Remove the two most harsh gaze smoothing levels.

AdamRoden:


JuliusSweetland:
@AdamRoden Is this necessary? I would lean towards keeping the options; people can just decrease the level of smoothing they use, no?

AdamRoden:
One day I decided to test the last setting and the result was that it tracked so slowly that Optikey was unusable. I couldn't type or get MC open to reset things. I needed help to restart Optikey.

PR #839, opened 2022-05-18
Feature/x64 tobii

kmcnaught:
Changes to SLN to make it easier to deploy with the correct target

Changes to installer to 
- fix x64 build 
- prep resources for future UI change (re: eye tracker selection)
- only give options on first install (since it's overridden by local settings later)


kmcnaught:
@JuliusSweetland please roll a pre-release of Pro only (these changes are pending for other apps). No need to make available for download, but please can you test it doesn't crash for you if you choose the Hiru option. You need to ensure you've installed the 64 bit drivers.

kmcnaught:
Just added commits that apply all the same AIP changes to the other apps (Symbol, Mouse, Chat) - they should all be in same state now. Nothing new to test right now, butat some point we'll need to test all 4 installers that you've rolled.

PR #841, opened 2022-06-21
Bundle Irisbond dependencies

kmcnaught:
These ffmpeg DLLs seem to be required by IrisbondHiruAPI.dll,
bundling them in the app directly avoids need to install any
Irisbond drivers for the Hiru

(these are the 32 bit DLLs - we will also bundle the 64 bit ones in 
our new v4 branch)


kmcnaught:
@JuliusSweetland

PR #843, opened 2022-06-23
Remove obsolete string

kmcnaught:
@JuliusSweetland one missed commit, removes instruction to download drivers for Hiru (since it is no longer required)


kmcnaught:
+ added fix for installer properties, which sometimes had stray quote characters leading to a crash when parsing in management console.

PR #871, opened 2022-10-11
Add Keyboard Editor to Management Console

AdamRoden:
It took an overhaul of dynamic keyboards and the migration of various pieces of code to different classes in order for this to be built properly.

We should get many stakeholders to beta test this one to ensure their versions of keyboard files get interpreted as expected. I'm happy to test any keyboards shared with me if people submit them.

Let the review begin.


JuliusSweetland:
@AdamRoden We are starting to work on a potential pre-release. Do you consider this PR ready for testing?

AdamRoden:
@JuliusSweetland The PR is ready to test.

JuliusSweetland:
Great thanks. I have to prepare a pre release without this pr first to be
able to test and stabilise that work, but this will be integrated after
that. Have a good weekend.

Issue #744, opened 2020-11-01
Lock left button single/double click function


Ar2rZ:
**Is your feature request related to a problem? Please describe.**
Got a request from one of users, who is using single/double click mouse click for a work.

**Describe the solution you'd like**
Single/Double click function lock - so, user locks single/double click button action- so user do not need to click on this button again. Mouse single/double  click is performed till user unlocks single/double click function..

**Describe alternatives you've considered**

**Additional context**

SCEhardt:
@JuliusSweetland  @AdamRoden @kmcnaught 

Any ideas on the best way to make this happen?  Other buttons can be locked because staying on them causes them to press twice in a row.  However on these mouse click buttons, the first press transforms the UI to a mouse click, which the OptiKey button doesn't respond to.  (Then when the mouse click is done, you're back at the initial state.)  

My first thought is maybe there's a way to detect the mouse click on OptiKey and lock the button if it's the same one currently doing the click.  Thoughts on this or other solutions?

kmcnaught:
There was a lot of previous discussion here: https://github.com/OptiKey/OptiKey/issues/421 

It mainly came down to making a decision on a reasonable UX pattern that wouldn't have any ambiguity between whether a user intended to dwell again on the key to lock/unlock it, or instead perform the current mouse event on the Optikey keyboard itself. And how that would behave in edge cases, where e.g. you want to click near the key. 

One option would be to go back to a previous suggestion from @JuliusSweetland of having a "lock mouse actions" key. We could probably extend this from the original suggestion, and have the following UX:
- select "mouse move and left click"
  - performs single click
- select "lock mouse action key" to hold down
- select "mouse move and left click"
  - enters a looping state until either a different mouse action is selected (which would itself be looped), or "lock mouse action" key is released

This would be able to support locking for any mouse actions, e.g. left click, left double click, right click, hover.

SCEhardt:
@AdamRoden I'm not sure where to apply the suggested code; I tried a couple things but didn't see a difference.

@kmcnaught Thanks for linking the other discussion.  I'm still hung up on the problem that mouse clicks don't activate the virtual keyboard.  I think this would lead to a dead-end in either the traditional lock behavior or the new button you suggested:
- If we make the key lockable in the same way as other keys such as Shift: It will never be possible to get to the locked state because the click action doesn't apply to the keyboard, so the button will always reset before it can be selected twice.
[...]

PR #874, opened 2022-11-17
Ongoing fixes

kmcnaught:
Several things from https://github.com/OptiKey/OptiKey/projects/2, commit messages speak for themselves

### DIFF
--- a/installer/OptiKeyChat.aip
+++ b/installer/OptiKeyChat.aip
@@ -666,8 +666,8 @@
     <ROW Action="SET_SHORTCUTDIR" Type="307" Source="SHORTCUTDIR" Target="[ProgramMenuFolder][ProductName]"/>
     <ROW Action="SET_TARGETDIR_TO_APPDIR" Type="51" Source="TARGETDIR" Target="[APPDIR]"/>
     <ROW Action="SetPropertyEyeTrackerSelected" Type="51" Source="EYETRACKER_SELECTED" Target="MousePosition"/>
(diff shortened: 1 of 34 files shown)


kmcnaught (commented):
@JuliusSweetland This looks like a bug but wasn't causing a problem, could you please pass your eyes over it too

AdamRoden (commented):
Regarding key selection while point selection is underway, how does this behave for you? I did the exact same thing, and when I tested I was able to select keys during point selection, but the selected keys never seemed to be acted on. This lead me to believe we may need to implement a queue so newly selected actions fire after the point selection action completes.

kmcnaught:
force-push to fix last commit and remove WIP re point selection

@AdamRoden I'd only just got that far as a first step, still investigating...

JuliusSweetland:
> 
@kmcnaught I can't see exactly what you are referring to, but all commits look good to me. Thanks.

kmcnaught:
@JuliusSweetland feel free to have a play with left click locking feature here. I've braindumped my current thoughts re: UX / next steps over at https://github.com/OptiKey/OptiKey/issues/744

kmcnaught:
I have force pushed with squashed commits and an update to fix unit tests that broke when selection event API changed

kmcnaught:
I just force-pushed with a revert of [fa0d466](https://github.com/OptiKey/OptiKey/pull/874/commits/fa0d46662c6bac933094cc210708df6cb4b0c140) before my work. This avoids a temporary bug introduced in fa0d which is later fixed in #871

This PR (#874) is ready for pre-release now, with the next followup being [ fa0d4666 + #871 ] for separate testing

Issue #798, opened 2022-03-04
Separate 'Language Enum' to UI and Keyboard


kevinlin18:
Hi @JuliusSweetland ! Is that possible to separate 'Language Enum' current used by UI language and keyboard language? or a PR about it is welcome?
There are 3 English options: Canada, UK, US actually shared same UI, the difference is dictionary. I think we cam merge them in UI list, keep separate in keyboard list.

I translated UI into Chinese Traditional and tried to make bopomofo IME work in Optikey.
Localisation is 100% worked, but IME is not in ideal condition.
So I think we can make a PR about UI language first.
But it will crash if someone choose Chinese on keyboard list, so this the second reason I think we should have a new Enum.

Sorry for post again, but I didn't get e-mail reply or response in #148 ,maybe GitHub didn't notify you

JuliusSweetland:
Hi @kevinlin18. Apologies for the lack of reply - I wasn't notified for some reason.

1.It would be possible to reduce the language list for the UI, but it's only English which is repeated in the drop down, so the impact would be pretty minimal. I'm happy to leave it.

2.I am happy that you are working on Chinese. This is a hard problem with the IME, but if it can be solved Optikey could help a lot of new people. In terms of the UI translation - that is great, but as you suggest it will not work until the IME problem can be solved. Once you have a way forward with that we can solve the crashing and integrate everything into a full Chinese solution.

Thanks for your work.

kevinlin18:
Thanks, we can solve that after IME is done.

Please have a try with 'zh-tw' branch, I'll explain later where I need your help 
https://github.com/kevinlin18/OptiKey/tree/zh-tw

https://youtu.be/Ax_l18Hz7QU

JuliusSweetland:
@kevinlin18 Hi Kevin. I have checked out the branch and tested, looked over the commits, and watched the video. All very impressive! I'm really excited about the prospect of Optikey having support for Chinese, so thank you! I have a few questions at this point:

1.It looks like this new language is oriented around Taiwan. My (extremely basic!) understanding is that Taiwan speaks Mandarin and writes in "Traditional" Chinese. This lines up with what I see in the code; the Language enum entry is Taiwan specific and the dictionary commit is commented "Bopomofo" which I understand to be Mandarin specific.

[...]

PR #876, opened 2022-12-04
WIP Add Chinese keyboards

kevinlin18:
Original discussion: https://github.com/OptiKey/OptiKey/issues/798

Support Chinese keyboards by integrate [Rime Input Method Engine](https://github.com/rime)
Features:
- UI translation: Chinese Traditional
- Quick switch between Chinese/English keyboards, can specific dictionary (Canada/UK/US)
- [Split UI, Keyboard language list in management](https://github.com/OptiKey/OptiKey/commit/793e3e74acfe0d9e6c4980c4188e7e8f192240a9)
- Supported input methods(keyboard) list:
  - Bopomofo
  - Cangjie
  - Luna Pinyin
  - Terra Pinyin

TODO
support stroke keyboard https://github.com/rime/rime-stroke

Known issues
'rimePreedit' not reset when leave a keyboard

Library:
[librime 1.7.3](https://github.com/rime/librime/releases/tag/1.7.3)


kmcnaught:
@JuliusSweetland  

I've asked a chinese friend to clarify some things for me, but in the meantime these are the existing issues I think are present:

- "Clear" sometimes clears the RIME input (i.e. removes suggestions) without clearing the scratchpad. It needs to clear both
- I don't think "Back Word" works 
  - I need to ask for clarification on how it would be expected to work
- When switching between Chinese and English or vice versa, the prediction still contains part of what was already typed leading to various strange behaviours. Something needs to change here but I've asked for clarification on what would be reasonable, for example:
  - auto-add a space when switching from chinese to english? Should it take the current top prediction, or clear the current input?
  - clear any un-used input when switching from english to chinese?
- When switching from Chinese to english, sometimes the predictions bar misbehaves, with the wrong term highlighted due to the shift from 4 -> 6 suggestions giving a weird scaling bug. This might disappear if we are clearing suggestions when switching.

JuliusSweetland:
Great to have some clear guidance. Thank you

kmcnaught:
I've been reviewing how the language is set in the installer, which needed some minor changes. Note that the Languages enum has been slightly overloaded - there is a single `Languages` enum whose constants describe either UI Language (all the previous ones, plus `ChineseTraditionalTaiwan`, and/or Keyboard Langauge + input methods (all the previous ones, plus 10 new ones like `ChineseSimplifiedBopomofo`, `ChineseTaiwanTraditionalLunaPinyin`, `ChineseTraditionalTerraPinyin`. So we have one enum list, but some enum constants don't correspond to a UI language, and some enum constants don't correspond to a keyboard language. This logic is mainly handled by presenting two different list of values to the user depending on the context. It feels like something that should be split out if we ever add another language that has more than one input keyboard available - the language and the input method should be nested. In fact, this would also suit the Simplified keyboards which are currently handled by a different boolean flag. 

_I'm not proposing this refactoring is essential now, but flagging it as something that should be considered next time we poke the nest. And it's important context for the following remarks._

In the installer, we allow people to choose a language, and (until now) use it for both the UI language and keyboard language. This can be separated in the management console if people want them to be different. This doesn't work for chinese though, because the list of available Chinese UI languages and Chinese keyboard languages are disjoint.

We also have a bunch of preexisting logic which tries to be smart about defaulting to the local culture if possible (e.g. if Windows is in French, you have "French" preselected in the languages dropdown but can choose another). 

What I've done, which I think is reasonable given the current constraints, is this:
- in the installer, the user is given the list of *keyboard* languages to choose from, e.g. all the input methods like `ChineseSimplifiedBopomofo`, `ChineseTaiwanTraditionalLunaPinyin`, `ChineseTraditionalTerraPinyin`. 
- This is used as the default keyboard language for Optikey
- If a "Chinese" option is chosen then the UI language is automatically set to `ChineseTraditionalTaiwan` which is the only translation we have available. 
- If the system language is "zh-CH" (China) or "zh-TW" (Taiwan) then the dropdown should default to _**one of the options**_ that supports Simplified (for China) or Traditional (for Taiwan). I think this is an okay heuristic, I don't actually know which it will pick (probably the first or last in the list). I haven't tested this as changing my PC to Chinese is scary, but I don't think it's critical as it's just a nice-to-have hint. It would also be okay if it failed and the dropdown defaulted to English.

kmcnaught:
General update - this is ready for testing. 

To Do:
- [ ]  Need a translation for  "English dictionary for Rime" which I think should just be "English dictionary" in the following screenshot
- [ ] Need to work out what to do with Symbols keyboard (based on testing)


screenshot

kmcnaught:
@JuliusSweetland this branch is now ready for testing both Chinese + touch input. You should be able to make an installer fine if required, or just share the `bin` directory for testing.

zhangbw20:
Thanks a lot for the developers! @JuliusSweetland When will this version be merged into the main brunch and get into the next version's release? We really look forward to seeing OptiKey that supports Chinese!

kmcnaught:
Hi @zhangbw20 this is already in [v4.1.0](https://github.com/OptiKey/OptiKey/releases/tag/v4.1.0), please try it out and give us some feedback. The way we support different eye trackers has changed in v4+, get in touch at [help@optikey.org](mailto:help@optikey.org) if you need any support in the transition.

zhangbw20:
> Hi @zhangbw20 this is already in [v4.1.0](https://github.com/OptiKey/OptiKey/releases/tag/v4.1.0), please try it out and give us some feedback. The way we support different eye trackers has changed in v4+, get in touch at [help@optikey.org](mailto:help@optikey.org) if you need any support in the transition.

Thanks a lot for your swift response. I got the 4.1 version and tried to install but it told me that "the installation of OptiKey Pro (v4.1) was prematurely terminated due to an error" even if we uninstalled and deleted v3.2.5 as well as things in registry and C disk's roaming, maybe I will ask help@optikey.org for help now. Thank you! After successfully installed and used the new version, we would like to participate in the feedback!
image

kmcnaught:
Oh dear. Did the installer give you an error log? If so, please send email it over. If the installer didn't open it for you, you should still be able to find it in your %TEMP% directory.

- Open File Explorer
- Type %TEMP% in the address bar (including the %) and press enter
- Look for a recent text file with "MSI" in the name, maybe something like MSI_abc123.log ? (with different letters).

zhangbw20:
> Oh dear. Did the installer give you an error log? If so, please send email it over. If the installer didn't open it for you, you should still be able to find it in your %TEMP% directory.
> 
> * Open File Explorer
> * Type %TEMP% in the address bar (including the %) and press enter
> * Look for a recent text file with "MSI" in the name, maybe something like MSI_abc123.log ? (with different letters).

Thanks, the email was sent to help@optikey.org, and in short we are not able to capture any log file at the error occurence. But we somehow find that the same OptikeyPro.exe (v4.1) could be installed on an English interface Windows computer smoothly even during the installation we choose the language to be for example "Luna Pinyin" and Chinese input was tested with a few words and we didn't find problems yet. (No error occurs on an English Windows computer, but 4 Chinese Windows computers were tested and they all failed. In other words, our observation currently is that when the installation interface is shown in Chinese Chracters, they never succeeded with v4.1.)

zhangbw20:
> > Oh dear. Did the installer give you an error log? If so, please send email it over. If the installer didn't open it for you, you should still be able to find it in your %TEMP% directory.
> > 
> > * Open File Explorer
> > * Type %TEMP% in the address bar (including the %) and press enter
> > * Look for a recent text file with "MSI" in the name, maybe something like MSI_abc123.log ? (with different letters).
> 
> Thanks, the email was sent to [help@optikey.org](mailto:help@optikey.org), and in short we are not able to capture any log file at the error occurence. But we somehow find that the same OptikeyPro.exe (v4.1) could be installed on an English interface Windows computer smoothly even during the installation even we choose the language to be for example "Luna Pinyin" and Chinese input was tested with a few words and we didn't find problems yet. (No error occurs on an English Windows computer, but 4 Chinese Windows computers were tested and they all failed. In other words, our observation currently is that when the installation interface is shown in Chinese Characters, they never succeeded with v4.1.)

When we select Region and Language -> Formats -> Format to be anywhere with English, then Win 10 and Win 11 systems can be successfully installed with OptikeyPro.exe (v4.1). But now we are finding problems with Win 7, as the person that really needs this has Win 7 computer: Changing formats of language did not work on Win 7 for the error during the state of configuring Optikey Pro - "EXE and MSI file signature mismatch". And any earlier installers till v3.2.0 all fails at the same place.

kmcnaught:
@zhangbw20 thanks for the updates. Is the Windows 7 machine running 32-bit or 64-bit Windows? I think these instructions would help you find out, if you don't know the answer

-  Click the Start button.
-  Select Control Panel.
-  In the search box, type Performance Information and Tools, and then select it from the list of results.
-  Select View and print detailed performance and system information.
-  In the System section, you can see what type of operating system you're currently running under System type.

The system language issue seems separate, I will try to replicate on my PC.

zhangbw20:
> @zhangbw20 thanks for the updates. Is the Windows 7 machine running 32-bit or 64-bit Windows? I think these instructions would help you find out, if you don't know the answer
> 
>     * Click the Start button.
> 
>     * Select Control Panel.
> 
>     * In the search box, type Performance Information and Tools, and then select it from the list of results.
> 
>     * Select View and print detailed performance and system information.
> 
>     * In the System section, you can see what type of operating system you're currently running under System type.
> 
> 
> The system language issue seems separate, I will try to replicate on my PC.

@kmcnaught Thank you! It's 64-bit Windows. We finally downloaded and successfully installed v2.32.0 (the latest version of v2 and it was successful for Win 7). We tested several later versions like v3.0.2 and after until v3.20, v3.25, and v4.1.0 but all failed with "EXE and MSI file signature mismatch". Although the earlier versions do not support Chinese, we can do much for the person in need as the software got installed since language difference is not the biggest problem in our case. Thank you again for your help!

kmcnaught:
@zhangbw20 super helpful troubleshooting, thanks for feeding back. I'll take a look at what changed in the installer between those versions, but I won't get a chance until next week. I'm glad the old version is proving useful in the meantime.

kmcnaught:
@zhangbw20 bad news - I created a virtual machine running Windows 7, and failed to replicate your issue...

Do you have any other Windows 7 computers you can run on to compare? Can you please confirm, when do you get the error "EXE and MSI file signature mismatch"? Does the installer load correctly and then hit an error later in the process, or does the error appear immediately upon running the EXE?

I will check out the language installer issue next - that sounds like it will affect many other users too.

kmcnaught:
I'm closing this PR, I've moved the ongoing actions to:
- windows 7 bug: https://github.com/OptiKey/OptiKey/issues/922
- fixes for installer on simplified Chinese systems: https://github.com/OptiKey/OptiKey/pull/921

zhangbw20:
> @zhangbw20 bad news - I created a virtual machine running Windows 7, and failed to replicate your issue...
> 
> Do you have any other Windows 7 computers you can run on to compare? Can you please confirm, when do you get the error "EXE and MSI file signature mismatch"? Does the installer load correctly and then hit an error later in the process, or does the error appear immediately upon running the EXE?
> 
> I will check out the language installer issue next - that sounds like it will affect many other users too.

@kmcnaught Thank you so much for your help! Two Win7 x64 Laptop with Simplified Chinese interface got this identical error. It occurs at when you click install in this interface. (Currently I don't have a Win7 at hand, and this is Win10, but the error exactly happens next moment.) Later within a few days I will try to test on other Win7 systems. I guess (only a guess) if the Interface for the system is Chinese, the error might trigger. While if the Windows interface is in English, then it will be fine. (At least for Win10, we tested 3 computers in Chinese and 1 computer in English, we need set the Chinese three Region -> Format -> Format to something not simplified Chinese and then the installation would be a success. )
OptikeyPro_MSI_EXE_error

kmcnaught:
I believe there are 2 separate issues. One is that the latest installers (all versions v3 - v4) will fail when the system language is Chinese (zh-CH) due to a bug in how translations are marked up for zh-CH and zh-TW. This is fixed by https://github.com/OptiKey/OptiKey/pull/921 and hopefully @JuliusSweetland can roll a new release with these changes soon. 

I believe the "EXE and MSI file signature mismatch" is a different issue, which only shows up on Windows 7. You noted that you could install v2.32 okay. This version (and older) were made with a simpler installer process, so hopefully there is something we can fix in the newer installers to resolve this. These older versions shouldn't be affected by the system language, FWIW. 

I would expect that the new installer that Julius will build (soon, hopefully!) will fix the first issue but not the second. If you are able to test on other Win7 systems, please let us know what you find out.

zhangbw20:
@kmcnaught Great! I guess the first problem would be solved and in the next version, Chinese speaking users hopefully might not need to change the language settings any more. 

On another Win7 64x PC (not laptop), v3.0.2 can be installed and used successfully, so maybe "EXE and MSI file signature mismatch" is not a general bug for Win 7 with Chinese Interface. v4.1.0.0 can also be successfully installed but need to change the Control panel language to English, otherwise "installer early stop" error would occur. But there's a problem on Chinese characters shown in garbled text when Lunar pinyin (simplified Chinese) was selected. 

image

PR #880, opened 2023-01-10
Ongoing v4

kmcnaught:
Update v4 from master branch by rebasing

Skipped these installer-related commits, TBC whether we need to manually apply any of them:

**d83f5f70 - vcredist as a prerequisite in the installer**
I believe this was for the legacy 32-bit Tobii integration and might not be relevant

**1680a09b... Update installer configs to v3.2.3**
(not relevant)

**4eb47947... Bundle Irisbond dependencies** 
This bundles some native ffmpeg dependencies to avoid an Irisbond install. I do not know whether it is required for the 64 bit Hiru drivers, and if so, whether we need different DLLs


kmcnaught:
I have tested that it builds and runs but not carried out extensive testing

kmcnaught:
As for https://github.com/OptiKey/OptiKey/commit/4eb479478e1f185d8b793e7d9f58ea9be9a6be6d, these are all 64-bit DLLs and the commit should have been made on the v4 (64 bit) branch! Not sure when that snuck in, maybe the PR was wrongly assigned. So this commit needs taking wholesale to `feature/x64-tobii`, and removing from `master`. I don't know whether we can/should do a similar thing for the 32bit app. I imagine the 32 bit Irisbond installer will unpack the dependent DLLs and put them somewhere on the client machine, in which case we can copy them from there.

kmcnaught:
It looks like past me already sorted out 32-bit and 64-bit DLLs for the respective branches, hooray! So https://github.com/OptiKey/OptiKey/commit/4eb479478e1f185d8b793e7d9f58ea9be9a6be6d is not required here.

I note that the Irisbond installer (which we are trying to avoid) comes with vcredist_2017_x86.exe and vcredist_2015_2017_2019_x64.exe so that is likely a dependency for Irisbond as well as Tobii. I'll check the versions and bundle appropriately.

kmcnaught:
I can add "Visual C++ Redistributable for Visual Studio 2015-2022 x64" which should cover all bases

kmcnaught:
@JuliusSweetland should be good to go now, but please do test your end too.

PR #894, opened 2023-04-14
Brazilian Portugueses Tranlation.

MardSilva:
Hi! 
I've added the translation file for (Brazilian Portuguese). Well, i've tried to add on the 'Dictionaries' folder, but, didn't solved. Only the Core folder had the Brazilian dictionary file (.dic). Just add on the other projects and folders who has the translation to work on Portiguese.


alexandre-mbm:
Olá! Como falante do mesmo idioma, eu gostaria de entender melhor o que está se passando.

kmcnaught:
@MardSilva Thanks for the contribution, I've fixed this up. The translations file only needs to be in Core but it needed adding as a `EmbeddedResource` and copying to the build directory. I've also hooked up a few more things. 

The only issue now is that the dictionary file you are using is enormous compared to what we normally use and there's a significant performance hit when loading the file. We normally aim for 75k-100k words in size. Here are Julius' original guidelines:

> i)It should be roughly 75k-100k words in size (it can be a bit less or more, but there is a performance overhead for huge dictionaries so there is no point choosing one that is 1 million words long).
ii)No entries of 1 character as the user can type these from the keyboard and they will be ignored.
iii)All words should be lower case, unless the case has actual meaning, e.g. proper nouns like names, abbreviations like "ATM", or some other meaningful use of capital letters. OptiKey will case words automatically otherwise.
iv)There is no need to manually inspect the whole word list, but sanity check it for mistakes, swear words and colloquialisms that wouldn't be used by the average person. It also doesn't need to contain obscure words. Generally speaking it should cover a normal vocabulary for an adult speaker.
v)If your dictionary includes ligatures (e.g. "œ" and "æ") then please convert them into the individual letters (e.g. "oe" and "ae"). If you think this is not appropriate please contact me to discuss options.
vi)The dictionary files have to be encoded in unicode (UTF-8). If you encode them in ASCII some characters (e.g. ü) will instead be displayed as the 'replacement character' (�). To do this just copy/paste the whole word list into Notepad and save the file with the Encoding set to "UTF-8"

Can you please try to find a more suitable word list?

kmcnaught:
The dictionary here has 260k words which is closer to what we want: [WinEdt Unicode Dictionaries](https://www.winedt.org/dict.html) 

But a smaller one would be better if possible please

kmcnaught:
(force pushed tiny change after renaming file but not updating in csproj)

Issue #897, opened 2023-05-05
Look To Scroll isn't working when Optikey docked vertically


kmcnaught:
Using current latest code (v3.2.5_prerelease-3).
 
When Optikey is docked at the top of the screen, you can maximise the built in mouse keyboard and use the LookToScroll key to set a centre point deadzone about which scrolling occurs. When Optikey is docked at the right hand side of the screen, nothing happens when you try to turn on look to scroll and set the centre point. 

See video:
https://www.loom.com/share/596b2f4b087d4a46b18d18a6b14c4b02

~~Fairly significant changes to the mouse layouts occured in these two commits, perhaps something was missed in the vertical layout?~~

~~ba54c012246b2598618f74f6df0cac415108616e~~
~~d4773de955d713f46ceb99c093ad85354d735327~~
[Edit: not relevant after investigation]

kmcnaught:
I stuck a bunch of logging in because debugging is very slow when LTS is involved. It turns out that look to scroll is turning on appropriately, but the logic that determines whether the current point is "in bounds" for the scrolling is returning false.

The culprit is the `FindLargestGapBetweenScreenAndMainWindow` method, which is not giving the correct answer for a vertical docking. 

I have also been trying to get LookToScrollActive working in a dynamic keyboard. I have tried various dock settings, and the only time I have made it work is if (a) Optikey itself is Floating **and** (b) the current dynamic keyboard is Floating.

https://github.com/OptiKey/OptiKey/blob/c308f488a61597cd3218fcbd6a920cfb47c875ef/src/JuliusSweetland.OptiKey.Core/UI/ViewModels/MainViewModel.LookToScroll.cs#L573-L584

kmcnaught:
@JuliusSweetland anything jump out at you?

JuliusSweetland:
@kmcnaught It's got to be a bad assumption in the 4 rectangle calcs. I would add a Debug.Print of the 4 rects and see what they come out as when you try. I would assume they are being incorrectly calculated somehow.

kmcnaught:
Turns out it's a multiple-monitor bug, and Anna and I both have 2-monitor setups (mine has primary on the right which is less forgiving of bugs). 

I'm raising a PR with a fix. I've tested with primary monitor on the left, and the right, with all 4 dockings, and with some dynamic keyboards. The change only affects docked keyboards so I haven't re-tested many floating scenarios. 

[...]

PR #900, opened 2023-05-25
Fix LookToScroll for 2-monitor setups

kmcnaught:
This resulted in #897, in which LTS wasn't considered valid with vertical docking of Optikey.


kmcnaught:
@JuliusSweetland can you please approve this and I'll cherry pick it onto v4 also

PR #902, opened 2023-06-06
Fix logging on restart

kmcnaught:
When an Optikey app was restarted (e.g. via Management Console, context menu, or crash handling) the old log file was still open for writing when the new log file was established and the rollover carried out.

The result was that the old log file (from the instance requesting the restart) was lost completely, and there was a gap in the list of rolled-over log files, e.g.
```
EyeMine.log
EyeMine.log.2
EyeMine.log.4
```

This change ensures that a new process starts a new log file, after the old process explicitly closes down the old log file.


kmcnaught:
@JuliusSweetland

PR #906, opened 2023-07-24
Add log for command line args

kmcnaught:
Previously if you entered a command line arg (file or folder) incorrectly it silently failed, so just adding a log for this situation.


kmcnaught:
@JuliusSweetland just flagging that I merged my own PR here, hoping to get tiny log in before you roll a release :-)

PR #909, opened 2023-08-11
Fix and update URL on EyeTracker page

kmcnaught:
This includes removing local overrides to the shared dialog to make sure they are all identical

No change required for v3 installer which has different dialogs.


kmcnaught:
@JuliusSweetland

PR #910, opened 2023-08-30
Touch screen + RIME integration

kmcnaught:


kmcnaught:
Full disclosure - I snuck in an extra feature for gamepad triggers, which lets you hold down your gamepad button to get repeat presses (after a delay). This has proven very useful for one user within my own coding fork, and I think will be useful for another user currently using offical Optikey. It doesn't affect anyone not using gamepad input as a switch. It was somewhat accidental that it came into this PR, but I would like it in there at some point so it was easier to handle the merge conflicts now rather than remove it and deal with later. Hope that's okay!

JuliusSweetland:
@kmcnaught Can you remind me what the state of this PR is? Should the pre-release I'm going to build contain both this PR and your latest plugin PR? Thanks

JuliusSweetland:
@kmcnaught I think this does need to be merged, but it now has merge conflicts after I merged your other (plugins) PR. Should I attempt to resolve, or are there things which are difficult? Ty

JuliusSweetland:
@kmcnaught I might have to ask you to attempt to rebase this against main I'm afraid. My understanding is that we want to merge both this and your plugin PR against v4 (main), not v3 (main-v3). I don't think this PR will work with v4 as it has different external assemblies, among other things. For example:
image

Let me know if I'm misunderstanding the plan and I'll try again. Thanks.

kmcnaught:
you're right. I don't know if I just picked the wrong base when I made the PR, or if we've just moved on since then. I'll try a new PR and resolve any conflicts

JuliusSweetland:
@kmcnaught Thanks that would be great.

kmcnaught:
@JuliusSweetland I have force-pushed a new version, and changed the PR's base branch. 

Please test. 

I have not yet tested the installer, that might have lost some changes in the merge relevant to chinese. But please don't delay on testing DLLs and chinese, it urgently needs a review of the features before a final review of everything being fully checked for deployment.

JuliusSweetland:
@kmcnaught Merged into main. Will test and update you.

PR #913, opened 2024-03-05
Add DLL support

kmcnaught:
## What is included in this PR:
- Split out some core functionality to define public interfaces for DLLs
- Add support for external DLLs to be used as PointsSource
- Add management console UI to allow installing & management of DLLs

In the constructor of `GithubPluginsSearch` you can set a github API token, for high-volume testing. I can send you one. 

### github errors
On startup, we asynchronously call out to github to update our list of available repos. If this fails for any reason, we log but do not toast the user. I think this is okay, but it does mean we could miss some unpredicted unrecoverable error that we might want to be aware of...? 

FWIW if the issue was a network connection, we try 3 times before giving up (to allow for users who have Optikey start automatically on startup when the network connection may not be ready)

### Coercion of application smoothing level 

Currently, if you install a new plugin, and then close the plugins window, it asks if you want to select your newly--installed plugin as the points source. If you say yes, it changes the point source and defaults the smoothing level back to zero. You can then change it to something else if you like. 
It does *not* prompt you or coerce anything if you choose a plugin in the dropdown manually. 
Is the current behaviour what we want?

Given that we are instantiating the points source at install, we could add to our interface a notion of “recommended smoothing level” that defaults to zero. I ruled this out previously as I wasn’t sure how to expose this without instantiating the class. The downside is that it locks us in a bit to the ‘meaning’ of our application smoothing levels. The upside is people who want the built in smoothing can have it, and others can implement their own and expect ours not to interfere (i.e. only if a user turns it on again)

### Test architecture
I had to set processor architecture for tests to get them to work with the 64 bit build - we may need to do this in CI ? You can also use the .runsettings file I added, but this also needs hooking up in VS/CI. 

### Failing test 
In `ThenSelectionProgressShouldBeProcessed`, the `AudioService` assertion is failing. I suspect this was broken back when we switched to 64 bit and I didn't realise tests weren't running. I'll try to track it down

### DIFF
--- a/.runsettings
+++ b/.runsettings
@@ -1,0 +1,1 @@
+<TargetPlatform>x64</TargetPlatform>
--- a/OptiKey.sln
+++ b/OptiKey.sln
@@ -5,6 +5,7 @@ VisualStudioVersion = 15.0.28307.1259
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "JuliusSweetland.OptiKey.Core", "src\JuliusSweetland.OptiKey.Core\JuliusSweetland.OptiKey.Core.csproj", "{6865C5AF-C6F0-4BDA-BA1C-7AE8BC225234}"
 	ProjectSection(ProjectDependencies) = postProject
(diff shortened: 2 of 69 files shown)


kmcnaught:
Also need to add a warning dialog that explains that files are coming from the internet, use at your own risk, etc

PR #917, opened 2024-05-01
Installer fixes + contract change

kmcnaught:
- Remove the the unused `KalmanFilterSupported` bit of the plugin interface. This means that previously-tested DLLs won't work, but I've updated the Smooth Mouse example (and others are no longer labelled so won't appear - they were just for testing)
- Fix properties being set by installer for chinese and for touch
- Chinese: split UI language and keyboard language (we support many more keyboards than UI languages)
- Touch: change triggers and multikey settings if touch input is chosen

This PR includes updating AIP files for all apps.


kmcnaught:
I have added the discussed changes to the installer dialogs, including translations

kmcnaught:
Plus some chinese translations (with help of GPT4) for the new features (touch, some gamepad options) that missed out the original translation effort. 

Ready for review / new pre-release please

PR #921, opened 2024-08-15
Fix install translations

kmcnaught:
Fixes issue where system language is chinese (zh-CH). Optikey only has zh-TW translations, so crashed on attempt to translate. This was a nuance particular to chinese, where our UI support and our keyboard support differ. So depending what you queried, zh-CH claimed to be supported, but didn't have translations. 

This PR:
- makes sure a missing translation doesn't crash installer
- adds enough (auto) translations for zh-CH to cover what's in the installer
- makes sure installer translation logic includes keyboard languages as well as UI languages


JuliusSweetland:
@kmcnaught v4.1.0 has been re-uploaded with this fix. Available here: https://github.com/OptiKey/OptiKey/releases/tag/v4.1.0

PR #932, opened 2025-05-09
look to scroll fix and conditional uiAccess

tqphan:
Adding back the option to bring window to front when using look to scroll.
Added conditional manifest for uiAccess (false for the debug and true for release)


kmcnaught:
Looks good to me. I'll give it a spin this week.

kmcnaught (commented):
Re: [Make the static splash screen respect the setting ShowSplashScreen](https://github.com/OptiKey/OptiKey/pull/932/commits/0e44be77e2df2c7b4f05dafba7bbc95a8a1ccbf1)

Can you explain this to me? My recollection is that we added the static splash screen to avoid a slow startup looking like Optikey wasn't loading, so the static splash doesn't add any time to the loading process. Is that your experience? My instinct is that it still plays an important role even if people don't want to see the full splash screen (which does delay your ability to use Optikey)

kmcnaught:
Thanks for the explanations. I am adding most of these changes into a PR I am preparing with some of my own stuff too. A couple of comments:

- I've reordered and 'fixed up' commits so each one stands alone
- I've rejected the change to the splash screen - showing the static one does not add any loading time, so I don't think it's helpful (to the general user) to remove it. If you feel strongly, I wouldn't mind you adding an additional setting to disable the static one too (but of course, even more booleans isn't exactly nice)
- You might be interested in the symbol parsing I've used for [OKGO symbols](https://github.com/kmcnaught/OKGO/wiki/Symbols) - I did the same thing with geometry, but also added some symbol libraries too. I tried to display them in the wiki, but hit issues with the github rate limit so those pages are a bit broken and need removing. The RPG symbols probably aren't relevant to Optikey, but the MaterialDesign ones are useful. 

See [my version of DynamicKeyboard.xaml.cs#L192](https://github.com/kmcnaught/OKGO/blob/c45402d4c44c7d2ecb69d7b15003d7472bb457e1/src/JuliusSweetland.OptiKey.Core/UI/Views/Keyboards/Common/DynamicKeyboard.xaml.cs#L192)

